Issue #589, opened 2013-10-09
Links in text should be underlined


Diewy:
The css of the backend doesn't underline links in a text by default.

WouterSioen:
Fixed in 4.0.0

PR #624, opened 2014-01-02
Backend links underlined

ghost:
As stated in issue #589 links in the backend should have been underlined


jeroendesloovere:
@guytarded Can you look for this?

PR #627, opened 2014-01-08
New module: xml sitemap

jonasgoderis:
This is a new module called Sitemap. With this new module it's possible to create (automatic and manually) an indexed xml sitemap & image sitemap based on the pages & modules within the website. 

Two basic modules (blog and faq) are already using this new feature by implementing the interface 'BackendSitemapInterface' into the model. So every new module within a project just needs to implement that interface so the'll be implemented into the sitemap too. 

The xml sitemap can be created manually through the module settings or automatically with the available cronjob. In the module settings you'll also get an overview of which xml files will be generated and all the modules that are not implementing the interface.


jeroendesloovere:
I've created a PR with a Sitemap module.
Each module can be hooked to this Sitemap Builder.
https://github.com/forkcms/forkcms/pull/2329

carakas:
At this moment we are trying to implement this, we are working on two different implementations each with their benefits, if you have time we would appreciate it if you would take a look at them
- https://github.com/forkcms/forkcms/pull/2318
- https://github.com/forkcms/forkcms/pull/2329

PR #863, opened 2014-09-08
Update to ckfinder 2.4

megasnort:
Update form ckfinder 2.3 to 2.4 (which is allowed by the oem-license).
Checkout the changelog here: http://cksource.com/ckfinder/whatsnew


Diewy:
:+1:

PR #867, opened 2014-09-14
Pagerfanta, revisited

turanct:
This is a revisited version of https://github.com/forkcms/forkcms/pull/789, an effort to make pagerfanta do the pagination, instead of having pagination all mixed up between our real controller/action code. Back then, the whole dataset needed to be loaded to be able to do the pagination.

I've now changed the code to work with pagerfanta's `CallbackAdapter`, which enables us to define two callbacks, one for a 'count' of the whole dataset, and one for a 'slice' of the dataset. As a result, we can now keep using e.g. `FrontendBlogModel::getAllCount()` methods for the counting, while also using pagerfanta in a memory efficient way. I've implemented this for the blog module. For the search module, there's this problem with caching.

Currently, the cache works like this:

- We get a page number
- We calculate the limit / offset for the search results based on the number of results per page
- We store the results for the given query + limit/offset in a file, **along with the pagination data**
- If the cached page is requested, the pagination data is loaded from the cache.

The problem is, to do that, the pagination data needs to be a class variable, accessible to the whole controller/action, which in turn is really bad for the separation of concerns that i'm trying to accomplish by making pagination less of a controller issue. That's why for the search module, i've done it like before: load the whole dataset in an array, and cache the whole thing. I'm aware this is an issue. Please don't merge if you don't feel comfortable about it. Help me find an elegant solution :goat:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
         "spoon/library": "1.3.*",
         "symfony/monolog-bundle": "2.3.*",
         "behat/transliterator": "~1.0",
+        "pagerfanta/pagerfanta": "v1.0.2",
         "jdorn/sql-formatter": "1.3.*@dev"
     },
     "support": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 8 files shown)


turanct:
Bear in mind that this could be an intermediate step between {straight db calls + old pagination} and {doctrine + nice pagination}

WouterSioen:
We'll work on this when doctrine is used.

PR #888, opened 2014-10-05
Date & time field support for Formbuilder

jessedobbelaere:
### Description
Added a datetime button to the Formbuilder module, to create a date or time input field inside your form (for simple reservation forms, etc).

This adds a new icon and button to the Formbuilder module with "date & time" support. You can enter the label, a default value, required checkbox and most importantly: a dropdown where you can choose between a date or time field. You'll get 

### Discussion
I find this a useful update to the Formbuilder **but it needs discussion because of a few problems**:

* ~~I use the Spoon `addDate` and `addTime` methods to create the input element. However, these methods don't seem to support the new HTML5 types: `type="date"` and `type="time"`. Spoon actually creates plain textfields, so no HTML5 date & time fields: the html5 datepicker doesn't appear, the time field doesn't have an up/down button. So currently you don't have HTML5 fields in this PR but you can add a date or time field to your form, and it adds the necessary class names for the jQuery UI datepicker. But ideally I'd love to see a HTML5 datepicker by default, and add the jQuery UI as a fallback datepicker.~~ I started using `addText` because then I can use `setAttribute('type', 'date')` to switch to a html5 date or time field. The type attribute is forbidden with `addDate` so that's why I need to use `addText`. So now you have html5 date & time fields in the frontend. I added some js code to check if the date type is supported, if not add a jquery datepicker fallback. The HTML5 date type automatically uses the most appropriate date format (dd/mm/yy) for your browser & location, but the jQuery datepicker needs a date format. It's currently hardcoded to dd/mm/yy but I should probably add it as a setting in the dialog? Same for "firstday" of the week.
* What should I do with the validation dropdown? Keep it or remove it? The HTML5 or jquery fallback datepicker automatically keeps the input validated right? But with no js or html5 field, it should validate the date or time somehow?
* The default value field isn't very useful except for time. However it could have more functionality, e.g. some way to use "+1 weeks" or something as default value. (If you don't fill in the default value, the Spoon addDate or addTime method uses the current date or time.)

### Screenshots
Some screenshots in dutch:
image

image

HTML5 datepicker (Chrome):
image

jQuery fallback (tested in Safari & IE):
image


jessedobbelaere:
I'm picking up this PR again, and I'm trying to implement the `short date format` from the settings of Fork into the datetime field instead of a hardcoded one. But.. while the format in the module_settings is something like this: `j-n-Y` in php, I should parse this into a html5 field but that uses something like this: `dd-mm-YYYY`. Any ideas how you can convert this? Couldn't find anything, but will continue this evening.

WouterSioen:
You could try something like this: http://stackoverflow.com/questions/16702398/convert-a-php-date-format-to-a-jqueryui-datepicker-date-format

It's rather nasty, but it does the trick, and I can't think of a cleaner way (at this moment).

jessedobbelaere:
Hmm should work.. What's the best place to put this script? jquery.frontend.js ? Or something more general?

jessedobbelaere:
I fetched the date format from settings, converted it to jquery dateformat, and parsed it as the data-mask attribute of the data field. 
Note that in chrome, opera etc you get the HTML5 date field, which doesn't allow a data mask. It will display the date according to your browser. So only the jquery ui datepicker (firefox etc) will display the date in this format. 

Now I have this problem: a HTML5 date field has a value formatted in YYYY-MM-DD, so when  you submit a form using chrome, the date e.g. "2015-02-17" gets saved in to the db. 
When you use jQuery UI (firefox), then the date is parsed and displayed according to the data-mask format that you choose in the fork cms general settings. When you save it to the db, you will get something like "17/02/2015" depending of the format you selected in the backend. So this isn't consistent. Any idea how to solve this?

Another problem is that I made a time field in the formbuilder module. When you have chrome, you get the html5 time picker. In firefox you just get an empty textbox. Should I search for a jquery fallback for time fields and implement this? 

I want to finish this PR so I can implement some additional stuff in Formbuilder that I can use in a project: possibility to provide a placeholder text for each of your fields in formbuilder (not that much work to integrate this). And possibility to enter a css class for each field in formbuilder.

And an offtopic question: This branch on my machine is a bit outdated. If I pull from the fork cms master repo, will those commits or a "merge branch message" appear here on this page when I push? How can I avoid that and keep this page clean?

WouterSioen:
@jessedobbelaere I would store dates/times in the database as timestamps. Thats the most uniform format possible.

To make sure your branch is up to date with the master again, you can use the rebase functionality from git. A great article on rebasing can be found here: https://www.atlassian.com/git/tutorials/merging-vs-rebasing

jessedobbelaere:
@WouterSioen I agree, but when you submit the form and fetch the value of the date field in PHP you have two options:
a) The visitor was using chrome so the HTML5 datepicker was used. This stores the date value as YYYY-MM-DD, but shows the date in the datepicker format according to the browser/language/...
b) The visitor uses firefox so the jQuery UI datepicker is used. This stores the date value in the format defined by the form cms `short_date_format` settings, for example "DD/MM/YYYY"

I should find an option to make jQuery UI show the date in the format defined by the user, but store it in the input value as YYYY-MM-DD. Or disable html5 datepicker and convert the jQuery UI date to a timestamp using the user-defined date format.

jessedobbelaere:
Seems like the HTML5 date field can only have a value in `yyyy-MM-dd` format as specified by W3C. So the jQuery UI datepicker should also return his value in this format, but display the `short_date_format` specified by the user in Fork CMS settings. Apparently, the only way to achieve this is to create an [alternative (cloned) hidden field](http://api.jqueryui.com/datepicker/#option-altField) with the same name and different id, that syncs with the date field but has a different format. In the php code, you can use this date (in `yyyy-MM-dd` format) to process it like it was a date from the HTML5 date field.

This is getting way too much than I was originally planning: "Add a date field to formbuilder" :confused:

jessedobbelaere:
Seems like the HTML5 date field can only have a value in `yyyy-MM-dd` format as specified by W3C. So the jQuery UI datepicker should also return his value in this format, but render the `short_date_format` specified by the user in Fork CMS settings. Apparently, **the only way to achieve this** is to create an [alternative (cloned) hidden field](http://api.jqueryui.com/datepicker/#option-altField) with the same name and different id, that syncs with the date field but has a different format. In the php code, you can use this date (in `yyyy-MM-dd` format) to process it like it was a date from the HTML5 date field. 

I had to make some changes to the frontend.js file. Before (!) the initialisation of the datepicker, a cloned/hidden field should be made. Then the original datepicker gets a new id and name with "display" suffix. And then the datepicker is initialized with the `altField` and `altFormat`. I have tested this and it works great. Open up Firefox, create a new date field in the formbuilder module and browse to the frontend form. You'll see the jQuery UI datepicker using the date format you specified in the form cms settings, but it will save and pass the value to php using the general "YYYY-MM-DD" format.

@WouterSioen I hope the cloned hidden field won't break anything? It's the only possible solution. I didn't touched the from, till and range datepickers, but maybe I should do the same there too.

I also have a question: the dates in the admin e-mail and in the database are still plain-text. So I'll need to save the dates as a timestamp. That should happen somewhere around [here](https://github.com/forkcms/forkcms/blob/1abab663e152f45c90616bde6c29d9c3c2f6ace1/src/Frontend/Modules/FormBuilder/Widgets/Form.php#L464-L495). But first the admin e-mail get sent, so I should format the date value like the `short_date_format` setting because it will be displayed in the e-mail. And then right before the values get saved to the db, I should modify the value to be a timestamp. Is that a good way to implement this?

What about the time fields? They don't have a jQuery UI fallback. In browsers (Firefox) that don't support a HTML5 time picker, you get a plain text field. Maybe look for a jQuery plugin or leave it for now?

PR #982, opened 2014-11-23
Added filter by title to content block backend index resolves #976.

### LINKED ISSUE #976: New feature: add filter by name in content blocks index
When having a lot of content block you need to navigate the page number n to find the one you need. Adding a filter would be great.

GhaziTriki:
Added filter by title to content block backend index resolves #976.

### DIFF
--- a/src/Backend/Modules/ContentBlocks/Layout/Templates/Index.tpl
+++ b/src/Backend/Modules/ContentBlocks/Layout/Templates/Index.tpl
@@ -12,12 +12,46 @@
 	</div>
 	{/option:showContentBlocksAdd}
 </div>
-{option:dataGrid}
+
 <div class="dataGridHolder">
-	{$dataGrid}
+	{form:filter}
+	<div class="dataFilter">
+		<table>
+			<tbody>
+			<tr>
+				<td>
+					<div class="options">
+						<p>
+							<label for="email">{$lblTitle|ucfirst}</label>
+							{$txtTitle} {$txtTitleError}
+						</p>
+					</div>
+				</td>
+			</tr>
+			</tbody>
+			<tfoot>
+			<tr>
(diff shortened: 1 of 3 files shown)


WouterSioen:
Can you only show these filters when there are more then 30 content blocks? It seems like an overhead to us to show this in sites with only 10 content blocks.

src/Backend/Modules/ContentBlocks/Layout/Templates/Index.tpl
```diff
@@ -12,12 +12,46 @@
...
+			</tfoot>
+		</table>
+	</div>
+	{/form:filter}
+	
+	{option:dataGrid}
+		{$dataGrid}
+	{/option:dataGrid}
+
+	{option:!dataGrid}
```
WouterSioen: This should be outside the dataGrid holder. The dataGrid holder should only be shown when there is a datagrid.

ghazispeakserve:
Hi @WouterSioen : the last two commits should do the work

WouterSioen:
I've tested this, and it doesn't work. You only set the filter after the datagrid has been loaded, so you always show all the data.

Instead off

        $this->loadDataGrid();
        if ($this->dataGrid->getNumResults() > $this->dataGrid->getPagingLimit())
        {
            $this->setFilter();
            $this->loadForm();
        }

You should have

        if (BackendContentBlocksModel::getCount() >= 30)  {
            $this->setFilter();
        }
        $this->loadForm();
        $this->loadDataGrid();

ghazispeakserve:
It is fixed. Sorry for the delay. I should do more pull requests. Great to work with a such community.

WouterSioen:
I'll close this for now, can be reopened after feedback.

Issue #941, opened 2014-11-12
Implement unit testing with codeception


ghazispeakserve:
Hi Guys,
After some hours of good effort I successfully implemented basic acceptance testing for Fork CMS with console coverage output.

![Fork CMS Codeception Tests](https://cloud.githubusercontent.com/assets/4991088/5007224/0a276998-6a4f-11e4-9819-771745d97a12.png)

Do you think a pull request is a good idea?

PR #983, opened 2014-11-23
Unit testing using codeception

ghazispeakserve:
Added first unit tests using codeception like discussed in #941. Now it is up to the community to add more tests, use selenium web driver and make it successful.


WouterSioen:
Hi @ghazispeakserve 

Have you already looked at these issues?:

> Also: can you make sure the existing parameters.yml file gets backed up before the tests and put back after running the tests, when a parameters.yml file exists? This way, your file isn't overwritten by running the tests. The database should also be cleared after running the tests. It's important that your tests don't change the state of both your instance of the website and the test version (and database).

When these are fixed, we can merge this in Fork.

ghazispeakserve:
I have added the updates you asked for. I have added a drop database only for travis. Dropping local test database can be problematic if the developer wants to check data after running tests and compare them with his development database.

WouterSioen:
Hi Ghazi,

I think it's necessary to drop the database on your local test database
too, since this is needed to be able to run the tests more then once.

Op zaterdag 31 januari 2015 heeft Ghazi Triki <notifications@github.com>
het volgende geschreven:

ghazispeakserve:
One of the problems is whit what db user we create it?

ghazispeakserve:
:sob: I will miss those commits (joking)

PR #1006, opened 2014-12-16
Fixes #661 Added a language parameter to the addMail function.

### LINKED ISSUE #661: Send out multilanguage emails through a backend with 1 language?
When using BackendMailer::addEmail in the backend, messages in an email are being rendered in the Backend Language, rather than the Working Language.

While I understand why this is done, it makes it impossible to send out multilanguage e-mails via the backend based on the frontend languages.

Example: I have a website with 2 languages on the frontend (nl/fr), the backend is accessible through 1 language (en). Say I'm receiving orders via an e-commerce module and I want to manually send out confirmation emails in the backend. When using BackendMailer, {$msgOrderSalutation} will always be rendered in English. If I'm using FrontendMailer, the following error occurs:

Notice: Use of undefined constant FRONTEND_LANGUAGE - assumed 'FRONTEND_LANGUAGE' in /frontend/core/engine/template.php on line 273

Notice: Use of undefined constant FRONTEND_LANGUAGE - assumed 'FRONTEND_LANGUAGE' in /frontend/core/engine/template.php on line 274

Notice: Use of undefined constant FRONTEND_LANGUAGE - assumed 'FRONTEND_LANGUAGE' in /frontend/core/engine/template.php on line 277

This makes sense, because there's no FRONTEND_LANGUAGE to be found.

I hope I'm making some sense here. The only solution that I could currently think of, is hardcoding the emails in both languages. But it would be great if BackendMailer::addEmail had an optional language parameter.

MathiasDewelde:


src/Common/Mailer.php
```diff
@@ -360,4 +388,4 @@ public function send($id)
             Model::triggerEvent('Core', 'after_email_sent', array('id' => $id));
         }
     }
-}
```
WouterSioen: No need to remove the break on the end of the file. All files in Fork end with an empty line. (You can probably set this as a setting in your editor, in Sublime Text, it's "ensure_newline_at_eof_on_save": true)

Issue #1021, opened 2015-01-05
Use Symfony2 Intl Component


ghazispeakserve:
Don't ForkCMS need to use Symfony2 Intl component instead of the outdated Spoon? As example for fetching countries and languages.

WouterSioen:
That would be awesome.

ghazispeakserve:
What if I prepare a pull request for that? or will we need to wait until Doctrine intergation?

WouterSioen:
I think you can already create a pull request without doctrine integration.

ghazispeakserve:
A question about languages.

Actually when calling this method:

```php
Intl::getLanguageBundle()->getLanguageNames(BL::getInterfaceLanguage());
```

we get more languages than ones installed in Fork CMS. Should we keep it like it is or add a filter?

ghazispeakserve:
My question has no sense because ForkCMS supports more languages than Spoon.

PR #1022, opened 2015-01-06
Using Symfony Intl component instead of Spoon to fetch countries.

ghazispeakserve:
ref #1021 using Sympfony Intl component for fetching countries.


WouterSioen:
Looks good at first sight, will test this later.

PR #1029, opened 2015-01-14
Parse js doctrine integration

MathiasDewelde:
The function $this->addJSData could not handle a doctrine entity yet.


WouterSioen:
We don't think it's a good idea to fully serialize objects and parse them to JS. If these objects grow later, more data (that's possibly not needed in JS would be added.)

We would prefer to add a ->getJSData() or something similar in the object and pass the returned array to the addJSData method.

PR #1037, opened 2015-01-19
Template modifier 'truncate', optional truncate on closest word

MathiasDewelde:
Make it possible to truncate text in the frontend on the closest word instead of an exact character count.


WouterSioen:
:+1:

jeroendesloovere:
:+1:

tijsverkoyen:
Wouldn't it be more logical if you search for the last space before the maximum length, instead of the next space. If you want to truncate on 160chars, and the next space is on 164, your text will be too long.

PR #1040, opened 2015-01-21
Bugfix when you defined an user to non-active

jeroendesloovere:
Sometimes an error was thrown when $user['active'] was not set.
So I fixed this with a minor change.


WouterSioen:
Can you readd this pullrequest without the reverted commits?

WouterSioen:
It also contains a lot of changes that where already added in the master branch.

PR #1057, opened 2015-02-10
Bug: Hide/show the localpage feature to fix the href-assignment.

tommyvdv:
When the linktype in the dialog is set to anchor, and the a-element
contains a hash; the link is not properly assigned and the hash-symbol
remains.


WouterSioen:
What has actually been changed in this PR? I can't see the difference between a minified and a clean file.

PR #1063, opened 2015-02-15
Add Symfony style functional tests (WebTestCase) to the InstallerBundle

WouterSioen:
There already is an open pull request with automatic tests (https://github.com/forkcms/forkcms/pull/983). This PR tries to integrate almost the same tests, but using the Symfony WebTestCase. This makes the tests a lot easier to understand and write for Symfony developers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
         - php: hhvm-nightly
 
 script:
-    - vendor/bin/phpunit -c app/
+    - phpunit -c app/
 
 before_script:
     - composer install
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,31 @@ matrix:
     allow_failures:
         - php: hhvm
         - php: hhvm-nightly
+
+script:
+    - vendor/bin/phpunit -c app/
+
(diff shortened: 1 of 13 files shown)


WouterSioen:
Note that some of these commits should be squashed when merging, since I had some trouble getting the tests to work in travis.

WouterSioen:
@ghazispeakserve I'd like your feedback on this one since you already implemented functional tests in the installerbundle.

I think this version is more "the Symfony way" and easier for more developers to add. But it could be that you have some feedback from your experience from adding codeception tests.

ghazispeakserve:
After taking a look at it is an improved "Symfonysed" version of my tests integration. If unit tests with phpunit can cover we need initially it is good. We can still move to codeception if we want to implement acceptance like checking the backend layout behaviour. This is my advice. Keep the good work I am still impressed by the community politeness.

WouterSioen:
@forkcms/core-contributors any feedback on this?

matthiasmullie:
:shipit:

tijsverkoyen:
:+1:

PR #1067, opened 2015-02-16
change $actionClassName to $actionClass because $actionClassName doesn't exist

carakas:


WouterSioen:
already fixed in https://github.com/forkcms/forkcms/commit/5f29e4b830e9035ef92bd111b39c0183af3b93a8

PR #1071, opened 2015-02-17
Only use Json for the Locale cache.

WouterSioen:


src/Backend/Modules/Locale/Engine/CacheBuilder.php
```diff
@@ -186,7 +93,6 @@ protected function buildJavascriptCache($language, $application)
                 }
             }
         }
-        $this->addSpoonLocale($json, $language);
```
tijsverkoyen: I don't see a replacement for this.
In the backend the months in the date pickers don't have localised names, because the "loc" item is missing.

PR #1073, opened 2015-02-17
Issue 969

vytenizs:
Tabs changed to spaces


WouterSioen:
We won't merge a PR to change one js file in coding styles. Every js file should be consistent. It's all or nothing.

Thanks for the effort though. A follow up PR containing all JS files would be welcome. (Though not necessary IMO).

vytenizs:
Ok. Ctrl + Alt + L isnt much effort at all :D

WouterSioen:
I'm glad you didn't loose too much work then :wink:

PR #1078, opened 2015-02-23
Fork functional tests

WouterSioen:
Functional tests on the installer bundle were already added. This PR adds the first functional tests on Fork actions itself. It's just one preview of how it could be implemented, and it could probably be finetuned a lot, but when this is merged, it will be really easy to add more functional tests for Fork, thus making it a lot easier to do refactorings.


app/Kernel.php
```diff
@@ -43,7 +43,7 @@ public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQ
      * This will disappear in time in favour of container-driven parameters.
      * @deprecated
      */
-    protected function defineForkConstants()
+    public function defineForkConstants()
```
WouterSioen: I'm not really happy with this, but it's necessary to have these constants in our functional tests. They should be refactored/removed/replaced by DIC parameters in the future though.

app/phpunit.xml.dist
```diff
@@ -6,6 +6,7 @@
          backupGlobals="false"
          colors="true"
          bootstrap="../autoload.php"
+         stderr="true"
```
WouterSioen: This makes sure \SpoonCookie and some other stuff setting headers doesn't break our scripts. PHPUnit breaks by default on setcookie, header and some other methods.

src/Backend/Modules/Blog/DataFixtures/LoadBlogCategories.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Backend\Modules\Blog\DataFixtures;
+
+class LoadBlogCategories
+{
+    public function load(\SpoonDatabase $database)
```
WouterSioen: slightly based on Doctrine's datafixtures, to make it easier to replace these later.

WouterSioen:
I refactored the redirects in Fork to not break the tests (by not making them do a redirect and make the redirectresponse bubble up). I also added more tests (category, archive, index, detail, missing items, wrong pages,...).

PR #1079, opened 2015-02-24
add comment to the documentation of the parse widget template modefier about random data

carakas:


WouterSioen:
The `@internal` PHPDoc tag is normally used for this. (instead of `WARNING`)

carakas:
fixed in 7e5bb1d

PR #1081, opened 2015-02-24
Blog functional tests

WouterSioen:
* Add functional tests for the Blog actions
* Refactor globals to be able to use redirects in functional tests
* Refactor redirects to not break the functional tests (so not use an exit, an bubble up the RedirectResponse)


WouterSioen:
Since there are no BC breaks anymore, it's merged in master.

PR #1082, opened 2015-02-25
Secure the app dir to avoid public access

DoFken:


WouterSioen:
Wouldn't it be better to use the same format as Symfony? https://github.com/symfony/symfony-standard/blob/2.7/app/.htaccess

This is also compatible with more apache versions.

DoFken:
I used the same .htaccess as in the library directory.
Question is, isn't it "simpler" to alter the root .htaccess that all folders except the src , and .md files aren't publicly available, except the src directory?
The best option in this case is still a separate web directory with only the publicly available content and use that as document root. No?

WouterSioen:
@DoFken A web folter containing only assets and an index php page is really something we'd like. This isn't that easy to achieve right now, though. The assets and files are spread in different places in Fork and it would take some time to refactor all these usages.

I think it's easier to deny everything from .htaccess files in the subdirectories. This makes our root .htaccess file not too cluttered.

tijsverkoyen:
I would prefer the Symfony-way as @WouterSioen mentioned. Of course only a webroot wit the assets and separate src folder would be much nicer, but isn't really doable for the moment.

dubgeiser:
Agreed on the webroot system... but can't we just merge this and have this fixed already?

WouterSioen:
I'd prefer to add the symfony version https://github.com/symfony/symfony-standard/blob/2.7/app/.htaccess which is compatible with apache 2.4.

dubgeiser:
@WouterSioen have you tried that and found it working?
Because, I just did on a 2.4.9 Apache on Yosemite and I could just access the logs.log file :-/
While @DoFken solution _did_ restricted access.

WouterSioen:
Hmmm, that's odd. Then I'll just merge this. This commit https://github.com/symfony/symfony-standard/commit/38c2a50958f3f2a191cca9fe9a703b183fd09065 made me think this was the fix to make it work on apache 2.4

dubgeiser:
yeah, we shouldn't blindly follow ;-)

PR #1086, opened 2015-02-26
Fix formbuilder reply to

ikoene:
Email validation is required when using a field as replyTo emailaddress.


WouterSioen:
dan you rebase this branch on the newest master?

PR #1088, opened 2015-02-26
Add module prefix to the dashboard widget checkboxes.

tommyvdv:
Dashboard widget checkboxes are oblivious of the fact that a widget can
have the same name in two different modules. Allowing only one of these
will add none/all of them to the group.

Specifying the module fixes this behaviour.


tijsverkoyen:
:+1:

WouterSioen:
looks like it's working well.  I don't like the bigger amount of nesting in the code though. Couldn't this be refactored to be a little cleaner?

PR #1093, opened 2015-02-26
Fix bug: trying to load config for missing module yields FatalErrorExcepion

turanct:
when loading a backend url for a module that doesn't exist, without
providing an action, a FatalErrorException occurs, because the config
class we're trying to load doesn't exist.


tijsverkoyen:
:+1:

PR #1096, opened 2015-02-26
Use late static binding.

### LINKED ISSUE #786: Use `static`-keyword instead of `self`where it is relevant
See http://php.net/manual/en/language.oop5.late-static-bindings.php and http://stackoverflow.com/questions/4718808/php-can-static-replace-self

dubgeiser:
This way we can accommodate inheritance better.
Changed everywhere for consistency.

Closes #786


src/Backend/Core/Js/ckeditor/ckeditor_php5.php
```diff
@@ -1,556 +1,556 @@
...
-		// String otherwise
-		if (strpos($val, '@@') === 0)
-			return substr($val, 2);
-		if (strtoupper(substr($val, 0, 9)) == 'CKEDITOR.')
-			return $val;
-
-		return '"' . str_replace(array("\\", "/", "\n", "\t", "\r", "\x08", "\x0c", '"'), array('\\\\', '\\/', '\\n', '\\t', '\\r', '\\b', '\\f', '\"'), $val) . '"';
-	}
-}
+<?php
```
WouterSioen: I don't think this should be in here?

dubgeiser: Yeah, just replaced everything... is it bad?  I mean, it shouldn't pose any real problems, and when CK editor is updated, this will be gone...

WouterSioen:
@dubgeiser it seems like this breaks the test. Can you check why? Test output can be found here: https://travis-ci.org/forkcms/forkcms/jobs/52298999

dubgeiser:
@WouterSioen my bad, went over this too quickly, inheriting classes cannot access private properties, poc:
```php
class Fubar
{
	private static $fubar = 'test';

	public static function testIt()
	{
		return static::$fubar;
	}
}


class Bobo extends Fubar
{
}


echo Bobo::testIt();

```
I'll try to fix this.

dubgeiser:
The more I think about, the more I think this is more trouble than it is worth...  anybody else got the same feeling?

WouterSioen:
Can you explain why? I have the feeling it could be useful in some cases to be able to extend a class and be able to use late static binding. This is for example a gist where I use late static binding a lot and it's really useful: https://gist.github.com/WouterSioen/e351c5a71b8d6b93e628

dubgeiser:
@WouterSioen I agree completely in the sense that late static binding can be useful, don't get me wrong ;-)
But, in Fork, where all but a few class properties are private (so static:: cannot be used to access them) + the static private methods that can be binded + the fact that there is pretty much no OO design in Fork (sorry if I offend anyone) makes it seem that going through everything to check visibility on a per class hierarchy base is not worth the effort...
However if the community deems this absolutely necessary (by supplying real-life use cases _in_ Fork), I'll be happy to do it.

WouterSioen:
I'll close this for now.

PR #1098, opened 2015-02-26
Issue 751

carakas:
The facebook-php-sdk was deprecated so it needed to be removed
Its been removed completely from fork since you can't just get an access-token with the facebook-php-sdk-v4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,6 @@
     "license": "MIT",
     "require": {
         "php": ">=5.3.2",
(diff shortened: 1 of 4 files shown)


WouterSioen:
@tijsverkoyen you know more about the Facebook integration, can you check this?

WouterSioen:
That nasty red notice in composer is gone, nice!

carakas:
:tada:

PR #1099, opened 2015-02-26
Redirect to the first allowed module and first allowed action

### LINKED ISSUE #550: Redirect in the BackendAnalyticsConfig can cause a redirect loop
If you go to `/private` without adding a module and action, Fork will search for an allowed module. In my case the first allowed module was analytics, so it redirects to /private/en/analytics.

Fork will now search for the default action of the analytics module. This is done by making a new instance of `BackendAnalyticsConfig`. The constructor of this config class checks if some settings are available. In my case these settings were not available (I guess this is just because no analytics account was connected).

The config will now check if we're currently on the settings or index action. Normally it should say we're on the index. The config checks this by using `Spoon::get('url')->getAction()`. But at that moment, probably because the querystring is still being processed, getAction returns `null`.

The config will now redirect us to `BackendModel::createURLForAction('index')`. Since no module and no language are set yet (for the same reason as above I guess), the URL we get is `/private///index`. So no module and action is set in the URL, so the complete process will now start over again, resulting in a redirect loop.

Landerstraeten:
This commit closes #550


src/Backend/Core/Engine/Authentication.php
```diff
@@ -192,7 +192,7 @@ public static function isAllowedAction($action = null, $module = null)
     {
         // always allowed actions (yep, hardcoded, because we don't want other people to fuck up)
         $alwaysAllowed = array(
-            'Dashboard' => array('Index' => 7),
+            //'Dashboard' => array('Index' => 7),
```
WouterSioen: no need to commit commented code.

src/Backend/Core/Engine/Authentication.php
```diff
@@ -280,6 +280,7 @@ public static function isAllowedModule($module)
         $modules = BackendModel::getModules();
         $alwaysAllowed = array('Core', 'Error', 'Authentication');
         $module = (string) $module;
+        $module = \SpoonFilter::toCamelCase($module);
```
WouterSioen: Can't you just cast the $module var to string in the toCamelCase call, instead of having two lines?

src/Backend/Core/Engine/Url.php
```diff
@@ -100,7 +101,7 @@ private function processQueryString()
         }
 
         // split into chunks, a Backend URL will always look like /<lang>/<module>/<action>(?GET)
-        $chunks = (array) explode('/', trim($processedQueryString, '/'));
+        $chunks = (array)explode('/', trim($processedQueryString, '/'));
```
WouterSioen: why did you remove the spaces after the casts?

src/Backend/Core/Engine/Authentication.php
WouterSioen: I don't even think this should be changed. I think only the changes in Authentication::processRegularRequest are needed to fix the issue.

PR #1100, opened 2015-02-26
Replace SPOON_DEBUG with kernel.debug

### LINKED ISSUE #812: Refactor usage of SPOON_DEBUG
Everything relying on SPOON_DEBUG should use the parameters

megasnort:
SPOON_DEBUG is now replaced with kernel.debug everywhere except in one place: Kernel.php where it is used in combination with SPOON_DEBUG_EMAIL.
Once the SwiftMailer is implemented for, SPOON_DEBUG_EMAIL can be removed and then SPOON_DEBUG can be a goner too. Closes #812

### DIFF
--- a/UPGRADE_3.9.md
+++ b/UPGRADE_3.9.md
@@ -25,3 +25,10 @@ use it like this:
 
 All occurences of sending emails in the core are replaced by this, so you can use
 this as a reference.
+
+
+## SPOON_DEBUG
+
+SPOON_DEBUG is removed. From now on you need to check if DEBUG is on by using the kernel.debug parameter, f.e.
+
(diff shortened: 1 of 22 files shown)


WouterSioen:
Wouldn't you just remove the SPOON_DEBUG constant?

megasnort:
I'm not sure I I understand your question completely. Do you mean why I didn't remove the last usage of SPOON_DEBUG in Kernel.php? Because the SPOON_DEBUG_EMAIL that is initialized using SPOON_DEBUG is still used in a few places (Which will be resolved shortly i heard). I wanted to make sure that this pull request alone wouldn't break anything.

tijsverkoyen:
Don't forget to update the UPGRADE_xx.md file.

PR #1104, opened 2015-02-26
Tags: Remove empty elements from array

### LINKED ISSUE #1103: Blog/tags array_fill exception
Just did a fresh git clone of Fork CMS and installed it. I went to the blog module and chose to edit one of the two installed example articles. When I save it, I get the following exception:

	ContextErrorException: Warning: array_fill(): Number of elements must be positive in /home/vagrant/apps/default/fork-cms/forktest/src/Backend/Modules/Tags/Engine/Model.php line 297

	in /home/vagrant/apps/default/fork-cms/forktest/src/Backend/Modules/Tags/Engine/Model.php line 297
	at ErrorHandler->handle('2', 'array_fill(): Number of elements must be positive', '/home/vagrant/apps/default/fork-cms/forktest/src/Backend/Modules/Tags/Engine/Model.php', '297', array('otherId' => '1', 'tags' => array(), 'module' => 'Blog', 'language' => 'en', 'db' => object(SpoonDatabase), 'currentTags' => array(), 'key' => '0', 'tag' => ''))
	at array_fill('0', '0', '?') in /home/vagrant/apps/default/fork-cms/forktest/src/Backend/Modules/Tags/Engine/Model.php line 297
	at Model::saveTags('1', '', 'Blog') in /home/vagrant/apps/default/fork-cms/forktest/src/Backend/Modules/Blog/Actions/Edit.php line 448

Same thing happens when I create a new blog article without tags. Seems like the error happens at:

    $placeholders = array_fill(0, count($tags), '?');

On line 282 there's a check if the tags array is not empty (`if (!empty($tags))`). The installed example articles don't have tags, but I saw in the debugger that the tags array actually isn't empty: `$tags[0] = ""`. Because of this, the array is not empty and the code gets executed while it should not....

megasnort:
Problem was indeed the empty element which caused the empty-function to return false. Closes #1103

### DIFF
--- a/src/Backend/Modules/Tags/Engine/Model.php
+++ b/src/Backend/Modules/Tags/Engine/Model.php
@@ -258,6 +258,9 @@ class Model
         // make sure the list of tags is unique
         $tags = array_unique($tags);
 
+        // remove empty elements
+        $tags = array_filter($tags);
+
         // get db
         $db = BackendModel::getContainer()->get('database');
 
--- a/src/Backend/Modules/Tags/Engine/Model.php
+++ b/src/Backend/Modules/Tags/Engine/Model.php
@@ -246,20 +246,16 @@ class Model
     public static function saveTags($otherId, $tags, $module, $language = null)
     {
         $otherId = (int) $otherId;
-        $tags = (is_array($tags)) ? (array) $tags : (string) $tags;
         $module = (string) $module;
         $language = ($language != null) ? (string) $language : BL::getWorkingLanguage();
 
         // redefine the tags as an array
         if (!is_array($tags)) {
-            $tags = (array) explode(',', $tags);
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Tags/Engine/Model.php
```diff
@@ -258,6 +258,9 @@ public static function saveTags($otherId, $tags, $module, $language = null)
         // make sure the list of tags is unique
         $tags = array_unique($tags);
 
+        // remove empty elements
+        $tags = array_filter($tags);
```
WouterSioen: can't you do this in one line? `$tags = array_filter(array_unique($tags));`

megasnort:
Made it a bit shorter.

PR #1109, opened 2015-02-27
Api test

tijsverkoyen:
Tests for the API! At least the core.xxx methods for now.


tijsverkoyen:
@WouterSioen the major increase of new issues on Scrutinizer is due a configuration-change.

app/phpunit.xml.dist
```diff
@@ -31,6 +31,7 @@
             <directory>../src/*Bundle/Tests/Controller</directory>
             <directory>../src/Frontend/Modules/*/Tests/Actions</directory>
             <directory>../src/Backend/Modules/*/Tests/Actions</directory>
+            <directory>../src/Api/*/Tests/Action</directory>
```
WouterSioen: This should be added to the excludes in the unit testsuite

src/Api/V1/Init.php
```diff
@@ -59,17 +59,17 @@ public function initialize($type)
...
-        define('BACKEND_CACHE_PATH', BACKEND_PATH . '/Cache');
-        define('BACKEND_CORE_PATH', BACKEND_PATH . '/Core');
-        define('BACKEND_MODULES_PATH', BACKEND_PATH . '/Modules');
-
-        define('FRONTEND_PATH', PATH_WWW . '/src/Frontend');
-        define('FRONTEND_CACHE_PATH', FRONTEND_PATH . '/Cache');
-        define('FRONTEND_CORE_PATH', FRONTEND_PATH . '/Core');
-        define('FRONTEND_MODULES_PATH', FRONTEND_PATH . '/Modules');
-        define('FRONTEND_FILES_PATH', FRONTEND_PATH . '/Files');
+        defined('API_CORE_PATH') || define('API_CORE_PATH', PATH_WWW . '/' . APPLICATION);
```
WouterSioen: We should think about putting these in one place, since this is the same in the frontend as in the backend.

I don't really think these are the constants we should remove the fastest, since they are always the same in every Fork application. (so they are really constant)

src/Api/V1/Tests/Action/CoreTest.php
```diff
@@ -0,0 +1,237 @@
+<?php
+
+namespace Api\V1\Tests\Action;
+
+use Common\ApiTestCase;
+
+class CoreTest extends ApiTestCase
+{
+    public function testErrorOutput()
```
WouterSioen: I would give this a more vebose name. Something like "testApiGivesErrorWithoutParameters"

src/Api/V1/Tests/Action/CoreTest.php
```diff
@@ -0,0 +1,237 @@
...
+        $this->assertEquals(
+            '400',
+            $request->filter('fork')->attr('status_code')
+        );
+        $this->assertEquals(
+            'error',
+            $request->filter('fork')->attr('status')
+        );
+        $this->assertNotEmpty(
+            $request->filter('fork')->attr('version')
```
WouterSioen: maybe you could check if this is equal to the content of the VERSION.md file?

WouterSioen:
Great addition! :+1: 

I've added some minor feedback.

app/phpunit.xml.dist
tijsverkoyen: Fixed in df6b7bc34dd294ed6adff0c7155efb9e93f156a2

src/Api/V1/Init.php
tijsverkoyen: I know, but this is something that should be in a separate PR

src/Api/V1/Tests/Action/CoreTest.php
tijsverkoyen: Absolutely correct, I gave all methods a more verbose name in 2424b667b65cdbf19608ad3aba9a3da9706a7072

src/Api/V1/Tests/Action/CoreTest.php
tijsverkoyen: Fixed in 3faecb2690956d7f2c3bb780efc8d28a53a949c4

tijsverkoyen:
And I processed the feedback ;-)

src/Api/V1/Init.php
WouterSioen: It was just some thought I had while reviewing the code ;)

PR #1111, opened 2015-03-02
Uploaded module zip filenames more reliably scanned for unexpected prefi...

nosnickid:
Running the legit command:

zip -r MyModule.zip src/Frontend/Modules/MyModule/ src/Backend/Modules/MyModule/

Does not include the src/ folder as a "dangling folder" entry, as was previously expected by the module zip uploader. 

This new method of extracting a prefix is more reliable under differing zip formats.


WouterSioen:
Looks good to me! :+1:

WouterSioen:
I've tested this and the file you can download here: https://github.com/WouterSioen/moduleMaker/archive/master.zip doesn't work anymore. This worked before these changes. Can you please look into this?

PR #1124, opened 2015-03-09
Feature: Confirmation mail for FormBuilder. Closes #1053

### LINKED ISSUE #1053: Formbuilder confirmation mail
I'm have been working with the formbuilder a couple of times, for different projects and a question I often get is to add a confirmation mail after submitting the form.

So, I have implemented this feature in a couple of websites already, and I thought it might be a good addition to Fork CMS.

Right now it works like this:
- When an e-mail field was present
- And when it is filled in
- And when the subject and content field in the backend (screenshot) were filled

The confirmation mail is send to the e-mailadres in the (first) e-mail field in the list of fields.

So, two questions:
- Would this be a good addition to Fork?
- Is the implementation ok like this? (I'm thinking a more explicit option checkbox "send confirmation mail" might make the backend a bit more user friendly)

schermafbeelding 2015-02-04 om 08 10 50

megasnort:
Here an attempt to provide an (optional) confirmation mail. 

- The confirmation-mail will be send to the reply-to e-mail field, if given.
- Only Dutch and English locale is provided
- A separate template is provided which contains, next to the confirmation message, the filled in fields that are also present in the other template that was already present in the module. If you want to exclude these fields, you can just modify the new template (Confirmation.tpl) and leave the fields out.


jessedobbelaere:
Just tried this and it worked and looks good :+1: 

[Confirmation](http://i.imgur.com/X8tiU8c.png)

WouterSioen:
This can only be merged in the 3.10 release, since this contains database changes.

src/Backend/Modules/FormBuilder/Actions/Add.php
```diff
@@ -55,6 +55,11 @@ private function loadForm()
         $this->frm->addText('email');
         $this->frm->addText('identifier', BackendFormBuilderModel::createIdentifier());
         $this->frm->addEditor('success_message');
+
+        // fields for optional confirmation mail
+        $this->frm->addCheckbox('mail_send');
```
WouterSioen: looking at this again, I think these variable names could be better. I'd use "send_confirmation_mail", "confirmation_mail_subject" and "confirmation_mail_content". This will make it all a little cleaner.

WouterSioen:
Can you do this PR again to the 3.10 branch? We will release this together with 3.10, so it belongs in there.

PR #1125, opened 2015-03-10
Handle formbuilder submitted event with Symfony's event dispatcher.

WouterSioen:
Note that I'll implement the event dispatcher for other modules too, but I'd like feedback on this first. Things that bother me right now:

* ~~all listeners are subscribed in the main config.yml file => they should probably be moved to module specific yml files?~~
* I left the old `FrontendModel::triggerEvent` line. Can this be removed (or at least deprecated)?
* class names of the event subscribers are rather long.

Other feedback is really welcome too.

### DIFF
--- a/src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedMailSubscriber.php
+++ b/src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedMailSubscriber.php
@@ -7,15 +7,26 @@ use Frontend\Core\Engine\Language as FL;
 use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Modules\FormBuilder\Event\FormBuilderSubmittedEvent;
 
+/**
+ * A Formbuilder submitted event subscriber that will send an email if needed
+ *
+ * @author Wouter Sioen <wouter@sumocoders.be>
+ */
 class FormBuilderSubmittedMailSubscriber
 {
     protected $mailer;
 
+    /**
+     * @param Swift_Mailer $mailer
+     */
     public function __construct(Swift_Mailer $mailer)
     {
         $this->mailer = $mailer;
     }
 
+    /**
+     * @param FormBuilderSubmittedEvent $event
+     */
     public function onFormSubmitted(FormBuilderSubmittedEvent $event)
     {
         $form = $event->getForm();
@@ -59,6 +70,12 @@ class FormBuilderSubmittedMailSubscriber
         }
     }
 
+    /**
+     * Converts the data to make sure it is nicely usable in the email
+     *
+     * @param  array $data
+     * @return array
+     */
     protected function getEmailFields($data)
     {
         return array_map(
--- a/src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedMailSubscriber.php
+++ b/src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedMailSubscriber.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Frontend\Modules\FormBuilder\EventListener;
+
+use Swift_Mailer;
+use Frontend\Core\Engine\Language as FL;
+use Frontend\Core\Engine\Model as FrontendModel;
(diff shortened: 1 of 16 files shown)


WouterSioen:
@forkcms/hackday @forkcms/core-contributors Can you please provide feedback and/or answers to my questions?

dennisdegreef:
How come the `FormBuilderEvents` class is made final, and the other new classes aren't? Is there a certain reason for this?

turanct:
@dennisdegreef In my opinion, newly created classes can almost always be made `final`, because most of the time, inheritance is not the way to go for extension points (using inheritance as extension point will in most cases violate the Liskov Substitution principle), there are countless other ways to do that. For those exceptions where inheritance is the only option, removing the `final` keyword is only one commit away.

So to answer your question, in fork, there's massive use of inheritance. It's a relic from the past. I'm pretty sure @WouterSioen and others are trying to slowly get rid of that and clean things up. That's why `final` would be present here. As to why it's not there for some other classes: i don't know. Maybe look into that @WouterSioen?

For example `FormBuilderSubmittedEvent` states it's an immutable event. It's not final though, so i could extend it and add setters to make it mutable. Other parts of the program would still treat it as immutable, but it's not! :skull:

dennisdegreef:
@turanct I totally agree with promoting the `final` keyword :) The fact that I couldn't figure out why not all new classes were made final was what made me curious :)

src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedMailSubscriber.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function onFormSubmitted(FormBuilderSubmittedEvent $event)
+    {
+        $form = $event->getForm();
+
+        // need to send mail
+        if ($form['method'] == 'database_email') {
+            // build our message
+            $from = FrontendModel::getModuleSetting('Core', 'mailer_from');
+            $message = \Common\Mailer\Message::newInstance(sprintf(
```
dennisdegreef: Instead of creating a hardcoded dependency here, would it be better to put it in a service so you can inject it?

dennisdegreef:
About the event subscriber class names being too long, It would be a matter of taste, so the project's decision. I would probably be as descriptive as possible on a concrete class, in which I would be okay with long class names. But i'm not one of the project maintainers :) Just my 2c.

As for all the even listeners in the main config, If you are viewing the module as something you would share among other projects, the question becomes, are the subscribers performing logic that you would want to have in all implementations? If they are all valid in this usecase only, I would leave them up to implement for the user (the application implementing the module) and thus leave them in the main yml.

src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedMailSubscriber.php
WouterSioen: I think creating the message instance is in fact the responsibility of a service that is sending an email. The mailer itself is injected using constructor injection though.

WouterSioen:
I think it would indeed be a great idea to have different services.yml files per module, that will only be loaded if the module is installed. I'll check if I'll be able to implement such thing.

Thanks for the feedback! I've made the other classes final too. No need to be able to overwrite these event subscribers (for now).

src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedMailSubscriber.php
dennisdegreef: You are right, I clearly misread what it actually did and was thrown off that `newInstance` was something like `getInstance`, indicating a singleton.

dennisdegreef:
Thank you for being the first _real_ user of @lonelypullrequests :dancer:

WouterSioen:
I've updated the pull request to allow module specific container extensions. This way, every module can add it's own services to the container (or add configuration options).

WouterSioen:
The tests work locally but not on travis :(

WouterSioen:
They work again!

PR #1129, opened 2015-03-13
Fx Blog: Revision images get deleted at the right moment. Closes #1114

### LINKED ISSUE #1114: Blog: image file gets deleted
Funny things happen when images are added to different versions of blogposts. I already have a solution (already implemented in a project), but want to check if you agree:

**case 1**
Problem: When upload a new picture, the previous one is deleted. When you reactivate a previous version of the blogpost, the image is not present anymore (but is in the database)
Solution: just don't delete the previous image

**case 2**
Problem: When you don't upload a new picture, the existing picture gets renamed. When you then reactivate a previous post, the image is not present (because is was renamed)
Solution 1: Don't rename the picture. The small price to pay is that the SEO (filename) of this image might be a little bit less relevant.
Solution 2: The other option is the copy to file every time the newly generated filename is different

**case 3**
Problem: When you check 'delete image', the image is deleted. If the image was used in previous versions, it isn't available there anymore.
Solution: just don't delete the previous image

Bottom line: The image should only be deleted when the post itself gets deleted.

megasnort:
This request fixes/modifies editing and deletion of blog posts:
- files of previous revisions are never overwritten or deleted
- files of revisions are deleted when the revision is deleted
- files of revisions are deleted then the blog post gets deleted

Some small modifications to the existing code.

- implode of placeholders now happens just once instead of every time they are used
- add a check for an empty array of ids because this would generate an error when using the placeholders
- get the db-container just once

### DIFF
--- a/src/Backend/Modules/Blog/Actions/Delete.php
+++ b/src/Backend/Modules/Blog/Actions/Delete.php
@@ -55,9 +55,6 @@ class Delete extends BackendBaseActionDelete
             // delete item
             BackendBlogModel::delete($this->id);
 
-            // delete the image
-            BackendModel::deleteThumbnails(FRONTEND_FILES_PATH . '/blog/images', $this->record['image']);
-
             // trigger event
             BackendModel::triggerEvent($this->getModule(), 'after_delete', array('id' => $this->id));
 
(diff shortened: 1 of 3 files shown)


WouterSioen:
looks good imo :+1:

PR #1133, opened 2015-03-18
Page Template: {option:isPageXChild} added

jeroendesloovere:
## Problem

When having a page tree structure with parents containing children.
You sometimes need to add extra elements for pages for a certain parent.

## Solution

You can now use the following if you want pages with a parent_id == 30 to have an extra paragraph:
```html
{option:isPage30Child}
<p>I'm a child for the parent page with ID = 30</p>
{/option:isPage30Child}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 3.9.2 (2015-xx-xx)
 --
 Improvements:
+* Core: every template can now check if it has a certain parent id with {option:isPage1Child}
 
 Bugfixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 3.9.2 (2015-xx-xx)
 --
 Improvements:
-* Core: every template can now check if it has a certain parent id with {option:isPage1Child}
+* Core: every template can now check if it has a certain parent id with {option:isChildOfPage1}
 
 Bugfixes:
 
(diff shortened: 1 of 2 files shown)


WouterSioen:
Do you have an example use case to convince me? :wink:

jeroendesloovere:
Sure, lots of examples can be given.
Lets assume the following page tree:
```
Home
> Services

>> Online (page id = 30)
>>> Websites
>>> Applications
>>> Newsletters

>> Offline
>>> Bannering
>>> Press
```
All pages (online, websites, offline, ...) have the same template layout.
But on the children for the "online" page I want something special to be seen, which is not editable by the customer and needs to be hardcoded.

This was impossible! But now you can do it as follows:

```html
# Frontend template
<div class="item {option:isPage30Child} zero-padding-top{/option:isPage30Child}">
    {option:isPage30Child}
        {include:Core/Layout/Templates/VerySpecificContentWithComplexLayoutAndIsNotEditable.tpl}
    {/option:isPage30Child}
</div>
```

WouterSioen:
Can you change the name of the new variable to "isChildOfPage\<number\>"?

jeroendesloovere:
@WouterSioen Done.

WouterSioen:
I'll squash and merge this.

WouterSioen:
Merged in 635c184725d6da1c90bda6399e736649300c01e6

PR #1135, opened 2015-03-18
Fixed add landing page

jeroendesloovere:
Steps te reproduce
* Set up Analytics API key
* Click ```add landing page``` (you see no dropdown containing link items)
* When you now enter something and click "validate", you get an error in validateForm(), thats because the field 'page_list' does not exists.

## Solution

This solution solves two bugs
* ```$pageList``` was not set at rule 95-96
* field 'page_list' does not exists at rule 114

### DIFF
--- a/src/Backend/Modules/Analytics/Actions/AddLandingPage.php
+++ b/src/Backend/Modules/Analytics/Actions/AddLandingPage.php
@@ -113,7 +113,7 @@ class AddLandingPage extends BackendBaseActionAdd
                 $this->frm->getField('page_path')->addError(BL::err('FieldIsRequired'));
             }
             if (!$this->frm->getField('page_path')->isFilled()
-                && !empty($this->linkList)
+                && $this->frm->existsField('page_list')
                 && !$this->frm->getfield('page_list')->isFilled()
             ) {
                 $this->frm->getField('page_path')->addError(BL::err('FieldIsRequired'));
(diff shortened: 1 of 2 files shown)


src/Backend/Modules/Analytics/Actions/AddLandingPage.php
```diff
@@ -111,7 +112,10 @@ private function validateForm()
             if (!isset($page)) {
                 $this->frm->getField('page_path')->addError(BL::err('FieldIsRequired'));
             }
-            if (!$this->frm->getField('page_path')->isFilled() && !$this->frm->getfield('page_list')->isFilled()) {
+            if (!$this->frm->getField('page_path')->isFilled()
+                && !empty($this->linkList)
```
WouterSioen: You can als ouse `$this->frm->hasField('page_path')`. This seems more logical to me (since this doesn't cause duplicate code in the form of the check on the $this->linkList variable.

jeroendesloovere: The function ```hasField()``` does not exists.

WouterSioen: I meant `existsField`, sorry

WouterSioen:
This will be fixed in our brand new analytics module that I'll make!

PR #1144, opened 2015-03-23
Adds 'copy to' to form builder fields

jonasdekeukelaere:


WouterSioen:
looks good to me

src/Frontend/Modules/FormBuilder/Widgets/Form.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Frontend\Modules\FormBuilder\Widgets;
 
+use Common\Mailer\Message;
```
jeroendesloovere: This 'use' can be removed. Since we get the class by using ```\``` in ```\Common\Mailer\Message```;

jonasdekeukelaere: Fixed in efc202d

WouterSioen:
We can't merge this since sending mail is now done in an event. Can you change your pr?

jonasdekeukelaere:
See #1183

PR #1158, opened 2015-04-08
Speed improvements for BackendDataGridFunctions::getUser()

jeroendesloovere:
Reduces the number of database calls drastically.


src/Backend/Core/Engine/DataGridFunctions.php
```diff
@@ -19,6 +19,8 @@
  */
 class DataGridFunctions
 {
+    protected static $dataGridUsers;
```
WouterSioen: you should make this an array, since you'll be adding items to it in the getUser method.

jeroendesloovere: @WouterSioen done.

PR #1159, opened 2015-04-08
Rename blockIsHTML to blockIsEditor.

### LINKED ISSUE #1147: blockIsHtml variable should be named blockIsEditor
This would reflect the difference between true and false more. Modules and widgets mostly return html too, so isHtml isn't really a good name.

dubgeiser:
blockIsEditor is better naming.
Backwards compatibility is preserved by equaling blockIsHTML to the
value of blockIsEditor.

Closes #1147.

### DIFF
--- a/src/Frontend/Core/Engine/Page.php
+++ b/src/Frontend/Core/Engine/Page.php
@@ -369,9 +369,12 @@ class Page extends FrontendBaseObject
                         // parse extra
                         $positions[$position][$i] = array(
                             'variables' => $block['extra']->getTemplate()->getAssignedVariables(),
-                            'blockIsHTML' => false,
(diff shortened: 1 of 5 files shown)


WouterSioen:
This looks rather similar to my PR: https://github.com/forkcms/forkcms/pull/1151 :wink:

dubgeiser:
hmz... the "issue" was open and unassigned when I picked it up.
Although I think this ("mine") solution is more DRY, I'll leave it up to you guys to decide which one to merge ;-)
The Scrutinizer fail surely wasn't caused by this code?!

WouterSioen:
I closed mine in favour of your implementation.

PR #1163, opened 2015-04-16
Bugfix in Add/Edit Groups - not-installed module widgets should not be executed

jeroendesloovere:
## Problem
When adding/editing a Group and you try to add widgets for modules that aren't installed.
The widgets throw errors, because their database tables don't exist.

## Solution
I added a check so we first check if this module is installed. If not, skip this widget from executing.


WouterSioen:
I'll merge this

PR #1187, opened 2015-05-13
Analytics api v3 - Fixes #1173

### LINKED ISSUE #1173: Analytics module <-> Google not working anymore since 20 april
**Problem**

Google has disabled connections from deprecated Google Apps APIs, since 20 april.
So it is impossible to connect the Analytics module using Google API Console server key.
[More info Google gives you](https://support.google.com/a/answer/6201352)

**Solution**

We need to use the new Google API and AUTH 2.0

WouterSioen:
Make a minimal version of the analytics module that works with the new Google API.


WouterSioen:
Note that it does have a lot less functionality then the previous version of the analytics module. It's rather clean though, so adding extra graphs/... should be easy. It now contains one dashboard and the functionality to couple an analytics profile.

screen shot 2015-05-13 at 17 38 51

WouterSioen:
@todo: make the code cleaner by injecting the module settings in the services. This will make it easier to write tests for them.

tijsverkoyen:
:+1:

PR #1193, opened 2015-05-14
Bugfix 1172 check all languages installed

carakas:


WouterSioen:
For some reason, scrutinizer couldn't run. I'm retrying the run. It looks good at first sight though, so I think we'll be able to merge this ;)

WouterSioen:
Looks like our continious integration passed now! I'd write 'intl' in the extension_loaded with a small letter i though. This is more consistent with the other extension_loaded checks. The documentation of extension_loaded says it's case insensitive, so it can't cause any problems.

PR #1196, opened 2015-05-15
Refactor the modules settings to live in a service.

WouterSioen:
We now have decent dependency injection in our modules settings! We are
now able to inject the modules settings service in other services and
mock it in tests, yay!

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -63,10 +63,10 @@ services:
         arguments:
             - @database
             - @service_container
-    modules_settings_cache:
+    fork.settings_cache:
         class: Common\Cache\InMemoryCache
-    modules_settings:
+    fork.settings:
         class: Common\ModulesSettings
         arguments:
             - @database
-            - @modules_settings_cache
+            - @fork.settings_cache
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -63,3 +63,10 @@ services:
         arguments:
             - @database
             - @service_container
+    modules_settings_cache:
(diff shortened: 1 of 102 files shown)


jeroendesloovere:
:+1:

tijsverkoyen:
:+1:

WouterSioen:
All modules (and the core) have been changed to use the new fork.settings service. Using the old methods will trigger an E_USER_DEPRECATED notice.

WouterSioen:
If people want to check the code without all the clutter (changing this everywhere), these are the relevant commits: d92a7222a27af4985a5e5764dd74254cee439944 and c171aae34517257ba67ff89a0abe370b608711aa

tijsverkoyen:
:+1:

GhaziTriki:
This is a such little a good refactoring but after some testing I found that it is impacting performances. On a local website with dev env and debug set to 1 to memory usage goes from 7.0 MB to 8.2 MB. Any clue about how to lower the memory usage down?

WouterSioen:
I think this is strange, because I refactored the modules settings to only be cached in memory in one place. In the past, it was possible that they were saved in both front and backend at the same time.

I don't think a memory usage of 7.5 MB is a problem? I've checked some frameworks (like Laravel, Symfony, Yii, ...) and they all use more memory then this.

If you want to lower your memory, you can try it by using a newer PHP version. From PHP5.5 and higher, php contains an opcode cache, which will reduce memory usage a lot. There are also some other changes in the PHP sourcecode to improve memory usage and garbage collection.

PR #1183, opened 2015-05-09
Feature formbuilder cc

jonasdekeukelaere:
First merge #1182

PR #1197, opened 2015-05-19
Fix "Illegal offset type" error on setReplyTo method

jessedobbelaere:
The setReplyTo method uses an array with $email as key and value. The variable $email is however an array containing: 'id' and 'value'. 

Because of this, the error "Illegal offset type" is thrown on line 62. We need a string here, so make sure to fetch the value from $email

After fixing this, my form works again perfectly.

Did it work properly in #1182 ?


jonasdekeukelaere:
Looks good now. My mistake, Somehow something went wrong copying it from my other PR https://github.com/forkcms/forkcms/pull/1183/files#diff-81aa05b9a7c704fbc2e17fe3497462e5R67

WouterSioen:
Thanks!

PR #1202, opened 2015-05-19
Only inject the modulesSettings in the configurator.

WouterSioen:
This has two benifits:

* We have a database call less (modules settings are cached  in the fork.settings service after 1 call).
* Our Configurator class has a responsibility less, it does not know where and how  the moduleSettings are saved.

I've also added some tests, which enabled me to cleanup the code a little more.

### DIFF
--- a/src/Common/Mailer/Configurator.php
+++ b/src/Common/Mailer/Configurator.php
@@ -3,13 +3,12 @@
 namespace Common\Mailer;
 
 use PDOException;
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Common\ModulesSettings;
 
-class Configurator implements EventSubscriberInterface
+class Configurator
 {
     /**
      * @var ModulesSettings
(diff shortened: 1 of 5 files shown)


src/Common/Mailer/Configurator.php
```diff
@@ -2,64 +2,46 @@
...
+    private $modulesSettings;
+
+    /**
+     * @var ContainerInterface
+     */
     private $container;
-    private $settings;
 
-    public function __construct($database, $container)
+    public function __construct(ModulesSettings $modulesSettings, ContainerInterface $container)
```
jeroendesloovere: Can you multiline this please, > 80 characters.

src/Common/Tests/Mailer/ConfiguratorTest.php
```diff
@@ -0,0 +1,105 @@
...
+            ->expects($this->once())
+            ->method('set')
+            ->with(
+                $this->equalTo('swiftmailer.mailer.default.transport'),
+                $this->isInstanceOf('\Swift_MailTransport')
+            )
+        ;
+
+        $configurator->onKernelRequest($this->getGetResponseEventMock());
+    }
```
jeroendesloovere: Can you add an empty rule here please.

src/Common/Mailer/Configurator.php
WouterSioen: > There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less.

-- PSR2

I think it makes sense here to keep it on one line, since it's not fully required to make this less then 80 chars. (Multiline constructors look so messy!)

src/Common/Tests/Mailer/ConfiguratorTest.php
WouterSioen: Fixed in the new version of the same commit :)

PR #1203, opened 2015-05-20
Bugfix analytics module when using deployment

jeroendesloovere:
BACKEND_CACHE_PATH worked. But when using capistrano deployment, every
time you deploy the cache (= secret json file) was deleted. So the file couldn't be
found and a Google Client Secret File error was thrown.


jeroendesloovere:
Closing due to better solution.

PR #1204, opened 2015-05-20
Bugfix analytics secret file

jeroendesloovere:


src/Backend/Modules/Analytics/Actions/Settings.php
```diff
@@ -160,14 +160,11 @@ private function validateSecretFileForm()
         }
 
         if ($this->form->isCorrect()) {
-            $fileField->moveFile(
-                BACKEND_CACHE_PATH . '/' . $this->getModule() . '/'
-                . $fileField->getFileName()
-            );
+            $authConfigFileContent = \SpoonFile::getContent($fileField->getTempFileName());
```
WouterSioen: It's not really necessary to put this in a variable, you can just do the getContent in the fork.settings->set method. I'd also use file_get_contents instead of SpoonFile::getContent, since we're trying to remove Spoon from our stack.

WouterSioen:
:+1: Looking good!

WouterSioen:
Merged in 00ac6abdea7601071481c2ca3a8920b393a6c3e7

I've squashed the PR because it was a lot of commits for not that many changes. This is better to not clutter our git history too much.

PR #1206, opened 2015-05-21
Add some headers to responses for increased security.

WouterSioen:
see https://www.owasp.org/index.php/List_of_useful_HTTP_headers for some more info


tijsverkoyen:
:+1:

PR #1208, opened 2015-05-22
Minor updates on PHP documentation.

GhaziTriki:
It should slightly improve the code quality.

### DIFF
--- a/src/ForkCMS/Bundle/InstallerBundle/Tests/Controller/InstallerControllerTest.php
+++ b/src/ForkCMS/Bundle/InstallerBundle/Tests/Controller/InstallerControllerTest.php
@@ -28,7 +28,7 @@ class InstallerControllerTest extends WebTestCase
     {
         $client = static::createClient();
 
-        // make sure we have a clean slate and our parametersfile is backed up
+        // make sure we have a clean slate and our parameters file is backed up
         $this->emptyTestDatabase($client->getContainer()->get('database'));
         $this->backupParametersFile($client->getContainer()->getParameter('kernel.root_dir'));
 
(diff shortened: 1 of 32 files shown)


WouterSioen:
:+1:

src/ForkCMS/Bundle/InstallerBundle/Tests/Controller/InstallerControllerTest.php
```diff
@@ -44,7 +44,8 @@ public function testInstallationProcess()
 
     /**
      * @param \Symfony\Component\DomCrawler\Crawler|null $crawler
-     * @param \Symfony\Bundle\FrameworkBundle\Client $client
+     * @param \Symfony\Bundle\FrameworkBundle\Client|null $client
```
WouterSioen: Why the `|null` part? (in both crawler and client?) We always expect a Crawler and a Client object. With null, our methods won't work.

WouterSioen: Also in the other methods in this class

WouterSioen: @GhaziTriki can you fix this?

GhaziTriki:
Another commit is coming in next hour for more improvements.

WouterSioen:
great!

GhaziTriki:
It should be the last commit for this pull request. I have noticed that there a lot of PSR-2 incompatibilities after running a fix with php-cs-fixer. Are you happy for me to create open an issue to discuss about rules to consider?

WouterSioen:
Yes, better PSR2 compliance is always welcome!

PR #1211, opened 2015-05-27
Increasing security in API when user is "God"

jeroendesloovere:
Even for "God" we have to verify the access_token.

**Problem**
When a "God" user connects for the first time with his email address, he will be given an "api_key" in the database. But on the next calls which require authentication, that api key will never be checked in the API, because we return true before even checking it.

So if any other person knows the email address of "God", no matter what API key he fills in, he can be able to access anything in the API.

**Solution**
Remove the "God" check. So every authenticated call needs to check the given api_key.


jeroendesloovere:
Can you check this out, @tijsverkoyen

PR #1212, opened 2015-05-27
Make composer create-project the default installation method.

WouterSioen:


tijsverkoyen:
:+1:

WouterSioen:
I would even change this on the Fork CMS website. Using and uploading a Zip Archive should be a fallback method in my opinion, and should not be encouraged as the default method.

tijsverkoyen:
ok.

PR #1214, opened 2015-05-27
Inverted Backend import/export icons

### LINKED ISSUE #1213: Import and export icons direction inverted
Hi Guys,

Looking at the export and import icons direciton in "Backend->Settings->Translations" I have a strange feeling. They don't look right, normally the import should have the arrow down because we are pulling data, and an arrow up for the export because we are pushing data out.

jeroendesloovere:
Fixes #1213

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Bugfixes:
 
 * Analytics: Auth config content saved in databse, because capistrano deployments didn't work with BACKEND_CACHE_PATH.
 * Api: increasing security when user is GOD.
+* Core: Inverted backend import/export icon.
 
 
 3.9.2 (2015-05-12)
--- a/src/Backend/Core/Layout/Css/imports/buttons.css
+++ b/src/Backend/Core/Layout/Css/imports/buttons.css
@@ -131,9 +131,9 @@ a.icon span {
 	a.iconUsers span {
 		background-position: 0 -264px !important; }
 	a.iconExport span {
-		background-position: 0 -374px !important; }
-	a.iconImport span {
 		background-position: 0 -396px !important; }
+	a.iconImport span {
+		background-position: 0 -374px !important; }
 	a.iconStats span {
 		background-position: 0 -420px !important; }
 	a.iconGoto span {


WouterSioen:
I don't agree here. I think the icons are logical the way they are now. An export has a down arrow in my opinion, since it **down**loads stuff for you, and an import has an up arrow since you'll **up**load stuff.

jeroendesloovere:
@WouterSioen I follow what @GhaziTriki said.
Because in the history I've also found the icons confusing...

This is how it was: 
schermafbeelding 2015-05-27 om 16 26 19

This is how it will be:
import-export-after

Examples in real life: Apple
They also use an arrow that points up for their "exporting" features.
schermafbeelding 2015-05-27 om 16 29 25

jeroendesloovere:
I've been an Mac user for a decade now and for me the "up" arrow is similar to "export/share/...". But I still remember that IE and Firefox for Windows use a "down" arrow for downloads...

WouterSioen:
http://www.fork-cms.com/ also uses a down arrow for download :)

jeroendesloovere:
@WouterSioen Yes, and I also found it obvious on that particular location that the arrow is pointing down...

jeroendesloovere:
I find that there's a distinction between "export" and "download". For download I would use "down" arrow en for export an up arrow... Strange shizzle...

WouterSioen:
In my opinion, an export is a download of some data in a certain format.

jeroendesloovere:
Endless discussion I guess.
Maybe we can add a "checkbox" in the core, so users can choose their preferred solution. :-P LOL

WouterSioen:
Please, let's not go there :p

GhaziTriki:
I think I will regret for starting this conversation :laughing:, normally in ergonomics if you choose a label then you choose an icon to go well with it. Actually download and export labels are different even if they have the same meaning, the export have the meaning of pulling out something, but downloading have the meaning of pulling in something.

WouterSioen:
I'll check how the export icons look in the bootstrap implementation.

WouterSioen:
Looks like it will be fixed when the bootstrap implementation is merged:

jeroendesloovere:
Thank god its closed :-P :+1:

PR #1218, opened 2015-05-30
Add classname inputfield to formbuilder fields

jessedobbelaere:
Adds a CSS classname field to the advanced panel of each formbuilder field. A small formbuilder feature that allows you to do more with form fields in the frontend. Use the CSS class field to resize fields using a grid classname, float fields next to each other, target fields with javascript selectors, hide a field label, etc.

I didn't add the css class to the header and paragraph elements in formbuilder, because those are too basic. 

Example:
Note that, of course, you can write multiple css classnames such as `.col-md-4 .field-email .js-email`
schermafbeelding 2015-05-30 om 02 41 23


WouterSioen:
Can you PR this to the 3.10 branch, so we'll have this with the bootstrap layout.

PR #1220, opened 2015-05-31
add mime type hinting to FormImage by default

carakas:


WouterSioen:
:+1: I did this a lot in Symfony projects, so I like this :)

PR #1221, opened 2015-06-02
RFC: Identify and fix Authentication issue(s).

### LINKED ISSUE #1175: Backend Bug for User who logs out.
**Problem**

Steps to reproduce:
* Add a new user which has no rights for the **groups** nor **users** module.
* Click **Log out** at the top right button
* Log in as that new user.

Then you see the error.
schermafbeelding 2015-05-05 om 10 28 32

**Solution**

Fixing the method which checks which module to redirect to.

tommyvdv:
* Identify and resolve issues causing unexpected login results.
* Add functionality to the users edit action to allow a user access even without an id parameter or index rights.
* Coming from the login form, the edit action will now edit "self" instead of redirecting to an error page. (d9194b1f200a16b8a5ca9d9e6c900051f8533134)
* Add more tests. (One of which fails due to an unexpected redirect. 7a526089b10d8c96107902bd15f15229192e1b97)

resolves #1175


src/Backend/Core/Engine/Authentication.php
```diff
@@ -427,6 +427,9 @@ public static function loginUser($login, $password)
             \SpoonSession::set('backend_logged_in', true);
             \SpoonSession::set('backend_secret_key', $session['secret_key']);
 
+            // Update/instantiate the value for the logged_in container.
```
WouterSioen: we never start comments with capital letters.

src/Backend/Modules/Authentication/Actions/Index.php
```diff
@@ -268,4 +259,73 @@ private function validateForm()
...
+     * and send the user on his way.
+     */
+    private function redirectToAllowedModuleAndAction()
+    {
+        $allowedModule = $this->getAllowedModule();
+        $allowedAction = $this->getAllowedAction($allowedModule);
+        $this->redirect($this->getParameter('querystring', 'string',
+            $allowedModule ?
+                BackendModel::createUrlForAction($allowedAction, $allowedModule) :
+                BackendModel::createUrlForAction('index', 'Authentication')
```
WouterSioen: 'index' should be written with a capital I

src/Backend/Modules/Authentication/Actions/Index.php
```diff
@@ -268,4 +259,73 @@ private function validateForm()
...
+     * allowed to access.
+     */
+    private function getAllowedAction($module)
+    {
+        if(BackendAuthentication::isAllowedAction('Index', $module)) {
+            return 'Index';
+        }
+        $allowedAction = false;
+        $finder = new Finder();
+        $finder->name('*.php')
```
WouterSioen: Can't you just get the allowed actions out of the db? It's stored in the groups_rights_actions table.

src/Backend/Modules/Authentication/Tests/Actions/IndexTest.php
```diff
@@ -77,5 +78,86 @@ public function testAuthenticationWithCorrectCredentials()
...
+    public function testPagesUserWithCorrectCredentials()
+    {
+        // The authentication class persist the previous user.
+        // In practice this situation will almost never occur.
+        // Logging in one user only to log out and subsequently
+        // loggin in with another seems a bit much without
+        // a redirect in place to refresh the application.
+        // Another way to make this test, test. Would be to insulate
+        // the client in each test.
+        Authentication::tearDown();
```
WouterSioen: Can't you put this in a tearDown method in this class? This way, you don't need this duplicate 8 lines long comment (and you can add it as a method comment this way)

src/Backend/Modules/Authentication/Tests/Actions/IndexTest.php
```diff
@@ -77,5 +78,86 @@ public function testAuthenticationWithCorrectCredentials()
...
+        $crawler = $client->request('GET', '/private/en/authentication');
+        $this->assertEquals(
+            200,
+            $client->getResponse()->getStatusCode()
+        );
+
+        $form = $crawler->selectButton('login')->form();
+        $this->submitForm($client, $form, array(
+            'form' => 'authenticationIndex',
+            'backend_email' => 'users-edit-user@fork-cms.com',
```
WouterSioen: I don't think these credentials are correct. They are not available in the test database. The tests fail with this error: "Your e-mail and password combination is incorrect."

The correct credentials for the user in the testdatabase are user noreply@fork-cms.com with password fork (as tested in the first passing test in this method).

src/Backend/Modules/Authentication/Actions/Index.php
```diff
@@ -39,8 +41,13 @@ class Index extends BackendBaseActionIndex
     public function execute()
     {
         // check if the user is really logged on
-        if (BackendAuthentication::getUser()->isAuthenticated()) {
-            $this->redirect($this->getParameter('querystring', 'string', BackendModel::createUrlForAction(null, 'Dashboard')));
+        if (BackendAuthentication::getUser()->isAuthenticated())
+        {
```
jeroendesloovere: Can you please change this to one line?
```php
if (BackendAuthentication::getUser()->isAuthenticated()) {
```

src/Backend/Modules/Users/Actions/Edit.php
```diff
@@ -57,6 +57,26 @@ class Edit extends BackendBaseActionEdit
...
         $this->id = $this->getParameter('id', 'int');
+        $this->error = $this->getParameter('error', 'string');
+        $this->loadAuthenticatedUser();
+
+        // If id and error parameters are not set we'll assume the user logged in
+        // and has been redirected to this action by the authentication index action.
+        // When this is the case the user will be redirected to the index action of this module.
+        // An action to which he may not have any user rights.
+        // Redirect to the user's own profile instead to avoid unnessary words.
+        if(
```
jeroendesloovere:
```php
if (
```

src/Backend/Modules/Users/Actions/Edit.php
```diff
@@ -57,6 +57,26 @@ class Edit extends BackendBaseActionEdit
...
+            $this->id === null &&
+            $this->error === null &&
+            $this->authenticatedUser->getUserId()
+        )
+        {
```
jeroendesloovere:
```php
) {
```

src/Backend/Modules/Users/Actions/Edit.php
```diff
@@ -57,6 +57,26 @@ class Edit extends BackendBaseActionEdit
...
+        $this->error = $this->getParameter('error', 'string');
+        $this->loadAuthenticatedUser();
+
+        // If id and error parameters are not set we'll assume the user logged in
+        // and has been redirected to this action by the authentication index action.
+        // When this is the case the user will be redirected to the index action of this module.
+        // An action to which he may not have any user rights.
+        // Redirect to the user's own profile instead to avoid unnessary words.
+        if(
+            $this->id === null &&
```
jeroendesloovere:
```
$this->id === null
```

src/Backend/Modules/Users/Actions/Edit.php
```diff
@@ -57,6 +57,26 @@ class Edit extends BackendBaseActionEdit
...
+            $this->error === null &&
+            $this->authenticatedUser->getUserId()
```
jeroendesloovere:
```
$this->id === null
&& $this->error === null
&& $this->authenticatedUser->getUserId()
```

src/Backend/Modules/Users/Engine/Model.php
```diff
@@ -253,6 +253,28 @@ public static function getGroups()
     }
 
     /**
+     * Get all relevant actions for a certain module.
+     *
+     * @param string $module The module name.
+     * @return array
+     */
+    public static function getModuleGroupsRightsActions($userId)
+    {
```
jeroendesloovere:
```
public static function getModuleGroupsRightsActions($userId) {
```

src/Backend/Modules/Authentication/Tests/Actions/IndexTest.php
```diff
@@ -77,5 +92,77 @@ public function testAuthenticationWithCorrectCredentials()
...
+
+    /**
+     * Login as a users user.
+     * This user only has the rights to access the users edit action.
+     * It should enable the user to edit his own user-account.
+     */
+    public function testUsersUserWithCorrectCredentials()
+    {
+        $client = static::createClient();
+        $client->setMaxRedirects(2);
```
WouterSioen: I believe you need more redirects here, the tests fail with "LogicException: The maximum number (2) of redirections was reached."

In the previous test, this is set to 10.

In the previous test, you also state `$this->followRedirects()` but not here. This line is not needed, since `$this->setMaxRedirects()` will put 'follow redirects' on true, as long as this max number is not reached.

src/Backend/Modules/Users/Actions/Edit.php
WouterSioen: @jeroendesloovere that's not really a rule in the coding styles. I like it more with the '&&' statements in the front, because it's easier to see tha it's a multiline if statement, but it is not required.

src/Backend/Modules/Authentication/Tests/Actions/IndexTest.php
WouterSioen: Looks fixed now :+1:

src/Backend/Modules/Authentication/Tests/Actions/IndexTest.php
tommyvdv: 2 redirects should be enough:
* user logs in and gets redirected to the first allowed action (being Users Edit);
* Users Edit can't find an id parameter and redirects to itself with the current user id.

In fact, 1 redirect should be enough for the testPagesUserWithCorrectCredentials test.

Setting the max redirects as tight as can be could help rule out unexpected behavior.

WouterSioen: There is in fact one more redirect:

* user logs in and gets redirected to dashboard (or url in his querystring)
* user gets redirect to the first allowed action  (being Users Edit)
* users edit can't find an id parameter and redirects to itself with the current user id.

I guess the tests should thus pass when you increase the max redirects from 2 to 3.

WouterSioen: Submitting the form is also a redirect, so you should increment the amount of possible redirects with 1.

If this is ok, and the tests are passing, we'll merge this.

WouterSioen: I was wrong about this one. The problem was an infinite loop caused by the backend reading $_GET parameters, wicht don't exist in our functional testsuite. I could fix this by fetchting the parameters from the request object. I'll merge this with my new commit.

WouterSioen: https://github.com/forkcms/forkcms/commit/e61545bc42153e2a2ced45491d3c96635b5bf957

This fixed it.

PR #1225, opened 2015-06-03
Fixes #1174, use date_format_short from settings to format $form->addDate.

### LINKED ISSUE #1174: Frontend input date field reformats date incorrect
I created a custom form on the frontend with the following code:
```php
$date = new \DateTime('today');
$form->addDate('start_date', $date->getTimestamp());
```

On the frontend in the code the date is written as `02/05/2015` but jQuery changes it to `05/02/2015` which is wrong because today is `02/05/2015`. I can't find the problem, it happens since I updated from 3.8.7 to 3.9.1.

ikoene:
Solves #1174

### DIFF
--- a/src/Frontend/Core/Engine/Form.php
+++ b/src/Frontend/Core/Engine/Form.php
@@ -161,8 +161,8 @@ class Form extends \SpoonForm
             throw new Exception('A date field with type "range" should have 2 valid date-parameters.');
         }
 
-        // @later	get preferred mask & first day
-        $mask = 'd/m/Y';
+        // set mask and firstday
+        $mask = Model::get('fork.settings')->get('Core', 'date_format_short');
         $firstDay = 1;
 
         // build attributes


src/Frontend/Core/Engine/Form.php
```diff
@@ -161,8 +161,8 @@ public function addDate(
             throw new Exception('A date field with type "range" should have 2 valid date-parameters.');
         }
 
-        // @later	get preferred mask & first day
-        $mask = 'd/m/Y';
+        // set mask and firstday
+        $mask = Model::getContainer()->get('fork.settings')->get('Core', 'date_format_short');
```
WouterSioen: You can just use `Model::get('fork.settings')->get('Core', 'date_format_short')`

Tip: if you commit this, you can use `git commit --amend` to add it to this commit. You'll need to force push though :)

WouterSioen:
:+1:

src/Frontend/Core/Engine/Form.php
ikoene: Fixed.

PR #1228, opened 2015-06-05
Zipped modules scanned more reliably for module location.

nosnickid:
Running the legit command:

    zip -r MyModule.zip src/Frontend/Modules/MyModule/ src/Backend/Modules/MyModule/

Does not include the `src/` folder as a "dangling folder" entry, as was previously expected by the module zip uploader.

This new method of extracting a prefix is more reliable under differing zip formats.

It also now functions if a module is uploaded into a single top level directory, eg

    /myModuleName/src/....


WouterSioen:
Thanks for the PR, I'll test this soon :+1:

PR #1229, opened 2015-06-05
[WIP] Module routing

vytsci:
Like @WouterSioen asked I made PR. DO NOT MERGE THIS. Code is incomplete, only Core part is done. This is for review and comments purposes. And if you will approve i will make all modules compatible with this feature. And you will be able to release this code any time you want. Thanks for reviews.


vytenizs:
This is only basic version of what can be achieved. But idea is that every module could have its own routing and that should be cool.

src/Frontend/Core/Engine/Model.php
```diff
@@ -326,6 +329,29 @@ public static function getModuleSettings($module)
     }
 
     /**
+     * Creates router for module based on routing file
+     *
+     * @param $module
+     * @return null|Router
+     */
+    public static function getModuleRouter($module)
```
WouterSioen: I'd put this method in the `Frontend/Core/Engine/Base/Config` class, it's only used there, so it would be more logical to be in there. This would also enable you to make the method not static.

src/Frontend/Core/Engine/Model.php
```diff
@@ -326,6 +329,29 @@ public static function getModuleSettings($module)
...
+
+        $file = FRONTEND_MODULES_PATH . '/' . $module . '/Resources/config/routing.yml';
+
+        $fs = new Filesystem();
+
+        if ($fs->exists($file)) {
+            return new Router(new YamlFileLoader(new FileLocator()), $file);
+        }
+
+        return null;
```
WouterSioen: This statement is not needed. If you don't add a return statement in a method, it will automatically return null.

src/Frontend/Core/Engine/Model.php
WouterSioen: Hmm, nevermind, it's also used in the navigation class.

vytenizs: At first i did like you said, but letter i realised that i need same functionality inside static methods. I do not like duplicate code, so i came to this solution.

src/Frontend/Core/Engine/Model.php
vytenizs: I will make another follow up.

src/Frontend/Core/Engine/Model.php
WouterSioen: It'd still be more awesome if we didn't need to make this a static method, but I'll think of how we can fix this :)

vytenizs: If we could pass Module object through methods instead of just module name, that would solve lots of problems. But this is huge work i think.

vytenizs: Then we could use something like $module->getConfig()->getRouter() and so on. And then we could put this code inside config.

WouterSioen: That would really be awesome. I don't have a lot of time soon, but I'll check if I can find some time to do this. This would likely fix a lot of problems and make a lot of code cleaner.

vytenizs: I will look into this, but i tried to think of something. First of all problem is that getURLForBlock method is used by spoon tampate and static module models. So there is not so many option. Maybe we could use singleton pattern and use getInstance when we need it, because i dont think that creating new module instance every time we need config is a smart solution.

WouterSioen: A lot of the static methods should be put in instantiated objects IMO. The Navigation object could for example be a service in our dependency injection container that gets other data injected.

vytenizs: Or we could load all modules into s2 container. So we could reach them with $this->getContainer()->get('Blog') or something like that.

vytenizs: Lots of Fork CMS statics should be gone with S2

WouterSioen: True, but I don't know if putting all modules in the container is a great idea. I think we should only put the current module in there. Or at least in some object thats in the container.

vytenizs: I mean modules should be accessible through container and if not what about singleton?

WouterSioen:
I talked this over with a colleague and we decided this:

* Backwards compatibility can be broken. If this system works, we don't need the `} elseif ($actionParameter !== null)  {` part anymore. Every module should define their routes this way.
* We'd like to set the _locale variable in the request in the processQuerystring method in Frontend\Core\Engine\Url. This would enable us to use translate routes
* For methods that don't depend on slugs (like blog/archive), we'd use translated routes.

It would look something like this (but it will probably need some finetuning):

````
archive:
    path:      /archive
    requirements:
        _locale:  en

archive:
    path:      /archief
    requirements:
        _locale:  nl
````

Is this something you can do? We really like the idea of having those mini routers per module, so it would be great if you could implement something like this.

**Comment by vytenizs** [comment]
Yes, actionParameter will become pointless with this feature. And yes it will completely break backwards compatibility, ofc if module does not have routing.yml it goes like before. I thought about that. If you are interested in this implementation i will work more on this and provide you with better code. This is just a nightly build made within few hours. I just wanted to show idea for module having its own routing.

**Comment by vytenizs** [comment]
All this routing could be made on PHP side, so we should be able to use action translations. If someone of you will have any great ideas, please write it down here. I will constantly check on this and i will try implement everything you write ASAP.

**Comment by vytenizs** [comment]
What if we could use action translations? for example we have action translation with reference "Archive" and translation "archive" we could turn reference code into underscore notation and fetch all translation into assoc array. For example.:

'__archive' => 'archive',
'__comments' => 'comments'

and so on. We can build custom route loader, load all translation and then check for parameters that matches translation reference code. Routing path should look like this:

path: /{__archive}/{year}/{month}

Custom loader would insert translation value into a path. So path would look like this:

path: /archive/{year}/{month}

And when times comes for comparison everything goes as planned. What do you think?

**Comment by vytenizs** [comment]
Here is bundle:
https://github.com/BeSimple/BeSimpleI18nRoutingBundle

But i dont like idea that end user cant change values through backend. I love action translations, so i wish to fork uses them...

**Comment by vytenizs** [comment]
Made some changes take a look. Now archives works with years and with years and month. Rss works too. {_action} is being replaces by action translation which is based on route name. Route names should be equal to action names, but snake case rather than camelCase. Waiting for comments.

**Comment by WouterSioen** [comment]
I think this is looking pretty good :)

**Comment by vytenizs** [comment]
But i still dont like that simple replace. Maybe someone could suggest more neat solution?

**Comment by WouterSioen** [comment]
Thats the only part thats a little messy.

The only way I see to make this cleaner, is by making a custom Router class `ModuleRouter` that extends Symfony's router, and just runs this code snippet in it's controller to rewrite it's routes to the translated versions.

It won't avoid that piece of code, but it will put it in a more reasonable place.

**Comment by vytenizs** [comment]
i dont see in {_action} part. Its like _controller or smth. But that replace part is a bit messy. This is fast solution and needs to be replaced. I will try to implement something better while extending s2 routing as soon as i will be available :)

**Comment by WouterSioen** [comment]
:+1:

**Comment by vytenizs** [comment]
Ok now i made some few changes that might be a little bit cleaner. Still one fragment is on to do list. I dunno where should be a proper place for an extended s2 component within fork CMS. Please review and comment.

**Comment by vytenizs** [comment]
Module config objects makes 3 duplicate code fragments if we somehow manage to store all those configs or lazy load them i dunno. We could solve lots of problems while making module configs accessible more easily.

File: src/Frontend/Core/Engine/Block/Extra.php [hunk 3]
```diff
@@ -163,23 +163,36 @@ public function execute()
     public function getAction()
     {
         // no action specified?
-        if ($this->action === null) {
+        if (null === $this->action) {
```
WouterSioen: I don't really like reversing the arguments in this conditional (called yoda conditions). It avoids assignments in your conditionals, but it has as tradeof that the readability of your code is less good. I think "$this->action is null" reads better then "null is $this->action" since you check the $this->action variable, not the null.

src/Frontend/Modules/Blog/Resources/config/routing.yml
```diff
@@ -0,0 +1,19 @@
+index:
+  path: /
+  defaults: { _action: 'index' }
+
+archive:
+  path: /{_action}/{year}/{month}
+  defaults: { _action: 'archive', month: null }
```
WouterSioen: I'd also add requirements. The year should at least be numeric.

src/Frontend/Core/Engine/Block/Extra.php
```diff
@@ -163,23 +163,36 @@ public function execute()
...
+
+            // if we have module routing lets check it first and set action parameter
+            if ($this->config->hasRouter()) {
+                try {
+                    $parameters = $this->config->getRouter()->match('/' . implode('/', $this->URL->getParameters()));
+                } catch (\Exception $e) {
+                    $parameters = array();
+                }
+
+                $actionParameter = isset($parameters['_action'])?$parameters['_action']:null;
```
WouterSioen: This doesn't follow psr2 coding styles. You should have spaces around the '?' sign and the ':'.

vytenizs:
Made an follow up and some changes. I dunno if its a good idea to put module config inside container and managed to load router within module config. But i was unable to remove getModuleRouter from Frontend Model. Dunno know what should be changed to achieve this, i thought about loading navigation as service, but while its huge refactor i did not do it. Waiting for comments :)

WouterSioen:
Putting the nav in the service container is would be a lot of work and isn't really needed for this PR. I think this should be another refactoring. This PR should just focus on the custom module routing.

vytenizs:
Ok probobly final version for now. Removed getModuleRouter from FrontendModel. One major issue with this code is that getUrlForBlock is called from within lots of cycles. And i still cant figure how to properly implements this into SpoonTemplate. Any suggestions?

vytenizs:
If everything will be confirmed i will start to implement this into whole blog module and later on other modules.

WouterSioen:
Hmm, sending an array to a modifier in spoontemplate is not possible I think (remember me if I'm wrong @tijsverkoyen, you know spoon better then I do). We could say that you have to assign this kind of url's to the template. Normal block url's like /blog/rss don't need this I guess?

vytenizs:
Could we use smth like sprintf? I think only solution is to generate urls manually. Take module url with action if possible and other parameters will go manually i guess. Because Spoon will not allow generate assoc arrays.

tijsverkoyen:
SpoonTemplatemodifiers doesn't except arrays

WouterSioen:
Are there places where the url can't be build in PHP instead of in our templates?

vytenizs:
Every url can be built within PHP Model. Ill look into this and will try to complete this PR as soon as I can.

WouterSioen:
That would be awesome!

WouterSioen:
@vytenizs any news on this?

vytenizs:
I will work on this this friday. Was unable to spare any time. Sorry.

WouterSioen:
No problem, we're looking forward to it.

vytenizs:
Maybe i will find some time tomorrow.

vytenizs:
Managed to merge with master. I will work on this tomorrow :)

vytenizs:
Nice managed to pass tests too :D

vytenizs:
Ok finally made lots of changes. Manage to refactor old links within blog module. Please take a look :)

WouterSioen:
I've tested this and it works great. I have some small remarks though:

- Blog detail: when the category url does not match the blogpost url, a 404 should be given, to avoid duplicate  content
- Backend: when editing a blogpost, you see the wrong url. It still contains '/detail/'.

Can you still change these two small things?

vytenizs:
Ok ill look into it ASAP.

vytenizs:
Take a look, I dunno if its proper solution within backend, but other options are way too much work and needs lot of remake of current code. If you have any suggestions please share :)

vytenizs:
If Fork CMS will implement this feature, I guess Fork needs more advanced widget management, because lots of pages will be unmanageable. I mean that before we had action name in our url so we were able modify widgets of a page, but now all widget from parent will be inherited and its kinda bad at some cases. Any suggestions?

WouterSioen:
Hmm, thats true. Another possibility could be to show all these module routes in the tree in the backend, thus allowing users to couple widgets to it?

vytenizs:
Should be cool to see action pages within pages tree as some kinda inactive pages (you cant delete, move and only widgets should be applied to them). But this is huge improvement :)

jeroendesloovere:
This PR is awesome.

Would we great if we can reduce the following rules (and more like Symfony2 controllers):
https://github.com/forkcms/forkcms/pull/1229/files#diff-f27c07cbee56b91e3f379a82ce5c8bc7R98

vytenizs:
I did not have enough time to dive deep into a details. Ofc this PR needs some tweaks, but currently I have no available time to develop this :) It needs lots of tweaks, this PR is only a vision of how future modules should be routed.

vytenizs:
And I tried to make this as much backwards compatible as I can. So that is why some features are missed.

jeroendesloovere:
@WouterSioen What do you think about it? Is there a way that we work more towards Symfony controllers for this?
public function execute($year, $month)

jeroendesloovere:
@vytenizs no problem. Your work is awesome and a good start for further enhancements.

vytenizs:
If u develop something bigger, it should go to a completely different release. This routing does not conflict with current one. Main purpose was not to break modules, that has older version of routing. But now 3.10 is on a end of a tunnel :D so this could be integrated also, because either way every module will have to be re-factored to fit 3.10.

carakas:
@vytsci any update on making this fork 4 compatible ?

vytsci:
Can anyone take this task from me? Im in a deep hole when talking about free time at the moment. :) Dont have event a free minute to rest.

WouterSioen:
Since there's no progress, we'll close this for now.

PR #1233, opened 2015-06-10
Move phpunit config to the project root.

dubgeiser:
This makes it a tad easier to run the tests; no '-c' option necessary, a
developer not reading the README can run 'phpunit' and have it behave
without unexpected results.


WouterSioen:
There CONTRIBUTING.md file and the travis.yml file should also be changed.

dubgeiser:
*doh*... stupid me, there I was thinking all was fine and dandy ;-)
Changes pushed.

WouterSioen:
The tests are green again :tada:

MathiasDewelde:
I know it's not a big deal but in Symfony the phpunit config file is also placed inside the app folder.
If we want to move to full-stack Symfony i would stick to placing it in the app folder?

WouterSioen:
I'm pro for moving it to the root folder. Having to specify the config location every time kind of sucks.

WouterSioen:
@MathiasDewelde We're moving towards Symfony full stack, but this doesn't mean we have to fully mimic what they are doing. If we have the feeling that putting this in the root folder is better for developer experience, I think we should do this, even though it's not what Symfony (standard edition) does.

dubgeiser:
@MathiasDewelde It also depends in which Symfony repo you're looking; the plain/core Symfony repo (just the framework) has the phpunit config in the project root, while the app demonstrating Symfony has it in the app/ folder...
see https://github.com/symfony/symfony/ and https://github.com/symfony/symfony-standard/ respectively.

WouterSioen:
I've read some issues on symfony-standard and there reasoning to put it in the app folder, is because they use the bootstrap.php.cache file thats strored in the app directory. We don't use this, but use just the vendor/autoload.php file. This means that their reason to put it in there does not apply to our code.

MathiasDewelde:
Ok i didn't know that they putted the phpunit config file in the app folder because of the bootstrap.php.cache file, thanks for pointing that out!

Anyway the only reason i commented was because i thought it was important to stick as close to the Symfony 'conventions' as possible to help new developers en contributers that are used to using Symfony. But it's perfectly fine for me if the file lives in the root folder. It will actually be easier to use :)

WouterSioen:
I've had to look it up too ;)

tijsverkoyen:
:+1:

PR #1234, opened 2015-06-11
Save log files separately for each environment.

WouterSioen:
This makes it a lot easier to debug issues in our test environment without having to go trough all the logs of our dev environment.


jeroendesloovere:
:+1:

PR #1235, opened 2015-06-11
Bugfix common model

jeroendesloovere:
**Problem**

I installed Fork CMS 3.9.4 and **startProcessingHooks()** was missing in `Common/Core/Model.php` and throwing an error.

**Solution**

Moved this function from Backend/Frontend to `Common/Core/Model.php`


WouterSioen:
@jeroendesloovere can you squash this?

PR #1239, opened 2015-06-12
Fix composer post install/update on windows

### LINKED ISSUE #1238: Composer post-install scripts windows
Hi

Although using the git bash on windows still makes it impossible to execute the post-install (& -update) scripts in the composer.json file and therefore the Fork CMS installation cannot be completed (javascript malfunctioning and partial css styles).
If the `php` and `sh` commands are explicitly defined, everything runs smoothly :)

```
"scripts": {
    "post-install-cmd": "php app/console assetic:dump . && cd tools && sh ./remove_cache",
    "post-update-cmd": "php app/console assetic:dump . && cd tools && sh ./remove_cache"
},
```

Tnx Glenn

glenncoppens:
Fixes #1238

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,8 +29,8 @@
         "issues": "https://github.com/forkcms/forkcms/issues"
     },
     "scripts": {
-        "post-install-cmd": "app/console assetic:dump . && cd tools && ./remove_cache",
-        "post-update-cmd": "app/console assetic:dump . && cd tools && ./remove_cache"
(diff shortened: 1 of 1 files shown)


WouterSioen:
The remove cache script is in fact a bash script and not a shell script, I think you should use the "bash" command instead of "sh"

glenncoppens:
Ah, there is indeed a difference between bash and sh. My bad :)
Either way, composer install works fine now.

WouterSioen:
Ok, looks good IMO :+1:

PR #1240, opened 2015-06-15
Template modifier show bool

carakas:


src/Backend/Core/Engine/TemplateModifiers.php
```diff
@@ -272,4 +272,31 @@ public static function truncate($var = null, $length, $useHellip = true, $closes
...
+     * @param string|bool $status
+     * @param bool        $reverse show the opposite of the status
+     * @return string
+     */
+    public static function showBool($status, $reverse = false)
+    {
+        if (!$reverse) {
+            if ($status === 'Y' || $status === 'y' || $status === 1 || $status === '1' || $status === true) {
+                return '<strong style="color:green">&#10003;</strong>';
+            } elseif ($status === 'N' || $status === 'n' || $status === 0 || $status === '0' || $status === false) {
```
WouterSioen: Why would you use an else, if you already have a return statement in your previous statement?

WouterSioen:
I think a switch statement would be more clear here, since you have a lot of options that lead to the same return values. This will make it cleaner to add more possible values later.

src/Backend/Core/Engine/TemplateModifiers.php
carakas: fixed in 586d6fe

carakas:
a switch structure won't work because a switch does loose comparision

src/Backend/Core/Engine/TemplateModifiers.php
```diff
@@ -272,4 +272,33 @@ public static function truncate($var = null, $length, $useHellip = true, $closes
...
+     */
+    public static function showBool($status, $reverse = false)
+    {
+        $showTrue = '<strong style="color:green">&#10003;</strong>';
+        $showFalse = '<strong style="color:red">&#10008;</strong>';
+
+        if ($reverse) {
+            if ($status === 'Y' || $status === 'y' || $status === 1 || $status === '1' || $status === true) {
+                return $showFalse;
+            } elseif ($status === 'N' || $status === 'n' || $status === 0 || $status === '0' || $status === false) {
```
WouterSioen: you can even drop the 'else' part from the 'elseif', because you return in the previous part.

WouterSioen:
good point about the switch statement, you can leave it this way :)

WouterSioen:
Also: tests would be nice (and really easy to add).

src/Backend/Core/Engine/TemplateModifiers.php
carakas: fixed in 586d6fe

ikoene:
There are an excessive amount of return statements no? Maybe assign the value to a var and return in the end.

WouterSioen:
Why would you let the code run further if you can just return a value?
Early returning is a good way to decreasy complexity by creating less code
paths and thus less possible bugs :)

Op maandag 15 juni 2015 heeft Koen <notifications@github.com> het volgende
geschreven:

ikoene:
Hm, I always thought having a single return was good practice, but your arguments make sense.

WouterSioen:
@ikoene FYI https://programmers.stackexchange.com/questions/118703/where-did-the-notion-of-one-return-only-come-from

ikoene:
Interesting read. Thanks for that.

src/Backend/Core/Engine/TemplateModifiers.php
```diff
@@ -272,4 +272,39 @@ public static function truncate($var = null, $length, $useHellip = true, $closes
...
+    /**
+     * Shows a v or x to indicate the boolean state (Y|N, j|n, true|false)
+     *
+     * @param string|bool $status
+     * @param bool        $reverse show the opposite of the status
+     * @return string
+     */
+    public static function showBool($status, $reverse = false)
+    {
+        $showTrue = '<strong style="color:green">&#10003;</strong>';
```
tijsverkoyen: Why a strong-element? wouldn't it be better to use a `<span>`-element with a classname. So the colors can be decided by the designer/frontender.

src/Backend/Core/Engine/TemplateModifiers.php
```diff
@@ -272,4 +272,39 @@ public static function truncate($var = null, $length, $useHellip = true, $closes
...
+
+        if ($status === 'Y' || $status === 'y' || $status === 1 || $status === '1' || $status === true) {
+            return $showTrue;
+        }
+
+        if ($status === 'N' || $status === 'n' || $status === 0 || $status === '0' || $status === false) {
+            return $showFalse;
+        }
+
+        return $status;
```
tijsverkoyen: I think you can simplify the whole codeblock like this:

```php
$result = false;

if (
	$status === 'Y' || $status === 'y' ||
	$status === 1 || $status === 1
) {
	$result = true;
}

if ($reverse)
{
	return ($result) ? $showFalse : $showTrue;
}

return ($result) ? $showTrue : $showFalse;
```

tijsverkoyen:
Also, squash the commits

src/Backend/Core/Engine/TemplateModifiers.php
WouterSioen: The code may be more simple, but I think that the extra `$result` variable makes it harder to comprehend.

tijsverkoyen: Another disadvantage of my code is that all other values will return the "false"-html.

WouterSioen:
merged in 56fd38a

PR #1244, opened 2015-06-19
Fix blog file names

tommyvdv:
A previous commit (54b434725394c3fa50367c85b921dc6e1891b584) exposed an issue with filenames/revisions and partially fixed in.
Action edit checks the name of the file and compares it to the previous image.
Because it now incorporates the revision id, this check was redundant since the result will always be true.

Comparing an excerpt of the name should, in most cases, return false.

### DIFF
--- a/src/Backend/Modules/Blog/Actions/Add.php
+++ b/src/Backend/Modules/Blog/Actions/Add.php
@@ -162,31 +162,32 @@ class Add extends BackendBaseActionAdd
                 $item['num_comments'] = 0;
                 $item['status'] = $status;
 
+                // insert the item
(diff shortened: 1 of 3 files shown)


WouterSioen:
Looking good to me :+1:

WouterSioen:
@tommyvdv can you squash the two "Add the revision id to the image filename." commits? If you don't know how to do this, there is info about it in our CONTRIBUTING guide: https://github.com/forkcms/forkcms/blob/master/CONTRIBUTING.md#squashing-commits

tommyvdv:
@WouterSioen, done.

tijsverkoyen:
:+1:

PR #1250, opened 2015-07-02
Make the Profile::setId method private. Fixes #1237

### LINKED ISSUE #1237: setId method Profile class should not be public
Hi,

Looking at the code of Profile.php, the method setId should not be public. I suggest making it private, setting the id of the profile should be possible by loading it from the database only.

The line of code is here https://github.com/forkcms/forkcms/blob/master/src/Frontend/Modules/Profiles/Engine/Profile.php#L304

WouterSioen:
It should not be possible to alter the id of an entity from outside this
entity.


tijsverkoyen:
:+1:

PR #1253, opened 2015-07-03
Remove unneeded require statements.

WouterSioen:
We user the autoloader generated by composer, which means we don't need
to add those require statements anymore.


carakas:
:+1:

tijsverkoyen:
:+1:

PR #1262, opened 2015-07-20
Fixes loading in duplicate JS & CSS files.

jeroendesloovere:
**Problem**

JS and CSS files should only be loaded in one time.

**Solution**

We added a check to see if the JS or CSS file is already added or not, so we don't load in duplicate files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Improvements:
 
 Bugfixes:
+* Frontend: no more duplicate JS-files.
 
 
 3.9.4 (2015-07-09)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Improvements:
 
 Bugfixes:
+
 * Frontend: no more duplicate JS-files.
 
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Improvements:
 
 Bugfixes:
+* Frontend: no more duplicate JS-files.
```
WouterSioen: there should be a blank line above a list in markdown

src/Frontend/Core/Engine/Header.php
```diff
@@ -151,18 +151,14 @@ public function addCSS($file, $minify = true, $addTimestamp = null)
...
 
-        $inArray = false;
-        foreach ($this->cssFiles as $row) {
-            if ($row['file'] == $file) {
-                $inArray = true;
-            }
-        }
+        $cssFile = array(
+            'file' => $file,
+            'add_timestamp' => $addTimestamp
```
WouterSioen: multiline arrays should always contain trailing comma's

CHANGELOG.md
jeroendesloovere: @WouterSioen Fixed

WouterSioen:
:+1: will be merged

jeroendesloovere:
@WouterSioen Should also be fixed in Backend I guess...

WouterSioen:
Merged in 92ae4cb970bde032375534da29bca485b9519bce (due to conflicts)

PR #1269, opened 2015-07-24
Use pascal casing for Info.xml

tommyvdv:
Slightly improves consistency of the module files.


WouterSioen:
We won't merge this since it breaks all modules, and these files will be removed when moving to bundles.

PR #1270, opened 2015-07-30
adds a function to checkboxes to get enum values easier

carakas:


WouterSioen:
It would be nice if you would change this in the existing modules too.

carakas:
I'll see if I feel like being nice tonight

WouterSioen:
Two points to be changed:

* rename method: "getActualValue" or something more logical.
* implement this in the existing modules or be nice.

carakas:
The 2 points have been changed

PR #1284, opened 2015-09-04
Present visitor with a 404 page when visiting a hidden page.

dubgeiser:
When a page is set hidden in the backend, and it is visited in the
frontend, Fork spits out a 500 error:

    The HTTP status code is not a redirect ("404" given).

This is coming from the Page class when it throws a RedirectException.
In 'normal' circumstances, this doesn't happen a lot, but it is _not_
impossible at all that it would happen.

Doing a 302 redirect might not be ideal, but it does result in a final
404 and it does not bork ;-)

Fixes #1284

### DIFF
--- a/src/Frontend/Core/Engine/Page.php
+++ b/src/Frontend/Core/Engine/Page.php
@@ -16,7 +16,6 @@ use Symfony\Component\HttpKernel\KernelInterface;
 
 use Common\Cookie as CommonCookie;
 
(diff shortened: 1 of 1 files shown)


WouterSioen:
:+1: Looks good to me!

WouterSioen:
We'll merge #1287 because it fixes some additional issues

PR #1286, opened 2015-09-16
Add start-of-body-scripts functionality. Closes #1277

### LINKED ISSUE #1277: Start of body scripts for Google Tag Manager
Google Tag Manager snippets need to be put after the opening <body> tag. This means we'll need a field in the backend similar to the current header and footer scripts. The snippets also need to be added to all templates after the <body> tag.

Since GTM is becoming more prevalent and because marketeers are a whiny bunch, I'd like this built into Fork.

Planning on PR'ing this after feedback.

ikoene:
This is equal the siteHTMLHeader and siteHTMLFooter functionality. You just add a snippet of code in the backend settings and it will be added right after the <body> tag in the template.

### DIFF
--- a/src/Backend/Core/Layout/Css/imports/core_modules.css
+++ b/src/Backend/Core/Layout/Css/imports/core_modules.css
@@ -458,13 +458,13 @@
 		/* Less important tabs go on the right side */
 
 		#pages.pagesAdd .ui-tabs-nav li,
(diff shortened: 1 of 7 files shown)


WouterSioen:
Looks good to me :+1:

WouterSioen:
Can you add this to the upgrade guide?

PR #1287, opened 2015-09-16
bugfix 500 error when page is hidden, incorrect record

Thijzer:
Lot's of different issues here. First off the cached page keys vs the Model::getPage Query is different. The cached page keys has no idea of hidden content while getPage does. So in this version getPage will return a 404 record when a page is hidden. Also the pageId is new reset to reflect the correction. 
I also move the method getContent more early in the process so newRelic will pass page the correct information and removed a nasty redirect in GetContent.


src/Frontend/Core/Engine/Page.php
```diff
@@ -104,8 +104,14 @@ public function load()
         // set tracking cookie
         Model::getVisitorId();
 
-        // get pageId for requested URL
-        $this->pageId = Navigation::getPageId(implode('/', $this->URL->getPages()));
+        // get page content
+        $this->record = $this->getPageContent(
+            // get pageId for requested URL
```
WouterSioen: Blank lines and comments in method calls seem messy to me. Can't you put this in the comment above the method call?

src/Frontend/Core/Engine/Page.php
```diff
@@ -225,36 +228,25 @@ public function getId()
     /**
      * Get page content
```
WouterSioen: I'd add an `@return array` doc

WouterSioen:
Looks good :+1: Can you handel the feedback?

PR #1288, opened 2015-09-16
AJAX :: LiveSuggest removed redirect, lowered the Complexity

Thijzer:
Removed Redirect in AJAX
Lowered the Complexity a bit of parsePagination


src/Frontend/Modules/Search/Ajax/Livesuggest.php
```diff
@@ -200,14 +198,9 @@ private function getRealData()
             $this->pagination['num_pages'] = 1;
         }
 
-        // redirect if the request page doesn't exist
+        // error if the request page doesn't exist
         if ($this->requestedPage > $this->pagination['num_pages'] || $this->requestedPage < 1) {
-            throw new RedirectException(
```
WouterSioen: shouldn't this change be handled in javascript too?

WouterSioen: poke @Thijzer

Thijzer:
I don't think you wish to inform a customer of a structural error, instead throwing an exception for logging would be a better approach. PS : It's hard to test this code.

PR #1290, opened 2015-09-17
Add leading slash to the getQueryString method. Closes #1260

### LINKED ISSUE #1260: Double language in redirect using queryString
Hi

Building a multi language site (EN, NL, FR), with only NL enabled at the moment, I'm trying to get the querystring and attach it to the login anchor (@jeroendesloovere login link widget) when a profile is logged out.
Eg the result is this (seems to be ok):
```
http://site.dev/nl/aanmelden?queryString=nl/contact
```

When logging in, the redirect happens but the url is malformed, being:
```
http://site.dev/nl/nl/contact
```

I parse the querystring into the template like this:
```
// assign queryString to login link (on each page)
if(!$this->URL->getParameter('queryString')) {
    $this->tpl->assign('queryString', $this->URL->getQueryString());
} else {
    $this->tpl->assign('queryString', $this->URL->getParameter('queryString'));
}
```

Template:
```
{$var|geturlforblock:'Profiles':'Login'}{option:queryString}?queryString={$queryString}{/option:queryString}
```

I can use trim to filter the language part out, but I thought that it might be a misbehaviour?

Tnx

ikoene:


WouterSioen:
This pull request should be done towards the 3.10 branch, since this is not backwards compatible, and can break external modules.

PR #1293, opened 2015-09-18
Locale Analyse and fix for index

Thijzer:
This PR represents 2 things.

1) The locale/index page no longer searches by default. Making this page load much faster. Also will make xml import load faster.

2) I dumped the locale/analyse for something new that in my humble opinion works better.
Also analyse now uses it's own model so it doesn't interfere anymore with the locale model.


src/Backend/Modules/Locale/Engine/Model.php
```diff
@@ -19,6 +19,7 @@
 use Backend\Core\Engine\Model as BackendModel;
 
 use Frontend\Core\Engine\Language as FL;
+use Common\ProjectIntel;
```
WouterSioen: This does not seem to be an existing class ;)

Thijzer:
I refactored it a little. It's a bit faster now with less duplicate code.
Locale from an uninstalled module is excluded.

Continous integration :
We could now write a test that checks the locale during development.

Any other suggestions are welcome

src/Backend/Modules/Locale/Actions/Analyse.php
```diff
@@ -76,7 +77,8 @@ private function loadDataGrids()
         /*
          * Frontend datagrid
          */
-        $this->dgFrontend = new BackendDataGridArray(BackendLocaleModel::getNonExistingFrontendLocale(BL::getWorkingLanguage()));
+        $getNonExistingFrontendLocale = BackendLocaleModel::getNonExistingFrontendLocale(BL::getWorkingLanguage());
```
WouterSioen: This is a strange variable name. I wouldn't include the 'get' prefix.

src/Backend/Modules/Locale/Actions/Analyse.php
```diff
@@ -112,7 +114,8 @@ private function loadDataGrids()
         /*
          * Backend datagrid
          */
-        $this->dgBackend = new BackendDataGridArray(BackendLocaleModel::getNonExistingBackendLocale(BL::getWorkingLanguage()));
+        $getNonExistingBackendLocale = BackendLocaleModel::getNonExistingBackendLocale(BL::getWorkingLanguage());
```
WouterSioen: This is a strange variable name. I wouldn't include the 'get' prefix.

src/Backend/Modules/Locale/Actions/Index.php
```diff
@@ -72,16 +79,26 @@ private function loadDataGrid()
...
-        // get all the translations for the selected languages
-        $translations = BackendLocaleModel::getTranslations(
-            $this->filter['application'],
-            $this->filter['module'],
-            $this->filter['type'],
-            $this->filter['language'],
-            $this->filter['name'],
-            $this->filter['value']
+        // if nothing is submitted
+        // we don't to fetch all the locale
```
WouterSioen: typo

src/Backend/Modules/Locale/Actions/Index.php
```diff
@@ -282,6 +306,10 @@ protected function parse()
             $this->dgActions->getNumResults() == 0
         );
 
+        $this->tpl->assign(
+            'hasSubmissions', $this->hasSubmissions
```
WouterSioen: in psr2, all variables in a multiline method call should be on it's own line.

src/Backend/Modules/Locale/Engine/AnalyseModel.php
```diff
@@ -0,0 +1,262 @@
...
+     * @param string $start front key string
+     * @param string $end back key string
+     * @param string $str the string that needs to be checked
+     * @return mixed return string or false
+     */
+    private static function getInbetweenStrings($start, $end, $str)
+    {
+        $matches = array();
+        preg_match_all("@$start([a-zA-Z0-9_]*)$end@", $str, $matches);
+        return (isset($matches[1]))? current($matches[1]): '';
```
WouterSioen: there should be spaces around the '?' and the ':'.

WouterSioen:
Tested and looking great :+1:

PR #1298, opened 2015-09-22
Avoid catching RedirectExceptions.

### LINKED ISSUE #1297: Mailmotor subscribe redirect does not redirect to include sent parameter
When debug is false, it looks like the mailmotor subscribe action fails to add the parameter `sent` to the URL. Making it seem like the subscription always fails. When in fact the form validation succeeds, CM subscription succeeds but the redirect fails.

```
…
// redirect
$this->redirect(
    FrontendNavigation::getURLForBlock('Mailmotor', 'Subscribe') . '?sent=true#subscribeForm'
);
…
```

edit: Relates to https://github.com/forkcms/forkcms/commit/592cc237.

tommyvdv:
RedirectExceptions are caught. This makes it look like the redirect is
not taking place.

resolves #1297


jeroendesloovere:
👍

PR #1301, opened 2015-09-23
PSR-6 compatible navigation caching

matthiasmullie:
Introduced a PSR-6 compatible cache. Filesystem-based by default
but can easily be swapped for other adapters (or even other
implementations, as long as they are PSR-6)

Navigation is no longer written to a file which is then included,
it'll now just read from *cache* (which may be non-filesystem)
This resulted in being able to drop a lot of code.

Ideally, I'd suggest to also do something similar for Locale caches.
And I noticed there's a Common\Cache\Cache now. It would
probably make sense to also migrate those as well, because:
* more adapters
* PSR compatible
I'm happy to help migrate the above things.


matthiasmullie:
Note: I have not yet tested these changes! I just put them up here to see if it's something you'd be interested in before I invest more time in it ;)

tijsverkoyen:
:+1:

WouterSioen:
:+1:  This is great!

But the implementation is not working yet `PHP Fatal error:  Class 'MatthiasMullie\Scrapbook\Psr6\Pool' not found in /home/travis/build/forkcms/forkcms/app/cache/test/appTestDebugProjectContainer.php on line 383`

src/Backend/Modules/Pages/Engine/CacheBuilder.php
```diff
@@ -10,10 +10,11 @@
  */
 
 use Backend\Core\Engine\Model as BackendModel;
+use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\Filesystem\Filesystem;
 
 /**
- * In this file, the pages cache is build
+ * In this file, the pages cache is built
```
WouterSioen: :+1:

matthiasmullie:
Oh, I just didn't commit composer.lock, so the dependency didn't get installed with `composer install`.
Other than that, I didn't really test this anyway. I will later - I just put it up to gauge the interest, before putting more effort into it.

WouterSioen:
It seems to work now too. All of our automatic tests are at least passing.

matthiasmullie:
I ended up not converting Locale cache: those JSON files are used by the frontend so we still have to store them like that anyway.

Biggest notable change: ModuleSettings are no longer immediately changed in cache. The cache is just deleted when the setting changes & it'll be fetched from SQL & then stored to cache again. This is mostly for concurrency control: to make sure 2 concurrent saves don't end up overwriting one another (and the benefit of doing immediate update is small anyway)
ModuleSettings is now backed by a real cache, though. It used to only be InMemoryCache, but now it'll be cached over multiple requests - until settings change & the cache is purged again.

matthiasmullie:
PS: I've done some testing, but it's been awhile since I've used Fork CMS myself, so please do some more testing before merging :)

src/Backend/Modules/Analytics/GoogleClient/Connector.php
```diff
@@ -176,21 +176,22 @@ private function getData($startDate, $endDate)
...
-                json_encode(
-                    array(
-                        'metrics' => $this->getMetrics($startDate, $endDate),
-                        'visitGraphData' => $this->collectVisitGraphData($startDate, $endDate),
-                        'pageViews' => $this->collectMostVisitedPagesData($startDate, $endDate),
-                        'sourceGraphData' => $this->collectSourceGraphData($startDate, $endDate),
-                    )
-                )
-            );
+        $item = $this->cache->getItem('analytics-'.$dateRange);
```
WouterSioen: We put spaces around our concat operator. (I know that PSR guidelines don't say anything about this, but it would be more conform with the Fork code base)

WouterSioen:
This looks good at first sight

src/Backend/Modules/Analytics/GoogleClient/Connector.php
matthiasmullie: FYI: php-cs-fixer removes spaces around concat, so I tend to follow that myself. But yes, most important is to keep it consistent throughout the project - I'll update it!

WouterSioen: It only does it if you use the symfony standards. Using --level=psr2, the
spaces don't get removed.

Op donderdag 24 september 2015 heeft Matthias Mullie <
notifications@github.com> het volgende geschreven:

WouterSioen:
I've tested this and it works really well. Thanks for the effort!

Issue #1300, opened 2015-09-23
Bug: Array to string conversion while deleting extras


vytenizs:
Hello, this error have been thrown "ContextErrorException: Notice: Array to string conversion in /media/storage/workstation/servers/http/goneplan/src/Backend/Core/Engine/Model.php line 206". This is due to usage of array_intersect with multidimensional arrays. You get extraData and it has multidimensional data in it.

vytenizs:
I dunno if i squashed correctly. Please let me know.

PR #1302, opened 2015-09-23
#1300 (Bug: Array to string conversion while deleting extras)

vytenizs:


src/Backend/Core/Engine/Model.php
```diff
@@ -198,18 +198,24 @@ public static function deleteExtra($module = null, $type = null, array $data = n
...
+            $delete = true;
+
             // match by parameters
             if ($data !== null && $extra['data'] !== null) {
                 $extraData = (array) unserialize($extra['data']);
 
-                // skip extra if parameters do not match
-                if (count(array_intersect($data, $extraData)) !== count($data)) {
-                    continue;
+                // delete extra if parameters do not match
```
WouterSioen: I'd change the comment here to "**do not** delete extra if the parameters do not match"

src/Backend/Core/Engine/Model.php
```diff
@@ -198,18 +198,24 @@ public static function deleteExtra($module = null, $type = null, array $data = n
 
         // loop found extras
         foreach ($extras as $extra) {
+            $delete = true;
```
WouterSioen: I'd rename this variable to "deleteExtra", this is more explicit

WouterSioen:
can you still squash the commits? (this is explained in the CONTRIBUTING.md guide)

src/Backend/Core/Engine/Model.php
```diff
@@ -198,18 +198,24 @@ public static function deleteExtra($module = null, $type = null, array $data = n
...
+
             // match by parameters
             if ($data !== null && $extra['data'] !== null) {
                 $extraData = (array) unserialize($extra['data']);
 
-                // skip extra if parameters do not match
-                if (count(array_intersect($data, $extraData)) !== count($data)) {
-                    continue;
+                // do not delete extra if parameters do not match
+                foreach ($data as $dataKey => $dataValue) {
```
WouterSioen: can't this be done more efficiently using http://php.net/manual/en/function.array-intersect-assoc.php?

WouterSioen: This is possible using http://php.net/manual/en/function.array-diff-assoc.php. This is probably more efficient

vytenizs: Have you tried this with multidimensional arrays?

vytenizs: Ive read that all array function acts tricky with multidimensional arrays

PR #1305, opened 2015-10-04
Remove jQuery UI for search

### LINKED ISSUE #1303: Remove jQuery UI in frontend
Jacob posted [a topic on the fork forum](http://www.fork-cms.com/community/forum/detail/why-include-jquery-ui-css-by-default) why there's jQuery UI css & js loaded into every Fork CMS page. 

As far as I know, it's only used for the search widget autocomplete list and the occasional datepicker in the frontend (not that common in normal websites, and if needed, people tend to use better libraries like [pickadate.js](http://amsul.ca/pickadate.js/)). 

Is it a good thing to remove jquery ui from the frontend, and just load it when the search module is used? That saves two requests (`jquery-ui.css`, `jquery-ui.js`) on pages that don't use search/datepickers... Or remove jquery ui completely and use [typeahead.js](https://twitter.github.io/typeahead.js/) for the search autocomplete?

jessedobbelaere:
* Remove jQuery UI in frontend. 
* Move the search functionality to the search module instead of the `frontend.js` file. 
* When the search module/widget is used on a page, [typeahead.js & bloodhound](https://twitter.github.io/typeahead.js/) is loaded and used to autocomplete the search results. 

*Warning: jQuery UI is used in the frontend for search autocomplete and datepicker fallback (from html5 datepickers). So there should be a solution for datepickers. Or just remove the jquery ui datepicker fallback from the `frontend.js` and leave only the html5 datepicker? Webdevelopers using forkcms should use [pickadate.js](http://amsul.ca/pickadate.js/) or similar when creating datepickers.*

### Preview with typeahead.js: 
No real changes noticeable. Wrote some css that's similar to the previous forkcms search autocomplete. 

schermafbeelding 2015-10-04 om 01 48 33

Resolves #1303


WouterSioen:
Can't you use a minified version of typeahead.bundle.js?

I'd also add a `if ($.isFunction($.fn.datepicker))` around the datepicker code in the frontend.
This way, frontenders can still add jquery ui (or another library if they want) on pages containing a datepicker.

jessedobbelaere:
@WouterSioen I added the minified version and datepicker check :-)

WouterSioen:
great :+1:

WouterSioen:
I've tested this and it works great :+1:

PR #1306, opened 2015-10-05
Make our container available statically

WouterSioen:
If we want to use our static model classes in our code, we need the container to be available.

### DIFF
--- a/app/console
+++ b/app/console
@@ -14,6 +14,7 @@ require_once __DIR__ . '/BaseModel.php';
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
+use Backend\Init as BackendInit;
 
 $input = new ArgvInput();
 $env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'dev');
@@ -33,7 +34,12 @@ $kernel = new AppKernel($env, $debug);
 
 // make our container available statically. This is needed to make our static models work
 $kernel->boot();
+$kernel->defineForkConstants();
 BaseModel::setContainer($kernel->getContainer());
+
+define('APPLICATION', 'Console');
+$backend = new BackendInit($kernel);
+$backend->initialize('BackendCronjob');
 
 $application = new Application($kernel);
 $application->run($input);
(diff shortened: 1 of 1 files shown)


jeroendesloovere:
:+1:

jeroendesloovere:
Probably best to use:
```php
require_once __DIR__.'/BaseModel.php';
```

```php
BaseModel::setContainer($kernel->getContainer());
```

jeroendesloovere:
Thats not all, also add
```php
define('APPLICATION', 'Console');
$backend = new BackendInit($kernel);
$backend->initialize('BackendCronjob');
```
Or some other "Backend" application name,
then everything works and defined constants like "FRONTEND_CACHE_PATH" can be found in BackendModel...

jeroendesloovere:
@WouterSioen, my last code was needed to let Symfony2 Commands use BackendModel (which use container). Did you already found a better solution then the one I used?

jeroendesloovere:
@WouterSioen I updated your PR. With my last commit, PATH_WWW and FRONTEND_FILES_PATH constants are getting defined.

app/console
```diff
@@ -29,5 +31,15 @@ if ($debug) {
 }
 
 $kernel = new AppKernel($env, $debug);
+
+// make our container available statically. This is needed to make our static models work
+$kernel->boot();
+$kernel->defineForkConstants();
+BaseModel::setContainer($kernel->getContainer());
+
+define('APPLICATION', 'Console');
```
jeroendesloovere: @WouterSioen We must define "APPLICATION" here. Otherwise we get the following error:
```php
PHP Notice:  Use of undefined constant APPLICATION - assumed 'APPLICATION' in ./src/Backend/Init.php on line 92
```
But I'm not sure if we should use `Backend` as value, or something new like for example: `Console`

WouterSioen: I think defining a new application (like console) makes more sense

jeroendesloovere:
@WouterSioen can you check why tests aren't succesful?

WouterSioen:
the tests are passing. Only the StyleCI (coding styles) check is failing

app/console
```diff
@@ -29,5 +31,15 @@ if ($debug) {
...
+$backend = new BackendInit($kernel);
+$backend->initialize('BackendCronjob');
```
WouterSioen: I'd initialize a new application "Console" here.

WouterSioen: This application could even do these calls I guess:
```
$kernel->boot();
$kernel->defineForkConstants();
BaseModel::setContainer($kernel->getContainer());
```

jeroendesloovere:
@WouterSioen When will this be merged?

PR #1313, opened 2015-10-22
Make the url of the last breadcrumb item available

jacob-v-dam:


WouterSioen:
looking good :+1:

jacob-v-dam:
Thanks!! Unfortunately I was not able to include the Google Structured data for the breadcrumbs without editing all the HTML, it might be an upgrade for the next release.

PR #1314, opened 2015-10-23
Encode the external redirect URL for pages

jonasdekeukelaere:
When a redirect URL has a double quote in it (not the best practice, but it
happens), it gives unexcepting results with the cache (because the cache is
JSON and the URL will be surrounded by double quotes. Adding the encoding for
the external redirect URL fixes wrong redirect URLs in navigation. Only add
encoding for the URL segments, because otherwise the whole URL (including ://)
will be encoded.

For example:
Redirect URL: http://www.google.be/Quote"HelloWorld"
Becomes: http://www.google.be/Quote in navigation

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -24,6 +24,7 @@
             <exclude>src/Frontend/Modules/Search/Tests/Actions</exclude>
             <exclude>src/Backend/Modules/Authentication/Tests/Actions</exclude>
             <exclude>src/Backend/Modules/Blog/Tests/Actions</exclude>
+            <exclude>src/Backend/Modules/Pages/Tests/Actions</exclude>
             <exclude>src/Api/*/Tests/Actions</exclude>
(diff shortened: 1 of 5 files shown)


WouterSioen:
@jonasdekeukelaere can you provide unit tests for this method? This should be really easy to do since it receives a string and returns one too.

jonasdekeukelaere:
@WouterSioen See 73530cd

WouterSioen:
Could you just squash the last two commits? This will keep our history cleaner

:+1:

jonasdekeukelaere:
Done

WouterSioen:
Hmm, what does this do with a url containing accents? (like `http://cédé.be/Quote"HelloWorld"`)? Can you add a test case for such a case too, because I guess only the part after the domain should be encoded. I think this is not the case right now.

jonasdekeukelaere:
Fixed in 971b8fc

PR #1315, opened 2015-10-23
Use Flysystem as cache backend

matthiasmullie:
Scrapbook now supports Flysystem (a filesystem abstraction) as a
cache adapter. The original Filesystem backend is still around,
but has been deprecated in favor of Flysystem, which is a popular
and proven filesystem library. It likely handles edgecases better
than Scrapbook's original Filesystem adapter did, and allows for
more versatility in filesystem backends (e.g. FTP, AWS, ...)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -64,10 +64,18 @@ services:
             - @fork.settings
             - @service_container
 
-    cache.adapter:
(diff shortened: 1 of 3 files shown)


WouterSioen:
:+1:

PR #1317, opened 2015-11-04
Only load the installer route when Fork is not installed.

WouterSioen:
This will make sure fork sites in production can't be reinstalled.


jeroendesloovere:
:+1:

PR #1328, opened 2015-11-14
Update the install_locale script + Symfony command

jessedobbelaere:
I was using the install_locale command but it's only possible to add the full path to a locale.xml file.  **Ideally, we could just use the modulename to import locale** and say `php install_locale.php -o -m blog` to install the blog module locale.xml. I quickly added this option...

#### Some remarks:
* Why are Exceptions used in this script? This doesn't make sense to me as they are not printed in the CLI. I first ran the script with errors but no output was given so I had no clue.
* **Is there already a possibility in Fork CMS to add Symfony commands?** I would like to refactor this script to an `app/console locale:install` command or something like that. Any input on how to achieve this in forkcms? What folder would Fork cms use for commands? `/src/Backend/Command` or `/src/ForkCMS/Command`? There's not a real backend bundle yet

### DIFF
--- a/app/console
+++ b/app/console
@@ -30,4 +30,5 @@ if ($debug) {
 
 $kernel = new AppKernel($env, $debug);
 $application = new Application($kernel);
+$application->add(new \Common\Command\ImportLocaleCommand());
 $application->run($input);
--- a/src/Common/Command/ImportLocaleCommand.php
+++ b/src/Common/Command/ImportLocaleCommand.php
@@ -1,0 +1,138 @@
+<?php
+
+namespace Common\Command;
+
+use AppKernel;
+use Exception;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Backend\Modules\Locale\Engine\Model as BackendLocaleModel;
+use Backend\Init as BackendInit;
+
+/**
+ * This is a simple command to install a locale file
+ * @package Common\Command
+ *
+ * @author Jesse Dobbelaere <jesse@dobbelaere-ae.be>
+ */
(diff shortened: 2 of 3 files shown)


WouterSioen:
You can perfectly do this using a Symfony Command. See my slides about this: http://woutersioen.github.io/Symfony-components-in-fork/#28

jessedobbelaere:
Awesome! So I'll remove this script and write a Symfony command in `/src/Common/Command/InstallLocaleCommand.php` when I find some time :-)

Would be nice if more scripts from the tools folder could migrate to Commands so we can delete the tools folder after a while? Or isn't that the plan?

WouterSioen:
I wouldn't remove the old script, for the sake of backwards compatibility (sumocoders uses it in it's deployment scripts (and if I remember correctly, Wijs does this too).

But a new version using the symfony console giving nice output would be really awesome

jessedobbelaere:
Yeah. And will the Common folder become a bundle? I saw it's not registered as one. Which is why each command should be added manually to app/console instead of automatically registering all the commands in the Command folder

WouterSioen:
Yes, but making Common a Bundle does not really make sense, since it's not really "Symfony" code. It's more of a collection of helper and base classes. I think it's cleaner to add the commands to app/console than to make the Common folder a Bundle

jessedobbelaere:
@WouterSioen I added the Symfony command which has better output. Syntax is:

`app/console locale:install [-o|--overwrite[="..."]] [-f|--file[="..."]] [-m|--module[="..."]]`

I would like to add a test, but not sure how I should mock the bootstrapping of Fork in this script and if I should use the database or just check the logic and if the tests exists. Any tips?

WouterSioen:
That's a really hard script to write tests for. I think it's even nearly impossible with the current structure.

jessedobbelaere:
Fixed :+1:

WouterSioen:
Tested and works great :+1:

PR #1337, opened 2015-12-21
Make sure redirect exceptions are always thrown.

### LINKED ISSUE #1336: Failing form redirects in production environment
Hi

Yesterday I deployed my application to the production server (php 5.6). When I ran fork (3.9.5) in debug mode everything worked (profiles registrations, login etc.). But when running it in non-debug mode, everything was fine, except redirects did not work when registering, logging in, ...

So for example: a user registers itself, fills out the form, clicks submit. The form is being validated, activation mail is sent, but then the redirect does not happen and apparently shows the filled out form again with the error message that is defined in the catch scope (Register.php). 

The same happened to other forms also (with try-catch block). I guess it's related to Tommy's previous issue #1297? So every redirect in a try-catch block should be modified to something like (it's not tested yet):

```
try {
    ...

    // redirect
    $this->redirect(SITE_URL . '/' . $this->URL->getQueryString() . '?sent=true');
} catch(RedirectException $e) {
    throw $e;
} catch (\Exception $e) {
    // when debugging we need to see the exceptions
    if ($this->getContainer()->getParameter('kernel.debug')) {
        throw $e;
    }
} finally {
    // show error
    $this->tpl->assign('registerHasFormError', true);
}

...
```

I'll try to add a PR as soon as possible if this is the way to go?

Grtz Glenn

WouterSioen:
Fixes #1336

### DIFF
--- a/src/Frontend/Modules/Profiles/Actions/Register.php
+++ b/src/Frontend/Modules/Profiles/Actions/Register.php
@@ -16,6 +16,7 @@ use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Core\Engine\Navigation as FrontendNavigation;
 use Frontend\Modules\Profiles\Engine\Authentication as FrontendProfilesAuthentication;
 use Frontend\Modules\Profiles\Engine\Model as FrontendProfilesModel;
+use Common\Exception\RedirectException as RedirectException;
 
 /**
  * Register a profile.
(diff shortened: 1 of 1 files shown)


glenncoppens:
:+1:

glenncoppens:
Ah Wouter, Is there any reason why you're not using multiple catch blocks for each type of exception instead of using the `instanceof` keyword?

WouterSioen:
I have no idea if throwing an exception will result in getting it catched in the next catch block. Should be tested :)

PR #1348, opened 2015-12-30
Blog tests

WouterSioen:
Some tests for the backend of the Blog module.

These are the first tests for logged in uses, and will make it easier to write functional tests for other modules in the backend.


jonasdekeukelaere:
:+1:

PR #1349, opened 2015-12-31
Make sure thumbnails never exceed the width of their parent element

mathiashelin:


WouterSioen:
Is this also fixed in the 3.10 branch?

PR #1351, opened 2016-01-04
[WIP] Feature backend twig

Thijzer:
WIP, You be warned!
I will use this thread for discussion and progress

The twig template wrapper is already working, All I had to do is make a copy of the frontend controller and remove a bunch of frontend stuff. 
It's a good time to figure out some best practices together, probably discus a couple of code standards for the new twig handles, functions and removing the ugly raw pipes.

Also the Spoon2Twig converter needs a rewrite, it's hardly working for the files it generates.

tinygrab screen shot 04-01-16 21 37 58

I'm not going to rush this feature.
So please be patient with me...


WouterSioen:
If both backend and frontend use twig, the SpoonTemplate support may be dropped in my opinion. What do you think @forkcms/core-contributors @forkcms/owners ?

This will need a bump to 4.0 though, if we fully switched to Twig.

tijsverkoyen:
I'm pro dropping SpoonTemplate support

ikoene:
Let's leave the past behind and go for Twig.

Thijzer:
It would mean we could dump a lot of stuff and build up a cleaner working TwigTemplate wrapper. Especially in the Frontend. 
I'm Pro.

WouterSioen:
Let's just do it! If you need help with it, just poke. This is an important feature so I think we're all willing to spend some work on this!

Thijzer:
Thanks @WouterSioen, These are the following mayor/minor steps we need to take.

Removing Spoon code
- removing all SpoonTemplate and dependencies.
- removing all tpl files
- removing the template parameter check

Replace or Refactor old spoon template functions.
since we are moving away from Spoon, it's a good time to reevaluate the spoon template function and how the operate.
Removing the raw pipes on functions where needed.

PS : with a optional 4.0 release on the way it might be nice the celebrate the release with a new frontend theme. I have no clue who is up for the task but it still would be nice.

Thijzer:
Here is a nice example for the new syntax 

I now have `{{ getmainnavigation('nav navbar-nav') }}` working

The backend looks like this
```
    /**
     *  Returns main-navigation-html
     *  syntax: {{ getmainnavigation($class) }}
     *
     * @param string $class Class attribute of ul list
     *
     * @return string
     */
    public static function getMainNavigation($class = null)
    {
        return BackendModel::getContainer()->get('navigation')->getNavigation(1, 1, $class);
    }
```
something like this -> `{{ getmainnavigation(param1, param2) }}` 
@WouterSioen or anyone agrees that this is the correct syntax for twig functions.

WouterSioen:
@Thijzer this looks good. I don't get why you should be able to send a class to this method though. The backend is not themable, so this doesn't really make sense to me.

Thijzer:
@WouterSioen I agree, that was how I found it.
I will remove the param from getMainNavigation

WouterSioen:
I've added some commits to the 3.10 branch to make sure the navigation uses Twig and to remove the SpoonTemplate classes from the core.

I think the TemplateCustom is not used yet though. This functionality should be implemented with Twig too.

WouterSioen:
It seems like all widgets and blocks were still parsed using spoonTemplate. This is now fixed in the 3.10 branch. I also fixed the testsuite.

WouterSioen:
I also updated everything from the master branch in the 3.10 branch :)

Thijzer:
I noticed :)  
I just did some large refactorings, removed code duplication working with common Base classes.
Everything in the Frontend is still working so I'm happy for now :).
My naming convention could be better though.

PS : I'll push the backend twig files for last.

WouterSioen:
Looks like you have a Fatal Error in your code:

`.PHP Fatal error:  Class 'Common\Core\Twig\Extensions\Model' not found in /home/travis/build/forkcms/forkcms/src/Common/Core/Twig/Extensions/BaseTwigModifiers.php on line 118`

Thijzer:
Thanks.

WouterSioen:
@Thijzer you did push all your changes? i'll probably have some time tonight to work on this.

WouterSioen:
didn't have time yesterday, it'll be a work for the weekend :)

Thijzer:
Looks OK to me, the issue is isolated. The only problem I see is when custom modules have similar form separations it will simply not work and they will have to refactor to include the form fields in order to make their action work again or make a second separate form to handle the included template.

Many Thanks Wouter.

Next up would be to list up all template functions and modifiers in a google doc and check what could be replaced or deprecated.

Secondly fix any last issues in the templates and the errors the CI is reporting.

Lastly fix the converter and remove any spoon template leftovers in ForkCMS.

tijsverkoyen:
nice work :+1:

WouterSioen:
the groups_rights_actions should also still be added to the template.

WouterSioen:
Closed in favor of the follow up PR #1423

PR #1354, opened 2016-01-06
Readd pages without widgets/blocks to the pages tree.

### LINKED ISSUE #1352: Cannot add new page without a widget on it
When adding a new page without any widgets on it. Fork reports the page as being saved. But the new page is not shown in the navigation.

Here is how to generate the problem:
* Install new forkcms (3.9.6) through "composer create-project forkcms/forkcms ."
* In the backend goto pages and click [add page]
* Remove the widget "search" from the top position
* Click [save]

The backend reports: page ".........." was added. But it does not show up in navigation on the left.

WouterSioen:
Fixes #1352


vytenizs:
Yeah but it will show empty pages within front navigation.

jonasdekeukelaere:
That's quite normal I think, if you don't want a page to show up in the frontend, you need to hide it in the settings of that page in the backend.

PR #1355, opened 2016-01-07
update the instruction for creating an oAuth certificate for the Anal…

carakas:
…ytics module so they follow the new website


WouterSioen:
Thanks!

PR #1359, opened 2016-01-14
Make sure our bad request header is formatted correctly.

WouterSioen:
This changes in a 500 because of the malformed header.


jessedobbelaere:
Hmm why do we set the header manually here, can we do a `throw new HttpException(400);` :) ? Also if you change this header, we should probably also add HTTP1/1 to the 503 header on line 160?

WouterSioen:
You're right, this was a little bit too much of a quick fix. i'll look into it.

WouterSioen:
@jessedobbelaere I've changed it to use a BadRequestHttpException and a ServiceUnavailableHttpException

jessedobbelaere:
awesome :) !

PR #1362, opened 2016-01-14
Set PHP version to 5.5.0 in installer errors

jonasdekeukelaere:


src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/step1.html.twig
```diff
@@ -21,7 +21,7 @@
         {% endif %}
         <div id="requirementsInformation">
           <h3><span class="{{ checker.errors.phpVersion }}">{{ checker.errors.phpVersion }}</span> PHP version</h3>
-          <p>PHP version must be at least 5.3.3, Before using Fork CMS, upgrade your PHP installation, preferably to the latest version. PHP version must not be 5.3.16 as Symfony won't work properly with it.</p>
+          <p>PHP version must be at least 5.5.0, Before using Fork CMS, upgrade your PHP installation, preferably to the latest version. PHP version must not be 5.3.16 as Symfony won't work properly with it.</p>
```
WouterSioen: The 5.3.16 sentence could be removed too, since it's not in the 5.5+ range

WouterSioen:
thanks!

src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/step1.html.twig
jonasdekeukelaere: Fixed in 5b0a862

PR #1367, opened 2016-01-15
Make sure our sitelinks searchbox matches our search field.

WouterSioen:
@guytarded this will probably fix your issue. (But sadly, it's impossible to test)


guytarded:
We'll see :)

PR #1370, opened 2016-01-20
Meta with services

WouterSioen:
This makes sure we can make a "repository" class which isn't static. If
the meta class finds this service in the container, it will use it
instead of static method calls.

### DIFF
--- a/src/Backend/Core/Engine/Meta.php
+++ b/src/Backend/Core/Engine/Meta.php
@@ -435,6 +435,8 @@ class Meta
      */
     public function save($update = false)
     {
+        $this->validate();
+
         $update = (bool) $update;
 
         // get meta keywords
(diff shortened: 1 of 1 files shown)


WouterSioen:
Example usage:

        $this->meta->setURLCallback('job_repository', 'getUrl');

Where `job_repository` is a service which contains a `getUrl` method.

WouterSioen:
@jessedobbelaere this could probably be useful for you, since you use repository services (and probably already encountered this issue).

Can you check this if you have time?

WouterSioen:
Note that I'm already using it in a project, so it probably works well ;)

src/Backend/Core/Engine/Meta.php
```diff
@@ -471,39 +478,16 @@ public function save($update = false)
...
-        if ($this->frm->getField('seo_index')->getValue() != 'none') {
-            $meta['data']['seo_index'] = $this->frm->getField('seo_index')->getValue();
-        }
-        if ($this->frm->getField('seo_follow')->getValue() != 'none') {
-            $meta['data']['seo_follow'] = $this->frm->getField('seo_follow')->getValue();
-        }
-        if (isset($meta['data'])) {
-            $meta['data'] = serialize($meta['data']);
-        }
-
```
jeroendesloovere: I see that you aren't setting the $meta parameter anymore.
Where has this gone to?

WouterSioen: this was done twice

PR #1374, opened 2016-01-25
Delete useless meta records 

bartdc:
This code will delete meta records of removed blog posts when reaching
20 revisions.


WouterSioen:
There's probably something wrong in there, because the tests say `Failed asserting that 500 matches expected 200.` (when editing a blogpost)

WouterSioen:
I guess it fails when there are no revisions to keep. (so you get an empty `IN()` in mysql)

PR #1376, opened 2016-01-27
Fixes Analytics ConnectorTest

jeroendesloovere:
Which throws error because ```the namespace PHPUnit_Framework_TestCase``` not exists.


jeroendesloovere:
When deploying using Capistrano the deployment fails on this.
This is a working solution.

WouterSioen:
using a use statement is just as good as adding a backslash. I don't get how this could fix any issues. It's exactly the same code, just with another syntax.

jeroendesloovere:
When I do this fix, I can ```cap production deploy``` again.
Otherwise it throws fatal error and will rollback.
```php
PHP Fatal error:  Class 'PHPUnit_Framework_TestCase' not found in /apps/production/releases/xxx/src/Backend/Modules/Analytics/Tests/GoogleClient/ConnectorTest.php on line 16
```

jeroendesloovere:
Don't know why, because in the Common folder there are also tests where the ```use``` statement has been used.

jeroendesloovere:
Conclusion:
in modules we should use ```\PHPUnit_Framework_TestCase``` to make everything work in production.

WouterSioen:
`extends \PHPUnit_Framework_TestCase` is *exactly* the same as `use PHPUnit_Framework_TestCase` and `extends PHPUnit_Framework_TestCase`

I don't see how this could make any difference.

WouterSioen:
@jeroendesloovere can you test this in this code base and PR it in there if your problem still exists? https://github.com/forkcms/forkcms/pull/1447

PR #1384, opened 2016-02-01
Fixes for Frontend Analytics tracking

### LINKED ISSUE #1383: Google Analytics: trackOutboundLinks not working for "ga"
**Problem**
In ```src/Frontend/Core/Js/frontend.js``` we check for ```typeof ga === 'object'``` on [this rule](https://github.com/forkcms/forkcms/blob/master/src/Frontend/Core/Js/frontend.js#L510).
But that is false. ```ga``` is a ```function```.
=> So all the outbound trackings are not working.

**Solution**
Check for ```typeof ga === 'function'```

jeroendesloovere:
**Fixes**
* ```ga``` is a function (not an object), fixes #1383 
* Tracking ```anchors``` is now working
* Tracking ```mailto``` is now working

**What I've added**
* An ':internal' check to track ```mailto``` and ```#```-links. So we can push these internal click events to Google Analytics (because :external does not do this).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Bugfixes:
 
 * Analytics: frontend "ga" is a function, not an object.
 
+Improvements:
+* Analytics: internal "anchors" and "mailto"-links are now being tracked.
+
 
 3.9.6 (2015-12-22)
 --
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+3.9.7 (xxxx-xx-xx)
+--
+Bugfixes:
+
+* Analytics: frontend "ga" is a function, not an object.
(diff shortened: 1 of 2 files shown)


WouterSioen:
@jeroendesloovere Do "target" attributes still work on links? The hasTarget check was used to handle that.

jeroendesloovere:
@WouterSioen
I thought that "hasTarget" was introduced so the GA-call had some time (100ms) to be successfully executed before the page was gone.

But the code screws up my "a:internal" link clicking because sometimes you have $('a#id').on('click', function(){}); and ```e.preventDefault()``` screws that.

WouterSioen:
@jeroendesloovere I'll close this since it's rather outdated and some other fixes are included in #1534

PR #1386, opened 2016-02-04
Common Language class

carakas:
This class will make it possible to have 1 function to get the correct language class.

This is useful for when you want to use the same code in the Back- and Frontend.
For instance in a trait.

### DIFF
--- a/src/Common/Language.php
+++ b/src/Common/Language.php
@@ -2,6 +2,8 @@
 
 namespace Common;
 
+use InvalidArgumentException;
+
 /**
  * This class will make it possible to have 1 function to get the correct language class.
  * This is useful for when you want to use the same code in the Back- and Frontend.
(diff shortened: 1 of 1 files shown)


src/Common/Language.php
```diff
@@ -0,0 +1,98 @@
...
+namespace Common;
+
+use InvalidArgumentException;
+
+/**
+ * This class will make it possible to have 1 function to get the correct language class.
+ * This is useful for when you want to use the same code in the Back- and Frontend.
+ * For instance in a trait.
+ *
+ * @author Jelmer Prins <jelmer@$sumocoders.be>
```
jeroendesloovere: Error found in email @carakas

carakas: thanks, fixed in https://github.com/forkcms/forkcms/pull/1476/files

WouterSioen: so critical, thanks for the fix!

PR #1387, opened 2016-02-05
Initialize the container and fork constants when creating a kernel

carakas:
Initialize the container and fork constants when creating a kernel instead of initializing it when handling the request.


WouterSioen:
I think the boot statement can be removed in some of the tools when you do this change.

WouterSioen:
poke @carakas have you checked the tools?

carakas:
@WouterSioen oops, nope I forgot, I'll look at this later today

PR #1388, opened 2016-02-06
Prevent dataloss datefield

carakas:
When there are errors in a form the datefield is reset. This fixes this and will restore the correct date.


WouterSioen:
@carakas can you fix merge conflicts and test this in fork 4.0.x?

PR #1393, opened 2016-02-10
Removes a line that doesn't do anything but potientially throw an error

StijnVrolijk:
This line has some weird logic in it and doesn't actually do anything.
The continue potentially throws a FatalErrorException: Error: Cannot break/continue 1 level


WouterSioen:
Hmm, looks like you did a lot more then this one line change ;)

WouterSioen:
@StijnVrolijk poke (PR wanted with only this one line change)

PR #1394, opened 2016-02-11
Location: Added zoom-level 1 and 2

jeroendesloovere:
**Problem**
I have client that has locations all over the world but the zoom-level only starts from 3.

**Solution**
You can now select 1 and 2 from the zoom-level.


jeroendesloovere:
Closing since there is a new PR https://github.com/forkcms/forkcms/pull/1518 for this

PR #1398, opened 2016-02-13
Upgrade CKEditor to version 4.5.10

### LINKED ISSUE #962: Switching to CKEditor 4.5.10
Is there any implication if Fork moves to CKEditor 4.4.5?

GhaziTriki:
Should fix #962


WouterSioen:
@GhaziTriki have you checked this upgrade in the pages module? Because I thought upgrading breaked the pages module?

GhaziTriki:
I didn't test the pages module. Is it fixed in #1429?

WouterSioen:
nope, this doesn't have anything to do with it. The problem with editors in the pages module is that ckeditor and jquery.ui dialogs don't work well together.

This is fixed in the 3.10 branch

carakas:
There are merge conflicts

carakas:
@GhaziTriki can you fix the conflicts?

GhaziTriki:
@carakas the conflict is solved, however I get this error in JavaScript console

`ckeditor.js:1114 Uncaught TypeError: Cannot read property 'label' of undefined`

carakas:
@GhaziTriki is the `samples/old`dir created by you?
I think the sylesheetparser plugin is removed with this pr and that shouldn't be the case I think.

I would also set a minimum height on the editor, now it is only one line and it looks a bit strange

GhaziTriki:
I have fixed the issue with missing `stylesheetparser` plugin. The folder `samples\old` contains examples for installed legacy plugins.

Please find below a screenshot of the fixed version of CKEditor.

image

PR #1399, opened 2016-02-16
Fix revision page viewing

bart-webleads:
Fixes "Error: Call to a member function addMetaData() on null" when viewing an existing revision page.
How to reproduce the problem?
Goto: http://demo.fork-cms.com/en/?page_revision=1

Fixes "Notice: Undefined index: parent_id" when viewing a page revision where the page is located in the root of the pages tree.
How to reproduce the problem?
Goto: http://demo.fork-cms.com/en/?page_revision=9999


WouterSioen:
looks good :+1:

PR #1402, opened 2016-02-17
Improvement: Add count as template modifier so you can check the amount of items in the template

carakas:


src/Backend/Core/Engine/TemplateModifiers.php
```diff
@@ -307,4 +307,16 @@ public static function showBool($status, $reverse = false)
...
+    /**
+     * Returns the count of the count of the array.
+     *
+     * @param array $data
+     *
+     * @return int
+     */
+    public static function count(array $data)
+    {
+        return count($data);
```
WouterSioen: Isn't there a Common TemplateModifiers class?

carakas: No there isn't, but maybe there should be one

WouterSioen: this could be a different pull request :)

carakas: you have a point there, I will look at this tonight

WouterSioen: I would wait for the twig integration ;) modifiers are completely reworked in there ;)

carakas: oke, let me know when you are finished with that :D

PR #1403, opened 2016-02-18
Bugfix: datagrid sequence "extra-params" to have valid json object

jeroendesloovere:
For example:

You have the following in your Backend Action:
```php
// enable sorting
$this->dgPosts->enableSequenceByDragAndDrop();
$this->dgPosts->setAttributes(
    array(
        'class' => 'dataGrid sequenceByDragAndDrop',
        'data-extra-params' => "{'module':'Products', 'type':'" . $this->category['type'] . "', 'category_id':'" . $this->categoryId . "'}",
        'data-module' => 'Categories',
        'data-action' => 'SequenceForItems'
    )
);
```
> You can not use ", because this breaks the html. So you must use `.

In older jQuery versions this wasn't any problem.
But in the newer jQuery version I noticed that the json validation has become stricter.

**So we must convert ` to " in our javascript.** (=> which is exactly what this PR does).

===
SCREENSHOTS
**Before fix**
schermafbeelding 2016-02-18 om 10 20 41

**After fix**
schermafbeelding 2016-02-18 om 10 20 26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 3.9.7 (xxxx-xx-xx)
 --
 Bugfixes:
-* Pages: removed single quotes converter in CacheBuilder. This fixes page titles with single quotes.
 * Backend: Core fix when using extra-params in a datagrid sequence reorder ajax call.
+* Pages: removed single quotes converter in CacheBuilder. This fixes page titles with single quotes.
 
 3.9.6 (2015-12-22)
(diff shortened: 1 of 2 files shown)


WouterSioen:
@jeroendesloovere can you fix the merge conflicts?

jeroendesloovere:
Closed since it has a new PR #1513 without merge conflicts

PR #1404, opened 2016-02-18
Fixes FAQ-category sequence reordering

jeroendesloovere:


src/Backend/Modules/Faq/Ajax/Sequence.php
```diff
@@ -34,15 +34,16 @@ public function execute()
 
         // loop id's and set new sequence
         foreach ($ids as $i => $id) {
-            // build item
-            $item['id'] = (int) $id;
-
-            // change sequence
-            $item['sequence'] = $i + 1;
+            // define category
+            $category = BackendFaqModel::getCategory((int) $id);
```
WouterSioen: no exists anymore?

jeroendesloovere: Nope, if ($category) does the same...?

PR #1409, opened 2016-02-26
Fixes view URL?page_revision=x

jeroendesloovere:
### Problem

When viewing a ?page_revision=x you get an error because $this->header is not defined yet, so you can't execute ```->addMetaData```.
schermafbeelding 2016-02-26 om 09 31 31

### Steps to reproduce

* Open /private
* Click on home and then click on "save as draft"
* Now click on the "preview page"-button
=> Error is shown when in dev modus

### Two fixes
* We need to define header first
* We need to fetch parent_id also, otherwise another error is thrown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 3.9.7 (xxxx-xx-xx)
 --
 Bugfixes:
+* Pages: viewing ?page_revision=x was not possible.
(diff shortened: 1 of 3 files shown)


jeroendesloovere:
Duplicate for https://github.com/forkcms/forkcms/pull/1399

PR #1418, opened 2016-02-29
Correct the different starting point for months

carakas:
Javastript starts at 0 php at 1


WouterSioen:
Hmm, I remembered this wasn't true for all browsers? Or am I wrong here?

carakas:
I think you are wrong. I did a quick search, but only found people complaining about it. And some references to libraries that compensate it for you

PR #1420, opened 2016-03-03
Fixes bug in utils.string.stripTags for Windows 7 + IE10

jeroendesloovere:
When trying to add new "editor" block in Pages module then you'll previously got error.

### DIFF
--- a/src/Backend/Core/Js/utils.js
+++ b/src/Backend/Core/Js/utils.js
@@ -166,6 +166,7 @@ utils.form =
  * @author	Tijs Verkoyen <tijs@sumocoders.be>
  * @author	Dieter Vanden Eynde <dieter@netlash.com>
  * @author	Matthias Mullie <forkcms@mullie.eu>
+ * @author	Jeroen Desloovere <info@jeroendesloovere.be>
  */
 utils.string =
 {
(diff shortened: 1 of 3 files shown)


src/Backend/Core/Js/utils.js
```diff
@@ -303,7 +304,13 @@ utils.string =
...
 	stripTags: function(value)
 	{
-		return value.replace(/<[^>]*>/ig, '');
+		if(typeof value === 'undefined') return '';
+		else
+		{
+			value = value.replace(/<[^>]*>/ig, '');
+		}
+
+		return value;
```
WouterSioen: I would restructure this code block this way:

```js
if (typeof value === 'undefined') {
    return '';
}

return value.replace(/<[^>]*>/ig, '');
```

This has exactly the same outcome, but is clearer imo.

jeroendesloovere:
Don't know if it has something to do with this...

**new bug**

### Steps to reproduce

1. Log in to /private + production environment 
2. Modules > Pages > a page > add editor block
3. Add some text
4. Click ok
=> then you don't see any text in the editor block.

schermafbeelding 2016-03-04 om 09 31 48
+ When saving the editor blocks aren't saved

WouterSioen:
:+1:

jeroendesloovere:
@WouterSioen Seems my fix above is improving things to go to the next step (from clicking "add new" to showing the editor). But then another bug pops up when clicking "OK"... Editor visualBlock is empty.

WouterSioen:
If I understood the discussion correctly, this fixes a notice but introducers a critical bug?

jeroendesloovere:
Closed since updating ```matthiasmullie/minify``` to version 1.3.34 fixed the problem.
See #1429 for more information.

PR #1425, opened 2016-03-03
Bugfix - Saving page results in Notice: Array to string conversion

V3nn3tj3:
Serialize data before saving in database

### DIFF
--- a/src/Backend/Core/Engine/Meta.php
+++ b/src/Backend/Core/Engine/Meta.php
@@ -477,6 +477,11 @@ class Meta
             $custom = null;
         }
 
+		//serialize data for save
+        if(!empty($this->data['data'])){
+            $this->data['data'] = serialize($this->data['data']);
+        }
+		
         // build meta
         $db = BackendModel::getContainer()->get('database');
 
--- a/src/Backend/Core/Engine/Meta.php
+++ b/src/Backend/Core/Engine/Meta.php
@@ -478,10 +478,10 @@ class Meta
         }
 
 		//serialize data for save
-        if(!empty($this->data['data'])){
+        if (!empty($this->data['data'])) {
             $this->data['data'] = serialize($this->data['data']);
         }
-		
+        
         // build meta
         $db = BackendModel::getContainer()->get('database');
 
--- a/src/Backend/Core/Engine/Meta.php
+++ b/src/Backend/Core/Engine/Meta.php
@@ -477,7 +477,7 @@ class Meta
             $custom = null;
         }
 
-		//serialize data for save
+        //serialize data for save
         if (!empty($this->data['data'])) {
             $this->data['data'] = serialize($this->data['data']);
         }


jeroendesloovere:
Is unserializing needed?

src/Backend/Core/Engine/Meta.php
```diff
@@ -477,6 +477,11 @@ public function save($update = false)
             $custom = null;
         }
 
+		//serialize data for save
+        if(!empty($this->data['data'])){
+            $this->data['data'] = serialize($this->data['data']);
```
carakas: I would only serialize this when you are inserting or updating it in the database. Now you are changing the data of the object itself.

V3nn3tj3:
Maybe it is, but at this time, the data in the meta object is never used after the save is called by anyone.
Could do this though to not edit the structure of the object after businesslogic is done...

`	//serialize data for save
        if (!empty($this->data['data'])) {
            $this->data['data'] = serialize($this->data['data']);
        }
		
        // build meta
        $db = BackendModel::getContainer()->get('database');

        if ($this->id !== null && $update === true) {
            $db->update('meta', $this->data, 'id = ?', array($this->id));
            $this->data['data'] = unserialize($this->data['data']);
            return $this->id;
        } else {
            unset($this->data['id']);
            $this->data['id'] =  (int)$db->insert('meta', $this->data);
            $this->data['data'] = unserialize($this->data['data']);
            return $this->data['id'];
        }`

carakas:
@V3nn3tj3 it is used my reply and line https://github.com/forkcms/forkcms/pull/1425/files#diff-69b87cf219518e05cec9630deeec96beR493

V3nn3tj3:
@carakas Problem is that it uses the data from the instance to put to database. So there are only a few choices to fix the problem.

1. Change instance data (and change it back after the data is saved)
2. Copy all the data from the instance into a second object to save to database (and that second object can be discarded after the database update/insert call)

carakas:
@V3nn3tj3 I think  the latter solution is the best one

WouterSioen:
+1 for the second solution.

jeroendesloovere:
@WouterSioen, can you merge this? 
I just encountered the same "Notice: Array to string conversion" error when trying to save a page.

WouterSioen:
Nope, I won't merge this (yet) for two reasons:

* the styleguide isn't followed
* the feedback isn't handled yet

jeroendesloovere:
@WouterSioen This is a valid bugfix.
@V3nn3tj3 Can you adjust the "spaces" (styleguide) as wouter requested so we can merge this.

src/Backend/Core/Engine/Meta.php
jeroendesloovere: @carakas the database inserting/updating is just below it,
so I see no reason why it may not be serialized at this line?

It is a valid bugfix, I tried it over here, because I had the same problem.

WouterSioen:
Yes, I know it is a valid fix, but I won't merge it without handled feedback and code style fixes.

V3nn3tj3:
I will tomorrow. I'm not considering @carakas his solution. There is no need for that, if you follow the flow of the program you see that the data will never be used again. And chances are slim of it ever beeing used again after it has been send to the database...

It is the design of the class that is bas. As this shouldn't be happing at all if the design was correct. Database principles state that you should always keep database data seperate from model data. In this example the model data is inserted into a database structure.

WouterSioen:
ok, poke me when it's fixed ;)

V3nn3tj3:
@WouterSioen ( •_•)σ

PR #1427, opened 2016-03-04
Load the mailer config for commands

carakas:


src/Common/Mailer/Configurator.php
```diff
@@ -45,4 +45,24 @@ public function onKernelRequest(GetResponseEvent $event)
             // we'll just use the mail transport thats pre-configured
         }
     }
+
+    public function onConsoleCommand(ConsoleCommandEvent $event)
+    {
+        try {
+            $transport = TransportFactory::create(
```
WouterSioen: isn't this a lot of duplicate code? I'd add a "configureMail" method that is called from both listeners.

carakas: fixed in fbd85ed

WouterSioen:
:+1:

PR #1431, opened 2016-03-11
Location: Google API key is available but not used

Katrienvh:
Google API key is available in the setting but not used in de location module


WouterSioen:
This contains all changes from the sumocoders fork

PR #1433, opened 2016-03-14
Add a default robots.txt

carakas:


robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+Disallow: /private
+Disallow: /api
+Disallow: /backend/ajax
+Disallow: /backend/cronjob
```
WouterSioen: I'd also add /frontend/ajax, /api and /install

carakas: fixed in 31623200c8bfafb3f3c3fb22651919b7d3252603

WouterSioen:
I don't know a lot about SEO, so I'll wait to merge this for now (until further feedback)

jeroendesloovere:
Don't know if we should add ```private``` in the list.

Because
1. Google has no knowledge of "/private", so the chance that it will index it is probably zero.
2. This is encouraging hackers, because they now know where the login is.

WouterSioen:
the private pages also have a `<meta name="robots" content="noindex, nofollow" />`  meta tag, so this should avoid indexing by default.

carakas:
@jeroendesloovere fixed in 180c19c

WouterSioen:
if there is no additional feedback, I'll merge this one.

PR #1451, opened 2016-04-08
Cleanup css

DaanDD:


src/Backend/Core/Layout/Css/screen.css
```diff
@@ -8074,8 +8074,8 @@ div.tabs ul li.selected {
 
 .bootstrap-tagsinput {
   display: block;
-  padding: 4px 6px 7px;
-  overflow: hidden; }
+  overflow: hidden;
```
carakas: why is this so important ?

DaanDD: Following the scss-lint standard makes it easier for multiple people to work on the same code.

PR #1453, opened 2016-04-14
Fix datagrid sequence missing column

StijnVrolijk:
The enableSequenceByDragAndDrop() method sets the 'sequence' column to hidden but there's no reason to actually fetch the sequence in most cases. Therefor I made a method to check wether a column is present and used that method to only hide the sequence column if it's present.


WouterSioen:
:+1:

PR #1461, opened 2016-05-02
Bug - hasColumn

V3nn3tj3:


src/Backend/Core/Engine/Meta.php
```diff
@@ -477,6 +477,11 @@ public function save($update = false)
             $custom = null;
         }
 
+        //serialize data for save
+        if (!empty($this->data['data'])) {
+            $this->data['data'] = serialize($this->data['data']);
```
carakas: isn't this a part of an other PR

V3nn3tj3: Yes, it is.
Not a genius when it comes to git, so don't know what happend or how to fix it :(

carakas: The problem is that you start from your own master branch and add your fixes to it, it is better to create a seperate branch for each pull request you want to make and keep your master in sync with the master of fork

carakas: that way your code will come into master when it is merged and won't come in other PR you make

V3nn3tj3: Thanks for the info, are there any other 'tricks'? Can i cancel this PR and create a new one from a seperate branch?

WouterSioen: you could checkout to master, create a new branch, and then cherry-pick the needed commits in there.

PR #1465, opened 2016-05-04
Add utf8mb4 support (emoji support 🎉)

jessedobbelaere:
Noticed that fork cms doesn't use utf8mb4 yet. Read: store U+1F4A9 PILE OF POO (💩) correctly.
Useful for doing things like i did with this PR title: add emoji's and stuff to titles and messages. You could put an emoji in the blog title or in other places for example.

<hr>

### Situation right now: 
Go to the forkcms demo backend and put an emoji somewhere in the middle of a blog post. Mysql will cut the text in half because of the unexpected 4th byte in the character!! You lose the whole text in the blog post! **Checkout this video of the issue: http://d.pr/v/19iHO**

### Solution:
Since MySQL 5.5.3 there is a new encoding released called utf8mb4 which does have full unicode support. MySQL 5.5.3 was released early 2010 so already a long time ago.
We have to use a character set `utf8mb4` instead of utf8, and a collate `utf8mb4_unicode_ci` instead of `utf8_unicode_ci` in forkcms. We also need to enforce the use of multibyte string functions in forkcms. 

- [x] Update sql to use a utf8mb4 charset with utf8mb4_unicode_ci as collate for the core & core modules
- [x] Checked if we use multibyte string functions wherever possible: http://php.net/manual/en/ref.mbstring.php
- [x] Tested if we can use emoji's in pagetitle, blog title, other fields, ... This [works](http://d.pr/i/1cSJL)

#### Warning
There is one thing to be careful about: utf8 characters are 3-bytes, while utf8mb4 chars are 4-bytes. MyIsam can only support 1000 bytes for a key. The most important step is to check the max length of columns & index keys! Well, mysql will throw an error during install of forkcms if thats the case, so i fixed it everywhere

So if you have this example:

```
CREATE TABLE IF NOT EXISTS `search_index` (
 `module` varchar(255) NOT NULL,
 `other_id` int(11) NOT NULL,
 `field` varchar(64) NOT NULL,
 `value` text NOT NULL,
 `language` varchar(5) NOT NULL,
 `active` enum('N','Y') NOT NULL default 'N',
 PRIMARY KEY (`module`,`other_id`,`field`,`language`),
FULLTEXT KEY `value` (`value`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci COMMENT='Search index';
```
Old primary key using utf8 was: 255 * 3 + 11 + 64 * 3 + 5 * 3 = 983 < 1000 bytes so valid.
New primary key using utf8mb4 will be: 255 * 4 + 11 + 64 * 4 + 5 * 4 = 1307 > 1000 bytes so invalid!
Solution: I used a plain old utf8  charset for module names in the install.sql files everywhere: `module varchar(255) CHARACTER SET utf8 NOT NULL,`. This makes a module name the only place where you can't insert emojis 😟  Other solution would be to lower the module name varchar to 245 or 240... Let me know if thats better

#### Read more
More info about utf8mb4:
https://mathiasbynens.be/notes/mysql-utf8mb4
https://florian.ec/articles/mysql-doctrine-utf8/
https://slashdot.io/blog/adding-emoji-support-to-your-blog-948181198


src/Backend/Modules/Search/Installer/Data/install.sql
```diff
@@ -26,14 +24,13 @@ CREATE TABLE IF NOT EXISTS `search_statistics` (
...
  `num_results` int(11) default NULL,
  PRIMARY KEY (`id`)
-) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci AUTO_INCREMENT=1 ;
-
+) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci AUTO_INCREMENT=1 ;
 
 CREATE TABLE IF NOT EXISTS `search_synonyms` (
  `id` int(11) NOT NULL auto_increment,
- `term` varchar(255) NOT NULL,
+ `term` varchar(245) NOT NULL,
```
jessedobbelaere: I lowered the varchar length of a synonym term here, so we can use emoji's to define a search synonymterm. Will this cause an implication somewhere? 

245*4 + 5*4 = 1000 == 1000 max allowed bytes for keys in MyISAM.

WouterSioen: I don't think this will cause any issues.

I would merge this in the 3.10 branch though, since it is backwards compatibility breaking (mysql tables change).

Could you change the target of the pull request?

src/Backend/Core/Installer/Data/install.sql
```diff
@@ -86,25 +86,25 @@ CREATE  TABLE IF NOT EXISTS `backend_navigation` (
...
+) ENGINE=MyISAM  DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci AUTO_INCREMENT=1 ;
 
 
 CREATE TABLE IF NOT EXISTS `hooks_subscriptions` (
-  `event_module` varchar(255) NOT NULL,
-  `event_name` varchar(255) NOT NULL,
-  `module` varchar(255) NOT NULL,
+  `event_module` varchar(255) CHARACTER SET utf8 NOT NULL,
+  `event_name` varchar(255) CHARACTER SET utf8 NOT NULL,
+  `module` varchar(255) CHARACTER SET utf8 NOT NULL,
```
WouterSioen: no utf8mb4 here?

WouterSioen:
Great job 😎

src/Backend/Core/Installer/Data/install.sql
jessedobbelaere: Also because the unique key is larger than the max. allowed 1000 bytes here... :( So if I'd use utf8mb4 here, the query would throw an error.
utf8 takes 3 bytes and utf8mb4 takes 4 bytes
Unique key size with utf8mb4 used: 100 * 4 + 100 * 4 + 100 * 4 = 1200 > 1000

So I changed the 3 fields used in the unique key to utf8, and now it would only take 3 bytes: 100 * 3 + 100 * 3 + 100 * 3 = 900 < 1000
(I only need to change two of them to utf8 actually but I chose to do all 3).

WouterSioen: Ok, I didn't read very thoroughly :p Thats a good point.

PR #1467, opened 2016-05-13
Improve action authorization

StijnVrolijk:
It is far more logical to check if everything is present and authorized for a single action than it is to check if an action is present among all authorized actions including ones we're not even requesting.
This also improves speed by around 30ms (rough estimate).

I also found it more logical to redirect to an error page when trying to access an unavailable action than it is to throw an exception which just results in a 500 error page on production anyway, often confusing end users.

### DIFF
--- a/src/Backend/Core/Engine/Action.php
+++ b/src/Backend/Core/Engine/Action.php
@@ -56,8 +56,8 @@ class Action extends Base\Object
     {
         $this->loadConfig();
 
-        // is the requested action possible? If not we redirect to the error page.
-        if (!$this->config->isActionPossible($this->action)) {
+        // is the requested action available? If not we redirect to the error page.
+        if (!$this->config->isActionAvailable($this->action)) {
             // build the url
             $errorUrl = '/' . NAMED_APPLICATION
                 . '/' . $this->get('request')->get('_locale')
--- a/src/Backend/Core/Engine/Action.php
+++ b/src/Backend/Core/Engine/Action.php
@@ -56,10 +56,15 @@ class Action extends Base\Object
     {
         $this->loadConfig();
 
-        // is the requested action possible? If not we throw an exception.
-        // We don't redirect because that could trigger a redirect loop
+        // is the requested action possible? If not we redirect to the error page.
         if (!$this->config->isActionPossible($this->action)) {
-            throw new Exception('This is an invalid action (' . $this->action . ').');
+            // build the url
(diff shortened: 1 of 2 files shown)


src/Backend/Core/Engine/Base/Config.php
```diff
@@ -127,36 +137,77 @@ public function setModule($module)
...
+            if ($this->isActionDisabled($actionType)) {
+                continue;
             }
+
+            // If the action file is missing, continue on to the next type
+            if (!$this->isActionFilePresent($actionType)) {
+                continue;
+            }
+
+            // The action is not disabled and the file is preset, this is an available action!
```
WouterSioen: `preset`? I guess you mean `present`? 🎁

src/Backend/Core/Engine/Base/Config.php
```diff
@@ -127,36 +137,77 @@ public function setModule($module)
...
+            return true;
         }
 
-        if (is_dir($path . '/Ajax')) {
```
WouterSioen: It's funny this code contained checks for Ajax actions, since this is never executed when doing ajax requests (as far as I can see on first sight)

src/Backend/Core/Engine/Action.php
```diff
@@ -56,10 +56,15 @@ public function execute()
...
 
-        // is the requested action possible? If not we throw an exception.
-        // We don't redirect because that could trigger a redirect loop
-        if (!in_array($this->getAction(), $this->config->getPossibleActions())) {
-            throw new Exception('This is an invalid action (' . $this->getAction() . ').');
+        // is the requested action available? If not we redirect to the error page.
+        if (!$this->config->isActionAvailable($this->action)) {
+            // build the url
+            $errorUrl = '/' . NAMED_APPLICATION
+                . '/' . $this->get('request')->get('_locale')
```
WouterSioen: you can just use `getLocale` instead of `get('_locale')`

src/Backend/Core/Engine/Base/Config.php
jeroendesloovere: @WouterSioen or the reverse

StijnVrolijk: Fixed that in bbcf093

src/Backend/Core/Engine/Action.php
StijnVrolijk: This is what sucks about not using DI :p fixed in 8d7e839

PR #1472, opened 2016-05-18
Send message.

bartdc:
The message was created but never send.


jeroendesloovere:
Would be awesome if someone can create a PR with an EventListener.

bartdc:
It's on my @todo, good idea.

Issue #1474, opened 2016-05-24
Can't add multiple "og:image:height" or "og:image:width" meta properties.


jeroendesloovere:
**What I want**

```
<meta property="og:image" content="http://example.com/rock.jpg" />
<meta property="og:image:width" content="300" />
<meta property="og:image:height" content="300" />
<meta property="og:image" content="http://example.com/rock2.jpg" />
<meta property="og:image:width" content="300" />
<meta property="og:image:height" content="300" />
<meta property="og:image" content="http://example.com/rock3.jpg" />
<meta property="og:image:height" content="1000" />
```
> Means there are 3 images on this page, the first image is 300x300, the middle one is also 300x300, and the last one is 1000px tall. [View source](http://ogp.me/#array)

**Problem**

That is currently not possible in Fork CMS.
Since "property"+"content" `og:image:width|300` already exists the first time, the second time it won't be shown.

This is the false result:
```
<meta property="og:image" content="http://example.com/rock.jpg" />
<meta property="og:image:width" content="300" />
<meta property="og:image:height" content="300" />
<meta property="og:image" content="http://example.com/rock2.jpg" />
<meta property="og:image" content="http://example.com/rock3.jpg" />
```

jeroendesloovere:
I'm working on a backwards fix right now.

PR #1475, opened 2016-05-24
Adding multiple og:image:width and :height is now possible.

jeroendesloovere:
Closes https://github.com/forkcms/forkcms/issues/1474

### DIFF
--- a/src/Frontend/Core/Engine/Header.php
+++ b/src/Frontend/Core/Engine/Header.php
@@ -21,6 +21,7 @@ use Frontend\Core\Engine\Base\Object as FrontendBaseObject;
  *
  * @author Tijs Verkoyen <tijs@sumocoders.be>
  * @author Matthias Mullie <forkcms@mullie.eu>
+ * @author Jeroen Desloovere <jeroen@siesqo.be>
  */
 class Header extends FrontendBaseObject
 {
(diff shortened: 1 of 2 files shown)


src/Frontend/Core/Engine/Header.php
```diff
@@ -333,14 +341,21 @@ public function addOpenGraphData($key, $value, $overwrite = false)
...
      * @param bool   $overwrite Should we overwrite the previous value?
+     * @param integer $width    The width of the image.
+     * @param integer $height   The height of the image.
      */
-    public function addOpenGraphImage($image, $overwrite = false)
+    public function addOpenGraphImage($image, $overwrite = false, $width = 0, $height = 0)
     {
+        // recast width and height
+        $width = (int) $width;
+        $height = (int) $height;
```
WouterSioen: I'd throw an exception if you add both height and width, since you only use one of then.

WouterSioen: Hmm, if you have both width and height, you add the meta data twice? Is this correct?

jeroendesloovere: Yes, doing something like:
```php
$this->header->addOpenGraphImage('my-image-01.jpg', false, 1600, 1200);
```
Will result in
```html
<meta property="og:image" content="my-image-01.jpg">
<meta property="og:image:width" content="1600">
<meta property="og:image:height" content="1200">
```

WouterSioen:
oops, sorry

PR #1477, opened 2016-05-29
Rename upgrade guide

jessedobbelaere:
Last version was 3.9.7 and then we switched to fork cms 4.0. So upgrade guide should be renamed to 4.0

Also should have some info about using twig in backend and frontend, the best practices and things to watch out for so if anyone has some input on this, let me know because i'm not yet familiar with the forkcms Twig syntax


carakas:
👍

PR #1490, opened 2016-06-03
Fixed some UI bugs

freshface:


src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
```diff
@@ -309,7 +309,7 @@
                             <span class="fa fa-times"></span>{{ 'lbl.Cancel'|trans|capitalize }}
                         </button>
                         <button id="changeTemplateSubmit" type="button" class="btn btn-primary">
-                            <span class="fa fa-plus-square"></span>{{ 'lbl.Add'|trans|capitalize }}
+                            <span class="fa fa-check"></span>{{ 'lbl.OK'|trans|capitalize }}
```
WouterSioen: Wasn't "add" more logical? You add a template, don't you?

WouterSioen:
Looks good at first sight 👍

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
freshface: You change the template in mu opinion :) because one is allready selected

PR #1491, opened 2016-06-06
Change urlencoding from RFC 1738 to RFC 3986

jessedobbelaere:
Thought it was a good idea to switch to RFC 3986 (% encoding) instead of RFC 1738 (+ encoding)? 
http://stackoverflow.com/questions/996139/urlencode-vs-rawurlencode#comment805150_996161

#### Difference in encoding standards
RFC 1738 encodes url's with spaces as plus (+) signs. Spec from 1994
RFC 3986 encodes url's with spaces as percent encoded (%20). Modern spec (2005) but not used as default in PHP :)

#### http-build-query
http://php.net/manual/en/function.http-build-query.php
Uses RFC-1738 by default, so we have to add RFC 3986 in the parameters 😐 

#### urlencode vs rawurlencode
http://php.net/manual/en/function.urlencode.php
http://php.net/manual/en/function.rawurlencode.php

URLENCODE - encodes a space as a plus sign '+'
RAWURLENCODE - encodes a space as '%20'


WouterSioen:
There are two tests containing the old decoding: https://travis-ci.org/forkcms/forkcms/jobs/135716795

jessedobbelaere:
@WouterSioen fixed the tests :-)

PR #1494, opened 2016-06-07
Don't use the source backend avatar. - closes #1487

### LINKED ISSUE #1487: Users add/edit doesn't create a source file for avatar
The source file is used though.

2 possible solutions:

* keep the source
* use a cropped version

I think using the cropped version is better, because people can upload really big images as avatar

WouterSioen:
This file isn't even created (except for in the installer).


carakas:
👍

PR #1495, opened 2016-06-07
Add missing / at the end of .gitignore line

carakas:


.gitignore
```diff
@@ -17,4 +17,5 @@
 /bundles/
 /js/
 /css/
-/bin
+/bin/
+
```
WouterSioen: why the extra line?

.gitignore
```diff
@@ -17,4 +17,5 @@
 /bundles/
 /js/
 /css/
-/bin
+/bin/
```
WouterSioen: for folders, ending with `/*` is even better.

.gitignore
WouterSioen: while your at it, `/app/sessions` can be removed

.gitignore
carakas: fixed in ee5a557deb5f231abc61755abae82145d8760050

.gitignore
carakas: fixed in 46eff83

.gitignore
```diff
@@ -4,17 +4,16 @@
...
-/bin/
-/bundles/
-/js/
-/css/
-/bin
+/bin/*
+/bundles/*
+/js/*
+/css/*
+/bin/*
```
jeroendesloovere: This folder is in this file twice...

WouterSioen: hmm, good catch.

carakas: nice one :)

WouterSioen: fixed in 836156f26a0469882cd9d0fd2a2336e279c79120

PR #1507, opened 2016-06-07
Add the missing god.jpg to the gitignore for the backend_users avatars

carakas:


WouterSioen:
This should be created during installation. https://github.com/forkcms/forkcms/blob/master/src/Backend/Modules/Users/Installer/Installer.php#L40-L62

PR #1510, opened 2016-06-08
getURL for 404 needs to account for the language.

freshface:


src/Backend/Core/Engine/Model.php
```diff
@@ -643,7 +643,7 @@ public static function getURLForBlock($module, $action = null, $language = null)
 
         // still no page id?
         if ($pageIdForURL === null) {
-            return self::getURL(404);
+            return self::getURL(404, $languag);
```
carakas: this should be `$language` I gues ?

carakas: but apart from that, isn't it a bit strange that we need to pass the language to show a 404 page ?

WouterSioen: Since language is a parameter of the method, passing it on is logical. (it will use the current language by default)

PR #1512, opened 2016-06-08
Faq.js: On Safari, "yes" was never true.

jeroendesloovere:
**Problem**
Faq.js: On Safari, "yes" was never true.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 
 Bugfixes:
 
-* Faq: Fixed frontend JS which did not found if "yes" is checked.
 * Core: Fixes date field with empty date.
 * Core: Fix lost ucfirst template modifier.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 Bugfixes:
 
+* Faq: Fixed frontend JS which did not found if "yes" is checked.
```
WouterSioen: this will not be included in 4.0.1 anymore

it's in fact easier when you don't add a changelog, since this gives merge conflicts all the time (and I add everything in here when releasing)

jeroendesloovere: I'm sorry, I forgot a 4.0.2.
Ok, I'll remove it from changelog, don't you forget to add it when merging it.

PR #1513, opened 2016-06-08
Backend: Core fix when using extra-params in a datagrid sequence reorder ajax call.

jeroendesloovere:
Closes #1403, because has merging conflicts

### DIFF
--- a/src/Backend/Core/Js/backend.js
+++ b/src/Backend/Core/Js/backend.js
@@ -1974,8 +1974,18 @@ jsBackend.tableSequenceByDragAndDrop =
                         var module = (typeof $table.parents('table.jsDataGrid').data('module') == 'undefined') ? jsBackend.current.module : $table.parents('table.jsDataGrid').data('module').toString();
 
                         // fetch extra params
-                        if (typeof $table.parents('table.jsDataGrid').data('extra-params') != 'undefined') extraParams = $table.parents('table.jsDataGrid').data('extra-params');
-                        else extraParams = {};
+                        if (typeof $table.parents('table.jsDataGrid').data('extra-params') != 'undefined') {
+                            // we define extra params
+                            extraParams = $table.parents('table.jsDataGrid').data('extra-params');
+
+                            // we convert the unvalid {'key':'value'} to the valid {"key":"value"}
+                            extraParams = extraParams.replace(/'/g, '"');
+
+                            // we parse it as an object
+                            extraParams = $.parseJSON(extraParams);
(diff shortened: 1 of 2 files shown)


src/Backend/Core/Js/backend.js
```diff
@@ -1974,8 +1974,18 @@ jsBackend.tableSequenceByDragAndDrop =
...
 
                         // fetch extra params
-                        if (typeof $table.parents('table.jsDataGrid').data('extra-params') != 'undefined') extraParams = $table.parents('table.jsDataGrid').data('extra-params');
-                        else extraParams = {};
+                        if (typeof $table.parents('table.jsDataGrid').data('extra-params') != 'undefined') {
+                            // we define extra params
+                            extraParams = $table.parents('table.jsDataGrid').data('extra-params');
+
+                            // we convert the unvalid {'key':'value'} to the valid {"key":"value"}
+                            extraParams = extraParams.replace(/'/g, '"');
```
WouterSioen: isn't it cleaner to fix this in php instead of in js?

jeroendesloovere: @WouterSioen That is not possible, since '"' breaks html.

For example:

You have the following in your Backend Action:
```php
// enable sorting
$this->dgPosts->enableSequenceByDragAndDrop();
$this->dgPosts->setAttributes(
    array(
        'class' => 'dataGrid sequenceByDragAndDrop',
        'data-extra-params' => "{'module':'Products', 'type':'" . $this->category['type'] . "', 'category_id':'" . $this->categoryId . "'}",
        'data-module' => 'Categories',
        'data-action' => 'SequenceForItems'
    )
);
```
> You can not use ", because this breaks the html. So you must use `.

In older jQuery versions this wasn't any problem.
But in the newer jQuery version I noticed that the json validation has become stricter.

**So we must convert ` to " in our javascript.** (=> which is exactly what this PR does).

===
SCREENSHOTS
**Before fix**
schermafbeelding 2016-02-18 om 10 20 41

**After fix**
schermafbeelding 2016-02-18 om 10 20 26

WouterSioen: but you can replace " with `&quot;` in your json? I believe this is accepted by browsers?

jeroendesloovere:
@WouterSioen ah, you already merged it. Didn't found time to checkout your proposed other solution.

WouterSioen:
Well, I had to release 4.0.3 because of the mysql bug, so I tested and merged this already :)

PR #1515, opened 2016-06-10
Include our bower_components in our repository.

WouterSioen:
This fixes two things:

* It makes fork more approachable (no npm needed)
* It makes sure versions are always exactly the same.

We used bowerphp in the demo site, but this didn't respect the version
constraints and gave issues.


mathiashelin:
:shipit:

PR #1518, opened 2016-06-10
Added zoom-level 1 and 2

jeroendesloovere:
**Problem**
I've got clients that have locations all over the world but the zoom-level only starts from 3.

**Solution**
You can now select 1 and 2 from the zoom-level.

Closes #1394


WouterSioen:
looks good to me 👍 (for in 4.1.0)

PR #1519, opened 2016-06-10
Fixed 404 breadcrumb for single language sites 

jacob-v-dam:


src/Frontend/Core/Engine/Url.php
```diff
@@ -364,9 +364,12 @@ private function processQueryString()
 
             // remove language
             if ($hasMultiLanguages) {
-                $URL = trim(str_replace('/' . $language, '', $URL), '/');
+                $URL = str_replace('/' . $language, '', $URL);
             }
 
+            // remove the first slash
+            $URL = trim($URL, '/');
```
WouterSioen: if it's only the first slash, wouln't `ltrim` be better?

jacob-v-dam: It would be better but with the original it was:
`$URL = trim(str_replace('/' . $language, '', $URL), '/')`

WouterSioen: ok. I'll test this one before merging it, because I can't predict the consequences of this change without testing. Thanks for the PR!

WouterSioen:
thx!

PR #1527, opened 2016-06-16
Fixes uppercase extension browser cache in .htacccess

### LINKED ISSUE #1448: File browser caching not working with uppercase extensions
When you have a file named .JPG the .htaccess will not use browser caching because it is case sensitive.

jacob-v-dam:
Fixes #1448


carakas:
@jacob-v-dam since you only have one actual change it might be a good idea to squash your commits

WouterSioen:
@carakas I can squash during merge, that's not really an issue ;)

PR #1533, opened 2016-06-17
Remove text that shouldn't be there

carakas:


carakas:
it breaks the error page in the backend

PR #1534, opened 2016-06-18
Refactor analytics

carakas:
I cleaned up the insides of the analytics module a bit.
I don't think I have changed public functions so this shouldn't be a breaking change.

I also made it possible to reset once a .P12 has been uploaded. This makes it easier to reset if a mistake (for instance in the email address) has been made


src/Backend/Core/Engine/Form.php
```diff
@@ -305,7 +305,7 @@ public function isCorrect($revalidate = true)
     /**
      * Parse the form
      *
-     * @param $tpl The template instance wherein the form will be parsed.
+     * @param $tpl TwigTemplate The template instance wherein the form will be parsed.
```
WouterSioen: isn't this ins the wrong order here?

src/Backend/Modules/Analytics/Form/DateRangeType.php
```diff
@@ -0,0 +1,138 @@
...
+ */
+final class DateRangeType
+{
+    /** @var Form */
+    private $form;
+
+    /**
+     * @var DateRange $dateRange
+     */
+    private $dateRange;
```
WouterSioen: inconsistent `@var` styles

src/Backend/Modules/Analytics/Form/SettingsStepType.php
```diff
@@ -0,0 +1,32 @@
...
+use Backend\Core\Engine\TwigTemplate;
+use Common\ModulesSettings;
+use Google_Service_Analytics;
+
+/**
+ * An interface so we can split the settings form up into multiple forms
+ *
+ * @author Jelmer Prins <jelmer@sumocoders.be>
+ */
+interface SettingsStepType
```
WouterSioen: what's the use case for this interface if you never typehint on it?

src/Backend/Modules/Analytics/Form/SettingsStepType.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * @param string $name
+     * @param ModulesSettings $settings
+     * @param Google_Service_Analytics $googleServiceAnalytics
+     */
+    public function __construct($name, ModulesSettings $settings, Google_Service_Analytics $googleServiceAnalytics);
```
WouterSioen: I don't think including a constructor in an interface is a best practice.

src/Backend/Core/Engine/Form.php
carakas: fixed in a3846a0

src/Backend/Modules/Analytics/Form/DateRangeType.php
carakas: fixed in 5369819

src/Backend/Modules/Analytics/Form/SettingsStepType.php
carakas: I do typehint to it https://github.com/forkcms/forkcms/pull/1534/files#diff-93d9fd5c110ab84c96cc9498a4391456R18

src/Backend/Modules/Analytics/Form/SettingsStepType.php
carakas: ok, removed in 6dacf34

src/Backend/Modules/Analytics/Form/SettingsStepType.php
WouterSioen: a phpdoc is not a typehint ;)

WouterSioen:
merge conflicts alert! ⚠️

Issue #1488, opened 2016-06-02
Fork 4.0.0 new UI issues


freshface:
- [x] Single element in panel http://jmp.sh/aJTJXnp
- [x] datagrid Blog/ Comments , column to wide http://jmp.sh/mRV4ehE
- [x] Formbuilder field icons http://jmp.sh/oxaQcef
- [x] Installer: previous button is underlined http://jmp.sh/NKJfgZ0
- [x] Installer: required asterisk should not be an icon http://jmp.sh/O3VTmZb
- [x] Installer: password strength http://jmp.sh/5UbTMoK update, should be same as http://jmp.sh/CrtSkte
- [x] Different use of save icons: http://jmp.sh/OHEnIq2 and http://jmp.sh/PyD8awU
- [x] Label when changing template: http://jmp.sh/ErDS0jP
- [x] Use of spaces http://jmp.sh/czAXSzv
- [x] thead hover state http://jmp.sh/QikVpyz
- [x] Should be a link http://jmp.sh/Ucw8KtN
- [x] Translations bug, no results without value http://jmp.sh/nuJRYb7
- [x] tables without a hover state http://jmp.sh/GJF4Q4w
- [x] Pages: template choice http://jmp.sh/cVbQQD4
- [x] Buttons http://jmp.sh/6RpkEto
- [x] Themes add block http://jmp.sh/6RpkEto
- [x] Themes delete block icon not sharp http://jmp.sh/vBosoxw
- [x] Missing translation http://jmp.sh/bqk1faQ
- [x] After installing a module via the cms the system logs you out.

jessedobbelaere:
In addition: 
* Installer: remove large space between label and field http://jmp.sh/O3VTmZb
* Installer: padding bottom under the buttons is small, while padding left&right of the form is larger. Make padding consistent
* Installer: Could use more blue instead of gray to make it less boring. Maybe a blue tint in the background color or somewhere, or put a logo of fork somewhere

freshface:
- Forgot password is not working when trying to login.

carakas:
what is wrong with using an icon as asterisk ?

freshface:
@carakas It is not consistent with the asterisk in the backend http://jmp.sh/LiUc8yo + it looks to big

carakas:
any update on this ?

carakas:
I've turned the list into a checklist, I'm doing some things now and checking them off as I go along

carakas:
[...]

PR #1542, opened 2016-06-21
Small layout fixes

jessedobbelaere:
Several small layout fixes. 

#### Summary
* I don't think `<a href="{{ geturlforblock:'Profiles':'Settings' }}"` has ever worked, doesn't look right? Syntax should be `<a href="{{ geturlforblock('Profiles', 'Settings') }}"`
* Indentation fixes + use spaces instead of tabs
* Move theme dropdown to the left of the template path: http://d.pr/i/18mRa
* When adding a new block in the theme template editor, it would add the new block on the first position. The way we calculate the current index is broken. Should be fixed now (proof): http://d.pr/v/1iqhX
* ~~Add a new template and you see a broken helper-text translation: http://d.pr/i/1a2yw~~ Seems already fixed by freshface in #1488
* When you add a new block (editor/module/widget) on a page, we used to have "editor" selected by default. Fixed that. Issue as seen here: http://d.pr/i/1iLkP

I'll leave this open for some days while I fix other issues I come along

### DIFF
--- a/src/Backend/Modules/Extensions/Js/ThemeTemplate.js
+++ b/src/Backend/Modules/Extensions/Js/ThemeTemplate.js
@@ -39,7 +39,7 @@ jsBackend.template =
 
 		// fetch position & block index
 		var positionIndex = $(this).closest('.jsPosition').find('input[id^=position]').attr('id').replace('position', '');
-		var blockIndex = $(this).prevAll('.jsBlock').length;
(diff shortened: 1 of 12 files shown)


WouterSioen:
Thanks! Give me a poke when all issues are fixed and I can merge this.

WouterSioen:
@jessedobbelaere I'll already merge this.

PR #1545, opened 2016-06-23
1544 content blocks dont work in fork 4

carakas:


src/Frontend/Core/Engine/TwigTemplate.php
```diff
@@ -56,6 +57,13 @@ class TwigTemplate extends BaseTwigTemplate
     private $baseFile;
 
     /**
+     * Base file location
+     *
+     * @var string
+     */
+    private $baseSpoonFile;
```
WouterSioen: what's up with this? Was this missing?

src/Frontend/Modules/ContentBlocks/Widgets/Detail.php
```diff
@@ -14,69 +12,25 @@
...
-    {
-        // assign data
-        $this->tpl->assign('widgetContentBlocks', $this->item);
+        // set a default text so we don't see the template data
+        if (empty($contentBlock)) {
+            $contentBlock['text'] = '';
+        }
+
+        $this->tpl->assign('widgetContentBlocks', $contentBlock);
+        // That's all folks!
```
WouterSioen: 😂

src/Frontend/Core/Engine/TwigTemplate.php
carakas: yup, it was created dynamically (saw the error in my ide while working on this

src/Frontend/Modules/ContentBlocks/Widgets/Detail.php
carakas: 👯

WouterSioen:
I'll give this a :shipit:

src/Backend/Modules/ContentBlocks/Engine/Model.php
```diff
@@ -22,6 +22,7 @@
  * @author Matthias Mullie <forkcms@mullie.eu>
  * @author Dieter Vanden Eynde <dieter.vandeneynde@netlash.com>
  * @author Jeroen Desloovere <jeroen@siesqo.be>
+ * @author Jelmer Prins <jelmer@sumocoders.be>
```
jessedobbelaere: Still find it strange that we do this credits thing in Fork, while we can do git blame :P ? If we would have 100's of collaborators on git this would be funny to see it grow

carakas: true
I normally don't do this but hey if they are on there why not me :D

WouterSioen: yes, I'm 👍  of deleting all this stuff in favour of the git blame stuff.

carakas: likewise :)

carakas: removed in https://github.com/forkcms/forkcms/pull/1553

PR #1548, opened 2016-06-24
Fixing bug in imageDelete when building the full path of the file

dmoerman:
There is a `/` missing in front of the subdirectory if there is one specified
Also removed array keys, so that the passed array can just contain the names of the folders of the files sizes.
For example: `[ '64x64' ]` instead of `[ '64x64' => null ]`


jeroendesloovere:
Hi @dmoerman, thanks for you PR.

PR #1553, opened 2016-06-26
Remove @author tags since git blame is more accurate

carakas:
In https://github.com/forkcms/forkcms/pull/1545#discussion_r68194791 the comment was made that the @author tag could be stripped from fork cms
because we already have `git blame`. Not everybody adds it and if they would
we would get long lists in the beginning of every file. This makes it more like clutter
So this commit removes the author tags from the entire fork cms code base.
I didn't remove them from libraries that have been baked in


jessedobbelaere:
👌👍

carakas:
this must have been my most boring contribution so far 😴

PR #1565, opened 2016-06-29
Fix the placeholders in the formbuilder. Fixes #1559

### LINKED ISSUE #1559: Placeholders formbuilder
There is a bug for having placeholders in textfields etc. 
@WouterSioen said that the rendering of the form tags has changed. It cloud be some issue in there.

WouterSioen:
They weren't used.

### DIFF
--- a/src/Frontend/Modules/FormBuilder/Widgets/Form.php
+++ b/src/Frontend/Modules/FormBuilder/Widgets/Form.php
@@ -214,6 +214,9 @@ class Form extends FrontendBaseWidget
                     if (isset($field['validations']['email'])) {
                         $txt->setAttribute('type', 'email');
                     }
+                    if ($item['placeholder']) {
+                        $txt->setAttribute('placeholder', $item['placeholder']);
+                    }
 
                     // get content
                     $item['html'] = $txt->parse();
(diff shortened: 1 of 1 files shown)


carakas:
👍

src/Frontend/Modules/FormBuilder/Widgets/Form.php
```diff
@@ -214,6 +214,9 @@ private function loadForm()
                     if (isset($field['validations']['email'])) {
                         $txt->setAttribute('type', 'email');
                     }
+                    if ($item['placeholder']) {
```
carakas: is it always available or would an isset like above be better ?

src/Frontend/Modules/FormBuilder/Widgets/Form.php
```diff
@@ -271,6 +274,9 @@ private function loadForm()
                     if ($item['required']) {
                         $txt->setAttribute('required', null);
                     }
+                    if ($item['placeholder']) {
```
carakas: idem

src/Frontend/Modules/FormBuilder/Widgets/Form.php
WouterSioen: it's always set (on line 151)

src/Frontend/Modules/FormBuilder/Widgets/Form.php
WouterSioen: it's always set (on line 151)

carakas: in that case :shipit:

PR #1574, opened 2016-06-29
Use |raw for pagetitle to avoid html entities

jessedobbelaere:
If you use a pagetitle "Test & Test", it would render as "Test &amp; Test" in the triton theme


carakas:
👍

jeroendesloovere:
Hmmm, I was wondering if we couldn't do this any better? Don't like it when we have to add |raw everywhere... Seems so stupid for a modern template engine

WouterSioen:
@jeroendesloovere that's the safest way to avoid injection bugs.

PR #1577, opened 2016-06-30
Fixes production environment to only log errors in /app/logs/prod.log

jeroendesloovere:
**We must only show errors in production log**

Because otherwise everything will be logged in production modus and
we end up getting a log that is > 100 MB and eventually will exceed the
maximum memory limit **and will crash the website**.
(This happened to me in one website, that's how I noticed.)


app/config/config_dev.yml
```diff
@@ -7,6 +7,13 @@ web_profiler:
     toolbar: true
     intercept_redirects: false
 
+monolog:
+    handlers:
+        main:
+            type:  stream
+            path:  %site.path_www%/app/logs/%kernel.environment%.log
+            level: debug
+
```
WouterSioen: this is not needed in `_dev` because it's in the main config file https://github.com/forkcms/forkcms/blob/master/app/config/config.yml#L27-L32

PR #1579, opened 2016-06-30
Update spoon library.

### LINKED ISSUE #1556: 27.6.2016 as dateformat does not work with SpoonFormDate

WouterSioen:
It contains a lot of new features and bugfixes.

Fixes #1556


carakas:
just tested it and it indeed fixes the not supported date format bug

PR #1581, opened 2016-06-30
Replace sprintf with format in templates

### LINKED ISSUE #1567: sprintf is used in some templates, format should be used
http://twig.sensiolabs.org/doc/filters/format.html

replace sprintf with format in the templates and depricate sprintf

carakas:
fixes #1567

### DIFF
--- a/src/Common/Core/Twig/Extensions/TwigFilters.php
+++ b/src/Common/Core/Twig/Extensions/TwigFilters.php
@@ -56,7 +56,7 @@ class TwigFilters
         $twig->addFilter(new Twig_SimpleFilter('is_array', 'is_array'));
         $twig->addFilter(new Twig_SimpleFilter(
             'sprintf',
-            'sprintf',
+            array(__CLASS__, 'deprecatedSprintf'),
             array('is_safe' => array('html'))
         ));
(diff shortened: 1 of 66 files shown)


src/Common/Core/Twig/Extensions/TwigFilters.php
```diff
@@ -109,4 +109,20 @@ public static function getFilters(&$twig, $app)
...
+    /**
+     * Show deprecated error for sprintf.
+     *
+     * @return string
+     */
+    public static function deprecatedSprintf()
+    {
+        trigger_error(
+            '|sprintf is deprecated.
+             Use |format(args)|raw instead',
```
WouterSioen: can't sprintf be used without raw?

WouterSioen:
👍

src/Common/Core/Twig/Extensions/TwigFilters.php
carakas: our implementation of sprintf did the |raw automatically
Ill make it more clear that the |raw is only needed if you want to include html

carakas: added in ab83a50

WouterSioen:
@carakas merge conflicts!

carakas:
@WouterSioen fixed

PR #1584, opened 2016-06-30
Just return if getting the navigation html goes wrong

### LINKED ISSUE #1575: getnavigation crashes when having only a home page AND setting parent to 1
Exception is thrown which states that the parent doesn't exist.
Possible solution: return empty navigationlist.

carakas:
This fixes #1575

### DIFF
--- a/src/Frontend/Core/Engine/TemplateModifiers.php
+++ b/src/Frontend/Core/Engine/TemplateModifiers.php
@@ -91,7 +91,7 @@ class TemplateModifiers extends BaseTwigModifiers
             $return = (string) Navigation::getNavigationHTML($type, $parentId, $depth, $excludeIds, $template);
         } catch (Exception $e) {
             // if something goes wrong just return as fallback
-            return;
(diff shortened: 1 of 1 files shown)


src/Frontend/Core/Engine/TemplateModifiers.php
```diff
@@ -87,7 +87,12 @@ public static function getNavigation(
         }
 
         // get HTML
-        $return = (string) Navigation::getNavigationHtml($type, $parentId, $depth, $excludeIds, $template);
+        try {
+            $return = (string) Navigation::getNavigationHTML($type, $parentId, $depth, $excludeIds, $template);
+        } catch (Exception $e) {
+            // if something goes wrong just return as fallback
+            return;
```
WouterSioen: the phpdoc states `return string`. I would return an empty string here to keep consistency

carakas: fixed in 5a38785

PR #1590, opened 2016-06-30
Only show the tags tab when the user has permission

### LINKED ISSUE #1562: New group get not allowed error on pages module
I created a new group with all rights active on the pages module? I can edit pages but still there is an that the module is not allowed.

carakas:
If the user had permission to edit/add pages but no access to the tags module
an alert with the error module not allowed was shown.
This was because the ajax call to get the tags would fail.
Now we only show that tabs if the user has sufficient permissions.

fixes #1562


WouterSioen:
@carakas merge conflicts!

carakas:
@WouterSioen fixed

PR #1594, opened 2016-07-01
.gitignore fixes

vytsci:


.gitignore
```diff
@@ -1,17 +1,22 @@
+# Configuration
+/app/config/parameters.yml
+
+# Libraries
+/bower_components
+/vendor
```
WouterSioen: the `/<folder>/*` format has better support in command line tools (like `ack` and `ag` for example)

.gitignore
```diff
@@ -1,17 +1,22 @@
+# Configuration
+/app/config/parameters.yml
+
+# Libraries
+/bower_components
```
carakas: we need bower_components for the zip install

vytsci: can u explain in more detail? :D

WouterSioen: We included both the bower components and the compiled sass files for the Fork backend in the repository for multiple reasons:

* easier for everyone to set up fork without the need of compilation/installation tools
* no issues during deployment (needing to bower install and compile sass)
* easier to create the zip download on the fork cms website

PR #1597, opened 2016-07-01
set module js files to minify and remove timestamp

vytsci:


WouterSioen:
why removing the timestamp?

carakas:
why would you remove the timestamp ? this can help for caching

carakas:
I would also not minify when the site is in debug mode

vytsci:
Sorry did not see that timestamp is last modified. Thought its preventing js for caching with current time timestamp.

vytsci:
Well this should be done in other place. Withing minifying algorithm.

vytsci:
and there it is
// no minifying when debugging
        if ($this->getContainer()->getParameter('kernel.debug')) {
            $minify = false;
        }

carakas:
ok nice, I did not know that 👍 awesome :D

carakas:
did you test in all the core modules that the minifying doesn't result in javascript errors ? if that is the case you get my 👍

vytsci:
Nope have not tested anything. Just checked that minifying works. Solving minifying errors is not the case at this point. If minifier is not working properly it should be a different PR with minifier fix.

Issue #1547, opened 2016-06-24
FormBuilder is missing a "only send mail" dropdown-choice


jeroendesloovere:
**Problem**

You can only choose options where everything gets saved to database.

**Feature request**
But sometimes customers don't want this to be saved in the database...

What should be done
- At a new option "only send mail" in dropdown
- Add a new "visible_in_backend" boolean for every form request (so it actually gets saved to the database, but you can't view them in the backend)
  ![schermafbeelding 2016-06-24 om 14 42 42](https://cloud.githubusercontent.com/assets/588616/16338043/494dda70-3a1a-11e6-8cfb-f03d4ba218bd.png)

WouterSioen:
Why would you ever not want it to be stored in the database?

jeroendesloovere:
Clients who are scared of "databases" and "top-secret" company information.

carakas:
if you would implement this I don't think you should secretly save it anyway.

guytarded:
Same idea here as @carakas . Just send is just send, not decieving your client

carakas:
Ill add this to https://github.com/forkcms/forkcms/pull/1602 in order to prevent merge errors

carakas:
since a database change is needed for this it will be added in 5.0.0

PR #1602, opened 2016-07-02
Make it possible to send a confirmation mail from the form builder

carakas:
- [x] add missing locals
- [x] prevent the data from being saved when on email only modus
- [x] show message on submission page that the messages aren't being saved into the database


carakas:
TODO: add implementation for #1547

carakas:
TODO: choose the subject and text that will be in the email to the customer

carakas:
fixes 1053

carakas:
fixes #1053

jeroendesloovere:
When will this be merged?

WouterSioen:
in a major version, since it contains BC breaks.

PR #1604, opened 2016-07-02
Fix install module button showing after module is installed

### LINKED ISSUE #1601: Install button should be removed after installation
If you click on the detail of a module after its installation the install button is still visible
If you click on it you get an error stating that the item does not exist. This is confusing since the module is intalled and does work

carakas:
The check for this was overwritten by the permission of the install action
fixes #1601

### DIFF
--- a/src/Backend/Modules/Extensions/Layout/Templates/DetailModule.html.twig
+++ b/src/Backend/Modules/Extensions/Layout/Templates/DetailModule.html.twig
@@ -95,7 +95,7 @@
       </div>
     </div>
   {% endif %}
-  {% if showExtensionsInstallModule %}
+  {% if showInstallButton %}
     <div class="row fork-module-actions">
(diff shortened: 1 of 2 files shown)


src/Backend/Modules/Extensions/Layout/Templates/DetailModule.html.twig
```diff
@@ -95,7 +95,7 @@
       </div>
     </div>
   {% endif %}
-  {% if showExtensionsInstallModule %}
```
WouterSioen: I think the `showExtensionsInstallModule` check should still remain there too. We only want to show the button if both the module isn't installed and the user has rights to install modules.

WouterSioen: ow, nevermind :D

PR #1605, opened 2016-07-02
Add TemplateModifier showBool alias to DataGridFunctions

### LINKED ISSUE #1599: Add TemplateModifier showBool alias to DataGridFunctions
we often forget where the function is and reimplement it. adding an alias to the datagrid functions that uses the template modifier probably will resolve this

carakas:
fixes #1599


tijsverkoyen:
👍

PR #1607, opened 2016-07-03
Add profiles SecurePage widget

jessedobbelaere:
If you have a normal page with some content on it, there's currently no easy way (as far as I know) to make the page only accessible for logged in users. I was creating something like this in a project, when someone asked me why this wasn't part of the profiles module yet, so here goes a PR...

This small widget can be added anywhere on a page and will check if a user is logged in, else it redirects to the login form.

### DIFF
--- a/src/Frontend/Modules/Profiles/Widgets/SecurePage.php
+++ b/src/Frontend/Modules/Profiles/Widgets/SecurePage.php
@@ -30,9 +30,10 @@ class SecurePage extends FrontendBaseWidget
 
         // Check if we're logged in, else redirect to the login form.
         if (!FrontendProfilesAuthentication::isLoggedIn()) {
+            $queryString = $this->URL->getQueryString();
             throw new RedirectException(
                 'Redirect',
-                new RedirectResponse(Navigation::getURLForBlock('Profiles', 'Login'))
+                new RedirectResponse(Navigation::getURLForBlock('Profiles', 'Login') . '?queryString=' . $queryString)
             );
         }
     }
--- a/src/Frontend/Modules/Profiles/Widgets/SecurePage.php
+++ b/src/Frontend/Modules/Profiles/Widgets/SecurePage.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Frontend\Modules\Profiles\Widgets;
+
+/*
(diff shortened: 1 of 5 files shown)


jessedobbelaere:
It's maybe a bit overkill to make a widget for this, but the other alternative is adding a checkbox in the individual page settings (if profiles module is installed) to restrict access.

WouterSioen:
I've also done this using a widget. This can then be used with "parseWidget" in a "secured" template or coupled to every page. This works quite well imo 👍

src/Frontend/Modules/Profiles/Widgets/SecurePage.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+class SecurePage extends FrontendBaseWidget
+{
+    /**
+     * Execute the extra
+     */
+    public function execute()
+    {
+        parent::execute();
+        $this->loadTemplate();
```
WouterSioen: isn't it possible to have a widget without a template?

src/Frontend/Modules/Profiles/Widgets/SecurePage.php
```diff
@@ -0,0 +1,39 @@
...
+
+        // Check if we're logged in, else redirect to the login form.
+        if (!FrontendProfilesAuthentication::isLoggedIn()) {
+            throw new RedirectException(
+                'Redirect',
+                new RedirectResponse(Navigation::getURLForBlock('Profiles', 'Login'))
```
WouterSioen: I'd add the current url as query string parameter to the login page.

src/Frontend/Modules/Profiles/Widgets/SecurePage.php
jessedobbelaere: I'm afraid not @WouterSioen . When I removed this line, it complained. And there doesn't exist a parameter to pass, to indicate there's no template needed. NULL is the default param and then it uses the template with the same name as the action/widget.

WouterSioen: ok

jessedobbelaere:
@carakas Querystring has been added :-)

PR #1610, opened 2016-07-04
Use help-block instead of text-info

### LINKED ISSUE #1600: Installation contains invisible text
schermafbeelding 2016-07-01 om 20 01 27
schermafbeelding 2016-07-01 om 20 01 31

bjorvack:
closes #1600


carakas:
I would rather fix the css than remove to remove the class name

StijnVrolijk:
@carakas this might affect things elsewhere. I'm also for changing the CSS but be sure to check for usage, @bjorvack

carakas:
@bjorvack even better might be to use the bootstrap class that exists for this `help-block`

WouterSioen:
+1 for help block instead of the custom class

PR #1611, opened 2016-07-04
Add image to page

carakas:
This adds the option to add an image to a page.
This can in the template be used for a header or something else


jessedobbelaere:
Cool, this is a feature I've always wanted 

I think Wijs used that here, some good examples:
http://bit.ly/29emKGk
http://bit.ly/29ia8S3
http://bit.ly/29emMhh

src/Backend/Modules/Pages/Actions/Add.php
```diff
@@ -123,22 +123,34 @@ private function loadForm()
         $this->frm->addRadiobutton(
             'hidden',
             array(
-                 array('label' => BL::lbl('Hidden'), 'value' => 'Y'),
-                 array('label' => BL::lbl('Published'), 'value' => 'N'),
+                array('label' => BL::lbl('Hidden'), 'value' => 'Y'),
+                array('label' => BL::lbl('Published'), 'value' => 'N'),
```
WouterSioen: oh god 👍

WouterSioen:
If I look at this implementation, it looks like the god user has to check this "allow_image" box for all page to make sure other users can upload images? I'd prefer a general setting (like the enable meta navigation setting) to make sure a user can upload images for all pages by default.

mathiashelin:
The image isn't added to the frontend yet.

carakas:
@mathiashelin isn't it just available in the page parameter in the template? Haven't rechecked it but I thought I did.

WouterSioen:
nope, the page is fetched using a query containing all parameters, not a `.*` type of query.
The queries in `Frontend\Core\Engine\Model` in `getPage` en `getPageRevision` should be altered to return the image.

carakas:
I've changed things up a bit, you now set on the template if it has an image, you can also define that in the theme.xml by adding the optional attribute `image="true"`
I've also implemented the feature in triton so now you can change the banner image on the home page

carakas:
I've also noticed that fork doesn't give errors when the file you uploaded is too big, I'll create a fix for that in an other PR

carakas:
no error when file size is too big is fixed in https://github.com/forkcms/forkcms/pull/1690

PR #1614, opened 2016-07-05
Template: |formatphone

jeroendesloovere:
**Problem**
Sometimes you get a number as a phone number and you have to style it.

**Solution**
I've added a Template modifier for this.
{$var|formatphone}


carakas:
this looks too specific for Belgium to me.
It also won't work with international numbers

WouterSioen:
I'm pretty sure about this. https://en.wikipedia.org/wiki/National_conventions_for_writing_telephone_numbers

WouterSioen:
damn boy, I don't where I jumped in know. There are full libraries for formatting phone numbers 😲 https://github.com/giggsey/libphonenumber-for-php

jeroendesloovere:
You're all right. Even in Belgium that code I wrote isn't accurate (056 xx xx xx, 09 xxx xx xx).
What should we do? 
Maby we should close this issue as it is domain logic related.

carakas:
I think that might be the best option, or use a library for that, but I would use a custom implementation for it

WouterSioen:
uhu, I think this is project specific and should probably not be added to fork itself.

PR #1616, opened 2016-07-05
Increase minimum php version to 5.6

carakas:
PHP 5.5 reaches EOL on 10 Jul 2016, after that we no longer support it for security reasons


jessedobbelaere:
Nice 👌

(haters gonna hate)

WouterSioen:
this was on my planning for the 10th, you beat me to it :D

WouterSioen:
this is BC breaking though, so this should be done in a major release...

carakas:
well, lucky 5.5 users :p they get extra support just because we recently released 4.0.0 :p

WouterSioen:
if we upgrade to fork 5.0.0 after 1 january, we should even bump to php 7

PR #1619, opened 2016-07-07
Get url for block take data into account

carakas:
You can add extra data to a module extra.
With this you can reuse your action by having different output based on that extra data
Until now it was not possible to specify the version of that action when using the function getURLForBlock

This PR will solve that problem


jeroendesloovere:
Nice!

Can you create a PR for this openstanding bug?
https://github.com/forkcms/forkcms/pull/592/files

tijsverkoyen:
👍

PR #1620, opened 2016-07-08
Add the ability to change the page parameter in the pagination query

jacob-v-dam:


WouterSioen:
Shouldn't the place where the current page gets fetched be changed too?

jacob-v-dam:
I wouldn't break compatibility, I left the blog module as it is. The module uses 'page' as the parameter query. I made it optional to change the query, by default it will be 'page'.

WouterSioen:
ok 👍

tijsverkoyen:
👍

PR #1623, opened 2016-07-12
Make the formatCurrency modifier use formatNumber. Fixes #1531

### LINKED ISSUE #1531: formatCurrency isn't listening to 'Core' 'number_format' setting.
**Problem**

In the backend you can choose the way to show "numbers" in the frontend.
But the currencies aren't looking to this, would be nice if you can change this in the backend.
Since now, you can only do this hardcoded in the method `formatCurrency` in `src/Frontend/Core/Engine/TemplateModifiers.php`.

WouterSioen:
This way, the configured number format is used instead of a default one.

Note that I didn't add localizedcurrency yet, since this adds a lot of new features which shouldn't be added in a bugfix release.

### DIFF
--- a/src/Backend/Core/Engine/TemplateModifiers.php
+++ b/src/Backend/Core/Engine/TemplateModifiers.php
@@ -50,21 +50,24 @@ class TemplateModifiers extends BaseTwigModifiers
 
     /**
      * Format a number
-     * syntax: {$var|formatnumber}
+     *    syntax: {{ $string|formatnumber($decimals) }}
      *
-     * @param float $var The number to format.
+     * @param float $number The number to format.
+     * @param int $decimals The number of decimals
      *
      * @return string
      */
-    public static function formatNumber($var)
+    public static function formatNumber($number, $decimals = null)
     {
-        $var = (float) $var;
+        $number = (float) $number;
 
         // get setting
         $format = Authentication::getUser()->getSetting('number_format', 'dot_nothing');
 
         // get amount of decimals
-        $decimals = (mb_strpos($var, '.') ? mb_strlen(mb_substr($var, mb_strpos($var, '.') + 1)) : 0);
+        if ($decimals === null) {
+            $decimals = (mb_strpos($number, '.') ? mb_strlen(mb_substr($number, mb_strpos($number, '.') + 1)) : 0);
+        }
 
         // get separators
         $separators = explode('_', $format);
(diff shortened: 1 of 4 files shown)


src/Backend/Core/Engine/TemplateModifiers.php
```diff
@@ -50,21 +50,24 @@ public static function formatFloat($number, $decimals = 2)
 
     /**
      * Format a number
-     * syntax: {$var|formatnumber}
+     *    syntax: {{ $string|formatnumber($decimals) }}
```
jeroendesloovere: string?

WouterSioen: it can accept strings (like the format_number php function)

jeroendesloovere: Indeed, but its strange that the lines below say that it's a float
```
@param float $number The number to format.
```

src/Common/Core/Twig/Extensions/BaseTwigModifiers.php
```diff
@@ -37,7 +37,24 @@ public static function formatCurrency($string, $currency = 'EUR', $decimals = nu
             default:
         }
 
-        return $currency.'&nbsp;'.number_format((float) $string, $decimals, ',', '&nbsp;');
+        return $currency.'&nbsp;'.static::formatNumber($string, $decimals);
```
jeroendesloovere: Shouldn't there be spaces between variables?

WouterSioen: well, PSR2 doesn't state anything about this, and symfony coding styles do it this way.

jeroendesloovere: When using php-cs-fixer --level=psr2, spaces are used.
I prefer spaces, since everywhere in Fork we use spaces.

WouterSioen: `php-cs-fixer --level=psr2` won't say a thing about this too ;) It doesn't care since PSR-2 doesn't say a thing about this.

jeroendesloovere: When doing php-cs-fixer fix without --level=psr2, then spaces are getting removed between variables. If you add --level=psr2, they remain ;-P

PR #1626, opened 2016-07-12
Make a macro for the required asterisk and tooltip.

WouterSioen:


src/Backend/Modules/ContentBlocks/Layout/Templates/Add.html.twig
```diff
@@ -11,15 +11,15 @@
...
+          {{ macro.required }}
         </label>
         {% form_field title %} {% form_field_error title %}
       </div>
       <div class="panel panel-default panel-editor">
         <div class="panel-heading">
           <label for="text" class="control-label">
             {{ 'lbl.Content'|trans|ucfirst }}
-            <abbr data-toggle="tooltip" title="{{ 'lbl.RequiredField'|trans|ucfirst }}">*</abbr><br>
+            {{ macro.required }}<br>
```
carakas: is there a reason there is a `<br>` tag here?

WouterSioen: it was already there. I only changed the required tooltip with the macro and didn't touch other stuff.

PR #1628, opened 2016-07-12
Add yml to editorconfig

tommyvdv:


.editorconfig
```diff
@@ -4,71 +4,24 @@
 
 root = true
 
-[*.php]
+[*.{js,php,py,yml}]
```
WouterSioen: I believe we'd split this up because there was poor support at the time for this format. Does this work well now?

PR #1634, opened 2016-07-19
Allow get parameters in og:image

### LINKED ISSUE #1630: Make file system check more robust for checking og:image urls
In https://github.com/forkcms/forkcms/blob/master/src/Frontend/Core/Engine/Header.php#L346 a check is made to verify if a given image really exists on the file system.
This seems a sane check, but it poses a problem:
Facebook fetches the images served via og:image and stores it on its own servers and while it is possible to make Facebook re-scrape urls and so make it "clear" its cache, it does _not_ so for images, it'll encounter the same url in og:image and serve that from its own servers.
This is a problem in the case we need to update an image served via og:image because if you have added it via Header#addOpenGraphImage(), you cannot, for example add a version via GET parameter... 
It does not seem like a bad idea to either allow for this, or make the file system check more robust so that time stamps can or cannot be added.
I have labeled it a bug because of the fact that, IMO, GET params should be able to be added in some form or other...
Thoughts?  Fixes?  PR's?

WouterSioen:
Fixes #1630

### DIFF
--- a/src/Frontend/Core/Engine/Header.php
+++ b/src/Frontend/Core/Engine/Header.php
@@ -342,7 +342,7 @@ class Header extends FrontendBaseObject
         $image = str_replace(SITE_URL, '', $image);
 
         // check if it no longer points to an absolute uri
-        if (mb_substr($image, 0, 7) != SITE_PROTOCOL . '://') {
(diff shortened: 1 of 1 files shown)


WouterSioen:
Note: Style CI issues are already fixed in https://github.com/forkcms/forkcms/pull/1633

tijsverkoyen:
👍

Issue #1209, opened 2015-05-22
Functional tests screw up the 'production' cache


dubgeiser:
When you run the functional tests, cached templates are overwritten... to reproduce (this is the case where I discovered it at least):

  * Go to the detail of the blog post, url: http://fork.dev/en/blog/detail/nunc-sediam-est
  * Run the test suite
  * reload the url: I got some editor content "Location"

when you run ./tools/remove_cache, all is well again.

This might not be considered critical, but I can see this causing confusing in CI/D systems.

WouterSioen:
We'll fix it this way:

* Cache files that should not be available through a url, will be put in the kernel.cache_dir, which makes them environment dependant.

* For other cache files (like locale and pages), we'll add the environment in the filename and we keep them in the same directory. locale.json will for example become locale.test.json and locale.prod.json.

PR #1636, opened 2016-07-19
Improve fork caching

WouterSioen:
This should fix https://github.com/forkcms/forkcms/issues/1209 mostly.

I only left the folders in there that need to be accessible using a browser (minified assets, locale file, editor link list)


tijsverkoyen:
:shipit:

PR #1638, opened 2016-07-20
Only show the successMessage when the form itself is submitted.

### LINKED ISSUE #1637: Other form shows up when one is submitted
In fork 4.0.5 I have a page which has two forms.
image

But when i submit the top form, I get the fields of the second form and the success message. The second form does not show the success message, it only shows the right form.
image

When I submit the bottom form. It is hidden and show the success message above the top form. Note the string that says "Dit is een test"
image

WouterSioen:
The widgets still share some data, since we still assign all variables
to one template. This makes sure we don't have the success message
anymore if the first formBuilder is submitted and the second isn't.

Fixes #1637


tijsverkoyen:
:shipit:

PR #1639, opened 2016-07-20
Run tests on php 71

WouterSioen:


WouterSioen:
labeled as bug because datagrids didn't render on php 7.1

PR #1642, opened 2016-07-22
Make sure widgets and blocks don't interfere with each other. Fixes #1637

### LINKED ISSUE #1637: Other form shows up when one is submitted

WouterSioen:
We now render them in complete isolation. We assign all core stuff in
global variables so they can be used in modules though.


WouterSioen:
fix verified by @jacob-v-dam on slack

PR #1645, opened 2016-07-25
Improve the visibility of hidden blocks in a page

StijnVrolijk:
I've noticed - or rather, NOT noticed - hidden blocks have very little indication
they are effectively hidden. I decided to add some more visual indication a block is hidden
by lowering the transparency of the hidden block.


carakas:
it looks good to me, but maybe a frontender should check if the approach is correct

StijnVrolijk:
Here you can see the different styles, from top to bottom:
Hidden
Hidden w/ hover
Visible
screen shot 2016-07-25 at 11 30 03
Here you can see the contrast a hidden block gives compared to a visible block
screen shot 2016-07-25 at 11 30 12

WouterSioen:
looks good to me. @mathiashelin or @Katrienvh can you review the sass stuff?

StijnVrolijk:
A clarification as why I chose to use opacity and not just a color:
There are already plenty of shades of gray (insert pun?) and the current color used, `$gray-light`, was too dark for my taste. The next shade, `$gray-lighter`, was too light. I thought it was a bad idea to add yet another shade so I used opacity to achieve a similar effect.

WouterSioen:
I think opacity is also more maintainable, since this will still work as intended when the original color is changed :)

carakas:
wouldn't something like this be better ? http://sass-lang.com/documentation/Sass/Script/Functions.html#lighten-instance_method because opacity also affects the text colour, and I don't know if anysurfer likes that

WouterSioen:
isn't it the text color that should be affected in our case?

StijnVrolijk:
@carakas this is good for people who use SASS, people who use the plain CSS will lose the advantage the opacity provides. I think using the opacity is a nice compromise

mathiashelin:
Looking good

StijnVrolijk:
Can I get a :shipit: ?

mathiashelin:
:shipit:

PR #1646, opened 2016-07-25
Remove amp; in parsePagination

dmoerman:
Remove `amp;` in the parsePagination method, because in the twig template the raw filter is not used. Because of this the paginations did not work with using `page` as `url parameter` when there were multiple parameters in the url


WouterSioen:
looks good to me

Issue #1640, opened 2016-07-20
Diacritics DB vs frontend presentation 


jan-oliva:
For example
When I create blog post and set title **"ěščřžýáíé"**  , Title is stored in database converted to HTML entities. 
On frontend title shown with html entities. When in widget template on frontend use |raw, title shown correctly. 
On backend title is shown correctly.

I think that texts (varchar)  should saved by as original text and  without must use |raw for display each varchar.

WouterSioen:
fix merged in #1649

PR #1649, opened 2016-07-26
Fix form inputs turning & into &amp; and other special chars

carakas:
because html isn't true htmlentities are applied to the input, text like "Ben & Jerry's" needs the raw parameter in twig because of that, this should not happen"


WouterSioen:
This should fix https://github.com/forkcms/forkcms/issues/1640 for new data

PR #1661, opened 2016-08-09
Replace irc with slack as support option in composer.json

### LINKED ISSUE #1659: Add Slack as support channel in composer.json
## Type

- Enhancement

## Problem description

Slack should be added as support channel to composer.json file.

carakas:
## Type

- Enhancement

fixes #1659 

## Pull request description

it replaces irc with slack as support option

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "bin-dir": "bin"
     },
     "support": {
-        "irc": "irc://irc.freenode.org/forkcms",
+        "forum": "https://fork-cms.herokuapp.com",
         "issues": "https://github.com/forkcms/forkcms/issues"
     },
     "scripts": {


composer.json
```diff
@@ -28,7 +28,7 @@
         "bin-dir": "bin"
     },
     "support": {
-        "irc": "irc://irc.freenode.org/forkcms",
+        "slack": "https://fork-cms.herokuapp.com",
```
WouterSioen: I don't think "slack" is a valid key in the "support" part?

Have you done "composer validate" after this change?

carakas: ./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
Defining autoload.psr-0 with an empty namespace prefix is a bad idea for performance

carakas: but it is not among the options listed in the documentation:
    email: Email address for support.
    issues: URL to the issue tracker.
    forum: URL to the forum.
    wiki: URL to the wiki.
    irc: IRC channel for support, as irc://server/channel.
    source: URL to browse or download the sources.
    docs: URL to the documentation.
    rss: URL to the RSS feed.

what would be a better pick in that case ?

PR #1667, opened 2016-08-10
Fixes fork for mysql 5.7

carakas:
## Type

- Critical bugfix

some queries don't work in mysql 5.7^

## Pull request description

In mysql 5.7 the sql_mode ONLY_FULL_GROUP_BY was enabled by default. This didn't work together with some old queries in fork
This should remove that from the session configuration so fork works on mysql 5.7 while we migrate to depricating SpoonDatabase in favour of Doctrine


GhaziTriki:
Thanks @carakas I have tested this update and it works fine for MySQL 5.7

WouterSioen:
nice one 👍

PR #1668, opened 2016-08-10
Pagination labels not showing.

bartdc:
- Non critical bug

Pagination vars were not parsed.

## Pull request description

Pagination vars are displayed correct now.


carakas:
It is indeed a bug, but I don't think it is critical, since it is only in the default triton theme.
Most websites have their own theme.

PR #1671, opened 2016-08-13
Fix incorrect language variable

jessedobbelaere:
## Type

- Non critical bugfix

/

## Pull request description

Go to the demo for example, http://demo.fork-cms.com/, and checkout the body class. It's empty, should have "en" or "nl" or a language. 
Checkout a dutch frontend website, http://demo.fork-cms.com/nl/blog, the dates are in english instead of dutch.
... numerous other examples

*Why this is happening?*
Everywhere in twig we use `LANGUAGE` which doesn't exist and renders nothing at all. We should use `BACKEND_LANGUAGE` and `FRONTEND_LANGUAGE` to make this working again, right?

### DIFF
--- a/src/Frontend/Core/Engine/Ajax.php
+++ b/src/Frontend/Core/Engine/Ajax.php
@@ -195,6 +195,7 @@ class Ajax extends \KernelLoader implements \ApplicationInterface
 
         // define constant
         defined('FRONTEND_LANGUAGE') || define('FRONTEND_LANGUAGE', $this->language);
+        defined('LANGUAGE') || define('LANGUAGE', $this->language);
 
         // set the locale (we need this for the labels)
         Language::setLocale($this->language);
--- a/src/Frontend/Core/Engine/Footer.php
+++ b/src/Frontend/Core/Engine/Footer.php
@@ -70,10 +70,10 @@ class Footer extends FrontendBaseObject
     protected function getFacebookHtml($facebookAppId)
     {
         // build correct locale
-        $locale = mb_strtolower(FRONTEND_LANGUAGE) . '_' . mb_strtoupper(FRONTEND_LANGUAGE);
+        $locale = mb_strtolower(LANGUAGE) . '_' . mb_strtoupper(LANGUAGE);
 
         // reform some locale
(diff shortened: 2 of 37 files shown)


carakas:
I am more in favour of using LANGUAGE everywhere instead of adding the BACKEND and FRONTEND prefix

carakas:
also replacing LANGUAGE with BACKEND_LANGUAGE or FRONTEND_LANGUAGE is not BC

carakas:
is it now a mix of both ?

carakas:
@jessedobbelaere is it now a mix of both ? because I see you removing the LANGUAGE from the backend, so that would not be BC I think

jeroendesloovere:
@jessedobbelaere I prefer LANGUAGE instead of BACKEND_ or FRONTEND_LANGUAGE

jessedobbelaere:
I prefer LANGUAGE too. I just found that LANGUAGE didn't work and FRONTEND_LANGUAGE did, so quickly fixed that. But i'll re-check and try to make LANGUAGE work everywhere

jessedobbelaere:
I added a new const "LANGUAGE" (left the FRONTEND_LANGUAGE const there for backwards compatibility) . Then changed every FRONTEND_LANGUAGE in the frontend to LANGUAGE.

I dunno though, every global constant is prefixed with FRONTEND. If we change language, would that mean we have to change them all to remove FRONTEND_ for consistency sake?
- FRONTEND_CACHE_PATH
- FRONTEND_CACHE_URL
- FRONTEND_CORE_PATH
- FRONTEND_CORE_URL
- FRONTEND_LANGUAGE
- FRONTEND_MODULES_PATH
- FRONTEND_FILES_PATH
- FRONTEND_FILES_URL
- FRONTEND_PATH

Maybe just using `FRONTEND_LANGUAGE` instead of `LANGUAGE` (not working currently) in the current twig templates is an easier solution. 

In the backend however, we use `LANGUAGE` in twig, but thats a twig variable and no global php constant: https://github.com/forkcms/forkcms/blob/master/src/Backend/Core/Engine/TwigTemplate.php#L119

Issue #1672, opened 2016-08-16
Backend Location module, getCoordinates is not functioning


jeroendesloovere:
## Type

- Critical bug

## Problem description

> array(3) { ["error_message"]=> string(74) "Browser API keys cannot have referer restrictions when used with this API." ["results"]=> array(0) { } ["status"]=> string(14) "REQUEST_DENIED" } Redirecting to /private/nl/location/edit?token=allf1mqjwr&id=24&report=added&var=test.

In google you can define two keys, a browser key and a server key, for getting the coordinates we need a server key, but in the frontend we need a browser key...

[Google source](https://developers.google.com/maps/faq#switch-key-type)

## Steps to reproduce

> Try adding a location in the backend location module

## Expected behavior / Proposed solutions

> "&key=x" can not be used in getCoordinates, so remove this

carakas:
@jeroendesloovere will you create a fix for this ?

carakas:
@jeroendesloovere can you recreate a fix for this? The previous MR was closed because it was getting too messy, it would be best to start from a clean branch based on forkcms/master

carakas:
don't forget to take https://github.com/forkcms/forkcms/pull/1673#issuecomment-241993336 into account

WouterSioen:
@jeroendesloovere will you have time for this today? If you haven't, we'll move this to the next bugfix release. (which will probably be 4.2.1)

PR #1673, opened 2016-08-16
Fix backend getcoordinates in location module

jeroendesloovere:


jeroendesloovere:
Fixes #1672

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -138,11 +138,8 @@ public static function getCoordinates(
         // define address
         $address = implode(' ', $item);
 
-        // define Google Maps API key
-        $apikey = BackendModel::get('fork.settings')->get('Core', 'google_maps_key');
-
         // define url
-        $url = 'http://maps.googleapis.com/maps/api/geocode/json?address=' . rawurlencode($address) . '&key=' . $apikey;
+        $url = 'https://maps.googleapis.com/maps/api/geocode/json?address=' . rawurlencode($address);
```
carakas: why did you remove the key?

carakas:
your builds are failing, since this is such a minor change it might be better to close this one and create a new pull request that doesn't have 14 merges in it :)

src/Backend/Modules/Location/Engine/Model.php
carakas: if it is the wrong key, shouldn't we add the correct key

jeroendesloovere: We need two keys if we want to do it right:
- **browser key** to show the google maps correctly
- **server key** to get the geocoordinates (in BackendLocationModel), but this one is optional. That's why I removed it in BackendLocationModel.

Currently we only use the browser key to show the "google maps".

jeroendesloovere:
schermafbeelding 2016-08-18 om 15 35 54

carakas:
@jeroendesloovere 
screen shot 2016-08-21 at 17 05 42
https://travis-ci.org/forkcms/forkcms/builds/152654762

bartdc:
A Server key can also show a map, besides e.g. get coordinates.

So if we use a Server key in the Backend and the same key in the Frontend, a Browser key is unnecessary.

PR #1674, opened 2016-08-17
Add missing Blog image validation

jessedobbelaere:
## Type
- Non critical bugfix

/

## Pull request description
While comparing the add.php and edit.php action of the blog module, I saw that the add.php action uses image validation, but the edit.php doesn't do that. Just copied the bit of add.php to edit.php to also use it there.


carakas:
these checks are included in the getErrors function already for the FormImage so we don't need to define them manually
see https://github.com/forkcms/forkcms/blob/master/src/Backend/Core/Engine/FormImage.php#L77

jessedobbelaere:
@carakas ok but then I'll maybe better remove them from the Add.php in the Blog module?

carakas:
good idea, and then maybe check if it is used in other modules as well?

PR #1681, opened 2016-08-21
Fix having to remove the cache directories after clearing the cache

carakas:
## Type

- Enhancement

Sometimes after clearing cache or preparing for reinstall I still get errors

## Pull request description

I usually remove the folders in the cache directory, after that the problem is fixed

I'm not sure if it has something to do with my setup (vagrant) so I want to know if you also have this problem and if this fixes it for you

### DIFF
--- a/tools/prepare_for_reinstall
+++ b/tools/prepare_for_reinstall
@@ -22,9 +22,7 @@ clear_file_caches() {
     rm -f `find ${SCRIPT_PATH}/../src/Backend/Cache/Locale/ ! -name ".gitignore" -type f ! -path *.svn/* -type f`
     rm -f `find ${SCRIPT_PATH}/../src/Backend/Cache/CompiledTemplates/ ! -name ".gitignore" -type f ! -path *.svn/* -type f`
 
-    rm -rf `find ${SCRIPT_PATH}/../src/Backend/Cache/dev/`
-    rm -rf `find ${SCRIPT_PATH}/../src/Backend/Cache/install/`
-    rm -rf `find ${SCRIPT_PATH}/../src/Backend/Cache/prod/`
+    rm -rf {SCRIPT_PATH}/../app/cache/*
     return 1;
 }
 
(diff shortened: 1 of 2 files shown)


jessedobbelaere:
Yeah won't hurt to add this I think

tools/remove_cache
```diff
@@ -21,5 +21,6 @@ rm -f `find ${SCRIPT_PATH}/../app/cache/ ! -name ".gitignore" -type f ! -path *.
 
 ${SCRIPT_PATH}/../app/console cache:clear
 ${SCRIPT_PATH}/../app/console cache:clear --env=prod
+${SCRIPT_PATH}/../app/console cache:clear --env=install
```
WouterSioen: I would even dare to add `rm -rf ${SCRIPT_PATH}/../app/cache/*`

tools/prepare_for_reinstall
```diff
@@ -21,6 +21,10 @@ clear_file_caches() {
 
     rm -f `find ${SCRIPT_PATH}/../src/Backend/Cache/Locale/ ! -name ".gitignore" -type f ! -path *.svn/* -type f`
     rm -f `find ${SCRIPT_PATH}/../src/Backend/Cache/CompiledTemplates/ ! -name ".gitignore" -type f ! -path *.svn/* -type f`
+
+    rm -rf `find ${SCRIPT_PATH}/../src/Backend/Cache/dev/`
+    rm -rf `find ${SCRIPT_PATH}/../src/Backend/Cache/install/`
+    rm -rf `find ${SCRIPT_PATH}/../src/Backend/Cache/prod/`
```
WouterSioen: Are these new paths? I can't remember having these in `src/Backend/Cache`?

carakas: hmm I think it should be app and that is was a copy mistake

tools/remove_cache
carakas: fixed in 03c25d921812d7f1d54f50a4725f9002d3da301f

tools/prepare_for_reinstall
carakas: fixed in 03c25d921812d7f1d54f50a4725f9002d3da301f

carakas:
fixes #1695

PR #1683, opened 2016-08-21
Remove SpoonDirectory

jessedobbelaere:
## Type

- Enhancement

We need to get rid of SpoonDirectory (and all of Spoon)

## Pull request description

Replace SpoonDirectory with Symfony's filesystem. Profiles module was the only place that still used SpoonDirectory.

### DIFF
--- a/src/Backend/Modules/Profiles/Installer/Installer.php
+++ b/src/Backend/Modules/Profiles/Installer/Installer.php
@@ -10,6 +10,7 @@ namespace Backend\Modules\Profiles\Installer;
  */
 
 use Backend\Core\Installer\ModuleInstaller;
+use Symfony\Component\Filesystem\Filesystem;
 
 /**
  * Installer for the profiles module.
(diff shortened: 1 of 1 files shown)


carakas:
This probably should have happened a long time ago, since Filesystem was introduced in 3.6

src/Backend/Modules/Profiles/Installer/Installer.php
```diff
@@ -38,10 +39,11 @@ public function install()
         $this->setSetting('Profiles', 'send_mail_for_new_profile_to_profile', false);
 
         // add folders
-        \SpoonDirectory::create(PATH_WWW . '/src/Frontend/Files/Profiles/avatars/source/');
-        \SpoonDirectory::create(PATH_WWW . '/src/Frontend/Files/Profiles/avatars/240x240/');
-        \SpoonDirectory::create(PATH_WWW . '/src/Frontend/Files/Profiles/avatars/64x64/');
-        \SpoonDirectory::create(PATH_WWW . '/src/Frontend/Files/Profiles/avatars/32x32/');
+        $fs = new Filesystem();
```
WouterSioen: I'd prefer `filesystem` as variable name over `fs`. Abbreviations are a bad practice.

jessedobbelaere: I agree, but for consistency sake, I changed all $fs vars used in Fork now

WouterSioen:
:shipit:

PR #1685, opened 2016-08-22
Remove obsolete validation

jessedobbelaere:
## Type
- Enhancement

Validation is already done in the FormImage itself in the backend, and in FrontendFormImage in the frontend. 

## Pull request description
Removed obsolete image validation from modules that use FormImage elements
See #1674

### DIFF
--- a/src/Backend/Modules/Blog/Actions/Add.php
+++ b/src/Backend/Modules/Blog/Actions/Add.php
@@ -125,15 +125,6 @@ class Add extends BackendBaseActionAdd
                 $this->frm->getField('category_id')->addError(BL::err('FieldIsRequired'));
             }
 
-            if ($this->imageIsAllowed) {
-                // validate the image
-                if ($this->frm->getField('image')->isFilled()) {
-                    // image extension and mime type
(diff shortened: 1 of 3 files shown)


jessedobbelaere:
Also, why does `Frontend/Core/Engine/Form.php` have multiple classes in one file, bad practice :o

src/Backend/Modules/Blog/Actions/Add.php
```diff
@@ -125,15 +125,6 @@ private function validateForm()
                 $this->frm->getField('category_id')->addError(BL::err('FieldIsRequired'));
             }
 
-            if ($this->imageIsAllowed) {
```
carakas: you removed too much here :)

carakas:
true, but it is being replaced with SymfonyForms so we can leave it I think

src/Backend/Modules/Blog/Actions/Add.php
jessedobbelaere: Hmm don't think so, the if-block underneath "validate the image" can be fully removed and you would end up with a blank if-statement that checks if imageIsAllowed?

carakas: ah, you are right indeed, I saw the isFilled but overlooked that it didn't throw an error when empty

PR #1690, opened 2016-08-24
Show error when file size is too big

carakas:
## Type

- Critical bugfix

If a file was bigger than the allowed upload size it was ignored like it wasn't uploaded.

## Pull request description

This adds a check to see if the file is too large, if that is the case, an error is thrown

### DIFF
--- a/src/Common/Core/Form.php
+++ b/src/Common/Core/Form.php
@@ -236,6 +236,39 @@ class Form extends \SpoonForm
         // create and return a time field
         return parent::addTime($name, $value, $class, $classError);
     }
+
+    /**
+     * @return string|null
+     */
+    public static function getUploadMaxFileSize()
+    {
+        $uploadMaxFileSize = ini_get('upload_max_filesize');
+        if ($uploadMaxFileSize === false) {
+            return null;
+        }
+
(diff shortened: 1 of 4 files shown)


src/Common/Core/Form.php
```diff
@@ -236,6 +236,39 @@ public function addTime($name, $value = null, $class = null, $classError = null)
...
+     */
+    public static function getUploadMaxFileSize()
+    {
+        $uploadMaxFileSize = ini_get('upload_max_filesize');
+        if ($uploadMaxFileSize === false) {
+            return null;
+        }
+
+        // reformat if defined as an integer
+        if (\SpoonFilter::isInteger($uploadMaxFileSize)) {
```
tijsverkoyen: As we try to eliminate Spoon, wouldn't it be more logical to use: http://php.net/manual/en/function.is-numeric.php instead of SpoonFilter::isInteger?

carakas: good point, I'll fix this

carakas: fixed in #5379db8

tijsverkoyen:
👍

WouterSioen:
🎉

Issue #1691, opened 2016-08-25
Remove Push Notifications from Core. Fixes #927


jeroendesloovere:
## Type

- Enhancement

## Problem description

> Push Notifications are currently in the core. But for them to work, you need to integrate this in your iOs / Android / Windows App. The fork cms app is currently not supported anymore. Fixes #927

## Expected behavior / Proposed solutions

> Remove from the core. If users want to integrate Push Notifications, they can use Doctrine Entities + Symfony Events + [Push Notifications Bundle](https://www.google.be/search?client=safari&rls=en&q=push+notifications+bundle&ie=UTF-8&oe=UTF-8&gfe_rd=cr&ei=LZW-V-G-F8rO8gfdrJbIAQ)

PR #1692, opened 2016-08-25
Deprecate push stuff

tijsverkoyen:
## Type

- Enhancement

#1691 

## Pull request description

> This Pull request will mark the methods related to push-notification as deprecated, and it removes the usage of these methods in the core-modules.


src/Backend/Modules/FormBuilder/Resources/config/services.yml
```diff
@@ -6,7 +6,3 @@ services:
             - "@fork.settings"
         tags:
             - { name: kernel.event_listener, event: form.submitted, method: onFormSubmitted }
-    formbuilder.submitted.notifier:
-        class: Frontend\Modules\FormBuilder\EventListener\FormBuilderSubmittedNotificationSubscriber
-        tags:
-            - { name: kernel.event_listener, event: form.submitted, method: onFormSubmitted }
```
WouterSioen: maybe deprecate this FormBuilderSubmittedNotificationSubscriber class too?

src/Frontend/Modules/FormBuilder/EventListener/FormBuilderSubmittedNotificationSubscriber.php
```diff
@@ -7,6 +7,8 @@
 
 /**
  * A Formbuilder submitted event subscriber that will send a notification
+ *
+ * @deprecated: no more support for the Fork-app.
  */
 final class FormBuilderSubmittedNotificationSubscriber
```
carakas: maybe also add the deprecation notice in here

src/Backend/Modules/FormBuilder/Resources/config/services.yml
tijsverkoyen: Was fixed in 0b2f33ba2ef36bc0f0b914a1eba645326520979a

PR #1696, opened 2016-08-26
Restrict access to the vendor directory with htaccess

carakas:
## Type

- Non critical bugfix

At the moment the vendor directory can be accessed from the browser, since we don't always know what is in there it might pose some risks

## Pull request description

It restricts access to the vendor dir via .htaccess


WouterSioen:
We've tested and improved this together so you get a :shipit: from me

carakas:
http://www.reactiongifs.com/homer-skipping/

PR #1251, opened 2015-07-02
Remove the timezones table from the database.

WouterSioen:
It looks like this is never used in Fork. If somebody needs a list of
timezones, the php method timezone_list_identifiers could be used.

https://secure.php.net/manual/en/datetimezone.listidentifiers.php

Fixes #1223

PR #1699, opened 2016-08-26
Fully delete timezones table

GhaziTriki:
## Type

- Enhancement

After insalling ForkCMS `timezones` table is created in the database.

## Pull request description

`timezones` table is now completely removed from core `install.sql`.


carakas:
has it also been deleted from tools/test_db.sql

carakas:
should have been removed with https://github.com/forkcms/forkcms/pull/1251

PR #1718, opened 2016-08-31
Extra type value object

StijnVrolijk:
## Type

- Enhancement

## Pull request description

This is just a helper class for the module extra type. It helps make sure the extra type will be inserted correctly in the database. In the next major release, only the ExtraType value object should be accepted by the `setExtra()` method to ensure correctness.

### DIFF
--- a/src/Backend/Modules/Blog/Installer/Installer.php
+++ b/src/Backend/Modules/Blog/Installer/Installer.php
@@ -10,7 +10,7 @@ namespace Backend\Modules\Blog\Installer;
  */
 
 use Backend\Core\Installer\ModuleInstaller;
-use Common\ExtraType;
+use Common\ModuleExtraType;
 
 /**
  * Installer for the blog module
@@ -145,18 +145,18 @@ class Installer extends ModuleInstaller
         $this->setNavigation($navigationModulesId, 'Blog', 'blog/settings');
 
         // add extra's
-        $blogId = $this->insertExtra('Blog', ExtraType::block(), 'Blog', null, null, 'N', 1000);
-        $this->insertExtra('Blog', ExtraType::widget(), 'RecentComments', 'RecentComments', null, 'N', 1001);
-        $this->insertExtra('Blog', ExtraType::widget(), 'Categories', 'Categories', null, 'N', 1002);
-        $this->insertExtra('Blog', ExtraType::widget(), 'Archive', 'Archive', null, 'N', 1003);
-        $this->insertExtra('Blog', ExtraType::widget(), 'RecentArticlesFull', 'RecentArticlesFull', null, 'N', 1004);
-        $this->insertExtra('Blog', ExtraType::widget(), 'RecentArticlesList', 'RecentArticlesList', null, 'N', 1005);
+        $blogId = $this->insertExtra('Blog', ModuleExtraType::block(), 'Blog', null, null, 'N', 1000);
+        $this->insertExtra('Blog', ModuleExtraType::widget(), 'RecentComments', 'RecentComments', null, 'N', 1001);
+        $this->insertExtra('Blog', ModuleExtraType::widget(), 'Categories', 'Categories', null, 'N', 1002);
+        $this->insertExtra('Blog', ModuleExtraType::widget(), 'Archive', 'Archive', null, 'N', 1003);
+        $this->insertExtra('Blog', ModuleExtraType::widget(), 'RecentArticlesFull', 'RecentArticlesFull', null, 'N', 1004);
+        $this->insertExtra('Blog', ModuleExtraType::widget(), 'RecentArticlesList', 'RecentArticlesList', null, 'N', 1005);
 
         // get search extra id
         $searchId = (int) $this->getDB()->getVar(
             'SELECT id FROM modules_extras
              WHERE module = ? AND type = ? AND action = ?',
-            array('Search', ExtraType::WIDGET, 'Form')
+            array('Search', ModuleExtraType::WIDGET, 'Form')
         );
 
         // loop languages
--- a/src/Backend/Modules/Blog/Installer/Installer.php
+++ b/src/Backend/Modules/Blog/Installer/Installer.php
@@ -156,7 +156,7 @@ class Installer extends ModuleInstaller
         $searchId = (int) $this->getDB()->getVar(
             'SELECT id FROM modules_extras
              WHERE module = ? AND type = ? AND action = ?',
-            array('Search', 'widget', 'Form')
+            array('Search', ExtraType::WIDGET, 'Form')
         );
 
         // loop languages
--- a/src/Backend/Modules/Blog/Installer/Installer.php
+++ b/src/Backend/Modules/Blog/Installer/Installer.php
@@ -10,6 +10,7 @@ namespace Backend\Modules\Blog\Installer;
  */
 
 use Backend\Core\Installer\ModuleInstaller;
+use Common\ExtraType;
 
 /**
  * Installer for the blog module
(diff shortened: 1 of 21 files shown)


jeroendesloovere:
Nice!

Just a small question: why did you use the classname "ExtraType" and not "ModuleExtraType"?

src/Backend/Modules/Blog/Installer/Installer.php
```diff
@@ -144,18 +145,18 @@ public function install()
...
+        $this->insertExtra('Blog', ExtraType::widget(), 'Archive', 'Archive', null, 'N', 1003);
+        $this->insertExtra('Blog', ExtraType::widget(), 'RecentArticlesFull', 'RecentArticlesFull', null, 'N', 1004);
+        $this->insertExtra('Blog', ExtraType::widget(), 'RecentArticlesList', 'RecentArticlesList', null, 'N', 1005);
 
         // get search extra id
         $searchId = (int) $this->getDB()->getVar(
             'SELECT id FROM modules_extras
              WHERE module = ? AND type = ? AND action = ?',
-            array('Search', 'widget', 'Form')
+            array('Search', ExtraType::WIDGET, 'Form')
```
carakas: is there a reason you didn't use ExtraType::widget() here?

src/Backend/Modules/Faq/Installer/Installer.php
```diff
@@ -68,7 +69,7 @@ private function addCategory($language, $title, $url)
             'modules_extras',
             $extra,
             'id = ? AND module = ? AND type = ? AND action = ?',
-            array($item['extra_id'], 'faq', 'widget', 'category_list')
+            array($item['extra_id'], 'faq', ExtraType::WIDGET, 'category_list')
```
carakas: idem

src/Backend/Modules/FormBuilder/Installer/Installer.php
```diff
@@ -61,7 +62,7 @@ public function install()
             'SELECT id
              FROM modules_extras
              WHERE module = ? AND type = ? AND action = ?',
-            array('search', 'widget', 'form')
+            array('search', ExtraType::WIDGET, 'form')
```
carakas: idem

src/Backend/Modules/Mailmotor/Installer/Installer.php
```diff
@@ -184,7 +185,7 @@ private function installPages()
         // get search extra id
         $searchId = (int) $this->getDB()->getVar(
             'SELECT id FROM modules_extras WHERE module = ? AND type = ? AND action = ?',
-            array('Search', 'widget', 'Form')
+            array('Search', ExtraType::WIDGET, 'Form')
```
carakas: idem

src/Backend/Modules/Tags/Installer/Installer.php
```diff
@@ -44,14 +45,14 @@ public function install()
...
-        $this->insertExtra('Tags', 'widget', 'Related', 'Related', null, 'N', 32);
+        $tagsID = $this->insertExtra('Tags', ExtraType::block(), 'Tags', null, null, 'N', 30);
+        $this->insertExtra('Tags', ExtraType::widget(), 'TagCloud', 'TagCloud', null, 'N', 31);
+        $this->insertExtra('Tags', ExtraType::widget(), 'Related', 'Related', null, 'N', 32);
 
         // get search extra id
         $searchId = (int) $this->getDB()->getVar(
             'SELECT id FROM modules_extras WHERE module = ? AND type = ? AND action = ?',
-            array('Search', 'widget', 'Form')
+            array('Search', ExtraType::WIDGET, 'Form')
```
carakas: idem

src/Common/ExtraType.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace Common;
```
carakas: I am not sure about the namespace
since more and more will be moved to things like this I think it is important to use a good structure from the start, since changing it can be non BC

jeroendesloovere: Indeed, Moving everything to Common is the same like creating another spoon library.

StijnVrolijk:
@jeroendesloovere I felt I didn't neet to specify it's an extra type for modules since there is no other 'extra' in Fork but you are right, it could be clearer.

src/Backend/Modules/Blog/Installer/Installer.php
StijnVrolijk: Because the query parameter should be a string. I felt it would be a bit silly to create an object from a string just to extract the string from it again. At first I used `(string) ExtraType::widget()` but it just seemed wrong, even though the `(string)` casting might not even be necessary

src/Backend/Modules/Faq/Installer/Installer.php
StijnVrolijk: ditto

src/Backend/Modules/FormBuilder/Installer/Installer.php
StijnVrolijk: ditto

src/Backend/Modules/Mailmotor/Installer/Installer.php
StijnVrolijk: ditto

src/Backend/Modules/Tags/Installer/Installer.php
StijnVrolijk: ditto

src/Common/ExtraType.php
StijnVrolijk: What do you suggest? Best I could come up with is `Common\ValueObject` but I deferred from that seeing as there are similar classes to this one in the `Common` namespace.

src/Backend/Modules/Blog/Installer/Installer.php
carakas: ok 👍

src/Common/ExtraType.php
carakas: true

carakas:
I think that if you rename it to ModuleExtraType this can be merged

StijnVrolijk:
And if I fix that weird test failure ;)

src/Backend/Modules/Blog/Installer/Installer.php
```diff
@@ -144,18 +145,18 @@ public function install()
...
-        $this->insertExtra('Blog', 'widget', 'Categories', 'Categories', null, 'N', 1002);
-        $this->insertExtra('Blog', 'widget', 'Archive', 'Archive', null, 'N', 1003);
-        $this->insertExtra('Blog', 'widget', 'RecentArticlesFull', 'RecentArticlesFull', null, 'N', 1004);
-        $this->insertExtra('Blog', 'widget', 'RecentArticlesList', 'RecentArticlesList', null, 'N', 1005);
+        $blogId = $this->insertExtra('Blog', ExtraType::block(), 'Blog', null, null, 'N', 1000);
+        $this->insertExtra('Blog', ExtraType::widget(), 'RecentComments', 'RecentComments', null, 'N', 1001);
+        $this->insertExtra('Blog', ExtraType::widget(), 'Categories', 'Categories', null, 'N', 1002);
+        $this->insertExtra('Blog', ExtraType::widget(), 'Archive', 'Archive', null, 'N', 1003);
+        $this->insertExtra('Blog', ExtraType::widget(), 'RecentArticlesFull', 'RecentArticlesFull', null, 'N', 1004);
+        $this->insertExtra('Blog', ExtraType::widget(), 'RecentArticlesList', 'RecentArticlesList', null, 'N', 1005);
```
jeroendesloovere: Is there a reason that you here used `ExtraType::widget()` instead of `ExtraType::WIDGET`?

carakas: because this is a function, by doing it like this you can when everything is running on this say that the parameter needs to be of the type ExtraType

StijnVrolijk: That is correct. For now, the value object `ExtraType` with type `ExtraType::WIDGET` just get's casted back to `ExtraType::WIDGET` but in the future we can assert that an `ExtraType` value object should be sent instead of a string. Doing this will make sure we have a valid extra type in our database

StijnVrolijk:
Let's see if travis lets me have this one

carakas:
I reran the previous build and it didn't fail the second time, so it should be fine

StijnVrolijk:
Huzzah!

PR #1719, opened 2016-09-01
ContentBlocks: Fixes template error when using {{ parsewidget('ContentBlocks', 'Detail', 1) }}

jeroendesloovere:
**Problem description**

When using `{{ parsewidget('ContentBlocks', 'Detail', 1) }}` in a template, the content block template was not found. Since it was not defined in the module.

The error was: Template "" is not defined in
Core/Layout/Templates/Home.html.twig at line 59
schermafbeelding 2016-09-01 om 10 16 06

### DIFF
--- a/src/Frontend/Modules/ContentBlocks/Widgets/Detail.php
+++ b/src/Frontend/Modules/ContentBlocks/Widgets/Detail.php
@@ -23,11 +23,19 @@ class Detail extends FrontendBaseWidget
             Locale::frontendLanguage()
         );
 
+        // Init template
+        $template = "Default.html.twig";
+
         // if the content block is not found or if it is hidden, just return an array with empty text
         // @deprecated fix this for version 5, we just shouldn't assign this instead of this hack, but we need it for BC
         if (!$contentBlock instanceof ContentBlock || $contentBlock->isHidden()) {
             $contentBlock = ['text' => ''];
+        } else {
+            $template = $contentBlock->getTemplate();
         }
+
+        // Load template
+        $this->loadTemplate('ContentBlocks/Layout/Widgets/' . $template);
 
         $this->tpl->assign('widgetContentBlocks', $contentBlock);
         // That's all folks!


src/Frontend/Modules/ContentBlocks/Widgets/Detail.php
```diff
@@ -23,12 +23,20 @@ public function execute()
             Locale::frontendLanguage()
         );
 
+        // Init template
+        $template = "Default.html.twig";
```
carakas: you don't need this, this should always be available

src/Frontend/Modules/ContentBlocks/Widgets/Detail.php
```diff
@@ -23,12 +23,20 @@ public function execute()
...
+
         // if the content block is not found or if it is hidden, just return an array with empty text
         // @deprecated fix this for version 5, we just shouldn't assign this instead of this hack, but we need it for BC
         if (!$contentBlock instanceof ContentBlock || $contentBlock->isHidden()) {
             $contentBlock = ['text' => ''];
+        } else {
```
carakas: you also don't need this

src/Frontend/Modules/ContentBlocks/Widgets/Detail.php
```diff
@@ -23,12 +23,20 @@ public function execute()
...
+            $template = $contentBlock->getTemplate();
         }
 
+        // Load template
+        $this->loadTemplate('ContentBlocks/Layout/Widgets/' . $template);
```
carakas: you can just add $contentBlock->getTemplate() here

jeroendesloovere: No, that's not true, since if you give a wrong "ID" you don't have a ContentBlock Entity.

carakas: this should also be wrapped in `if ($this->get('parseWidget', true)) {`

carakas: if you give a wrong id the thing should not work :)

carakas:
also, this is a quick fix, you should investegate why it works when parsing in the page but not when running parse widget

carakas:
This was a quickfix, the real issue is fixed in https://github.com/forkcms/forkcms/pull/1720

PR #1720, opened 2016-09-01
Make sure the real data is loaded when using the template modifier parsewidget

carakas:
## Type

- Non critical bugfix

The data was never loaded, it just added the id if given
This caused widgets that need other data to crash
I've added a static constructor that will try to get the real data from the table modules_extras


jeroendesloovere:
The only difference is that my PR-fix was using $moduleExtraId and yours $moduleExtraDataId.

carakas:
that wouldn't be backwards compatible

PR #1723, opened 2016-09-01
Make sure that extra things added to the data of a content block aren't lost when updating

### LINKED ISSUE #1713: ContentBlocks Entity ->updateWidget() updates all data values
## Type

- Enhancement

## Problem description

The updateWidget() updates the data array completely, which means that custom fields can't be used. Proposal to only update "extra_label" which sets the title of the ContentBlock.

[Code](https://github.com/forkcms/forkcms/blob/master/src/Backend/Modules/ContentBlocks/Entity/ContentBlock.php#L304)

Only update "data" "extra_label" in `src/Backend/Modules/ContentBlocks/Entity/ContentBlock.php`

carakas:
## Type

- Non critical bugfix

fixes #1713 

## Pull request description

Make sure that extra things added to the data of a content block aren't lost when updating

We now get the current data and only overwrite the things that we need

### DIFF
--- a/src/Backend/Modules/ContentBlocks/Entity/ContentBlock.php
+++ b/src/Backend/Modules/ContentBlocks/Entity/ContentBlock.php
@@ -299,17 +299,20 @@ class ContentBlock
 
         // update data for the extra
         // @TODO replace this with an implementation with doctrine
-        Model::updateExtra(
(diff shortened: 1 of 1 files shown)


StijnVrolijk:
:shipit:

PR #1730, opened 2016-09-05
Update spoon library to fix is price check on spoon form text field

carakas:
## Type

- Non critical bugfix

## Pull request description
the check of spoon form isPrice was not using the parameters correctly


carakas:
and apparently a use statement

PR #1741, opened 2016-09-20
Form fields classes are applied

### LINKED ISSUE #1734: FormBuilder: classname property not applied for all fields
## Type

- Non critical bug

## Problem description

Setting CSS class name property for any form field is not applied at the frontend.

## Steps to reproduce

- Create a form.
- Add a new field.
- Set a custom CSS class name.
- Load the form in the frontend.

The style should be applied to form field.

GhaziTriki:
## Type

- Non critical bugfix

Fixes #1734


GhaziTriki:
@StijnVrolijk @carakas thanks

PR #1742, opened 2016-09-20
Make sure the column name check is case insensitive

### LINKED ISSUE #1737: Backend datagrid "Edit" button layout
## Type

- Non critical bugfix

## Problem description

In the backend, when typing `$this->dgRecent->addColumn('Edit'` this is not supported.
You need to type `$this->dgRecent->addColumn('edit'` to get the button layout right.

add/Add/edit/Edit/... should all work the same

P.S.: in older fork cms 3.9.* this worked just fine.

StijnVrolijk:
## Type

- Non critical bugfix

Fixes #1737

### DIFF
--- a/src/Backend/Core/Engine/DataGrid.php
+++ b/src/Backend/Core/Engine/DataGrid.php
@@ -623,13 +623,12 @@ class DataGrid extends \SpoonDataGrid
      */
     private function decideIcon($name)
     {
-        $icon = null;
         $name = mb_strtolower($name);
 
-        if (isset($this->mapIcons[$name])) {
-            $icon = $this->mapIcons[$name];
+        if (!isset($this->mapIcons[$name])) {
+            return null;
         }
 
-        return $icon;
+        return $this->mapIcons[$name];
     }
 }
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Engine/DataGrid.php
```diff
@@ -117,7 +117,7 @@ public function addColumn(
 
         // known actions that should have a button
         if (in_array(
-            $name,
+            strtolower($name),
```
carakas: you should use `mb_strtolower`

StijnVrolijk: well spotted, sir!

Ive also save the lowercased version for further use in the code

src/Backend/Core/Engine/DataGrid.php
```diff
@@ -113,11 +113,13 @@ public function addColumn(
         $image = null,
         $sequence = null
     ) {
+        // make sure we use a lowercased column everywhere (useful for checks, icons and more)
+        $name = mb_strtolower($name);
```
carakas: You shouldn't do this as it can break css, we only need it lowercased for the icon check

PR #1744, opened 2016-09-20
Implemented form submit detection for backend form edit action.

### LINKED ISSUE #1738: FormBuilder: hitting enter when editing a form field validates the main form
## Type

- Non critical bug

## Problem description

When editing a from field form then typing "Enter" in keybaord, the page is main form is validated instead of the one in prompt.

## Steps to reproduce

- Create a form.
- Create a form field.
- Hit "Enter" after editing a form field without clicking on "Save".

The form field should be validated instead of the main form. The page should not be reloaded.

GhaziTriki:
## Type

- Non critical bugfix

Fixes #1738

## Pull request description

Added custom form submit handler. If a form field modal window is open, we trigger the click on save button event instead of submitting the main form.

### DIFF
--- a/src/Backend/Modules/FormBuilder/Js/FormBuilder.js
+++ b/src/Backend/Modules/FormBuilder/Js/FormBuilder.js
@@ -79,12 +79,34 @@ jsBackend.FormBuilder.Fields =
             jsBackend.FormBuilder.Fields.defaultErrorMessages = defaultErrorMessages;
         }
 
+        // submit detection handler for the main form and modal field form
+        jsBackend.FormBuilder.Fields.bindFromSubmit();
+
         // bind
         jsBackend.FormBuilder.Fields.bindDialogs();
         jsBackend.FormBuilder.Fields.bindValidation();
         jsBackend.FormBuilder.Fields.bindEdit();
         jsBackend.FormBuilder.Fields.bindDelete();
         jsBackend.FormBuilder.Fields.bindDragAndDrop();
+    },
+
+    /**
+     * Bind the form submit action
+     */
+    bindFromSubmit: function () {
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/FormBuilder/Js/FormBuilder.js
```diff
@@ -88,6 +91,25 @@ jsBackend.FormBuilder.Fields =
...
     /**
+     * Bind the form submit action
+     */
+    bindFromSubmit: function () {
+        $("#edit").submit(function (e) {
+            // check if a modal window is already open
+            $('.jsFieldDialog').each(function () {
+                // if a modal window is open we prevent the event from propagating
+                if ($(this).css('display') != 'none') {
+                    e.preventDefault();
```
StijnVrolijk: `e.preventDefault()` is unnecessary here since `return false` effectively equals `preventDefault` and `stopPropagation`

src/Backend/Modules/FormBuilder/Js/FormBuilder.js
```diff
@@ -88,6 +91,25 @@ jsBackend.FormBuilder.Fields =
...
+                    $(this).find('.jsFieldDialogSubmit').trigger('click');
+                    return false;
+                }
+            });
+            // continue processing with the default browser submit handler
+            return true;
```
StijnVrolijk: I think this is also unnecessary since it doesn't really do anything?

src/Backend/Modules/FormBuilder/Js/FormBuilder.js
GhaziTriki: removed

PR #1747, opened 2016-09-21
API Client fixed

Diewy:
## Type

- Non critical bugfix
- Enhancement

> The API Client didn't use Twig yet
> The API Client gave a 500 error
> The API Client couldn't call any API methods

## Pull request description

> The API Client was looking for a template called "index.tpl" which is renamed to "Index.html.twig". This PR updates the template to a twig syntax and adds the most basic Twig implementation for the API Client so it can render the template.

> The javascript to pass the parameters used deprecated methods which stopped it from calling any API methods.


src/Api/V1/Client/Js/client.js
```diff
@@ -97,14 +97,14 @@ jsClient.parameters =
...
-}
+};
 
 jsClient.ajax =
 {
 	init: function()
 	{
 		// a call was initialized
-		$('.submit').live('click', jsClient.ajax.submitFormHandler);
+		$('.submit').click(jsClient.ajax.submitFormHandler);
```
StijnVrolijk: Wouldn't `.on('click', jsClient.ajax.submitFormHandler);` be preferable here?  
It has the same functionality`live` has as opposed to just `click`. I don't know the context of how this is used but I think it'd be wise to keep the functionality the same and just update the deprecated stuff.  

See http://api.jquery.com/live/ for more information

src/Api/V1/Engine/Client.php
```diff
@@ -11,18 +11,30 @@
...
+use Twig_Loader_Filesystem;
 
 /**
  * Client for the Fork CMS API.
  */
 class Client extends Api
 {
     /**
-     * @var \SpoonTemplate
+     * @var TwigTemplate
```
StijnVrolijk: I believe this requires a use statement

PR #1755, opened 2016-09-29
Update the schema instead of ignoring when the schema for the table already exists

carakas:
## Type

- Enhancement

## Pull request description

instead of silently failing we should update the schema


jeroendesloovere:
This is a nice enhancement, but also could be dangerous in same occasions...

carakas:
not doing something might also give troubles as it wouldn't do what people expect and get confused.

Also if you get data loss because you tried to create the table for that entity you have a programming error IMHO

PR #1759, opened 2016-10-03
Fix incorrectly resolving css file into theme

jessedobbelaere:
## Type
- Non critical bugfix

/

## Pull request description
Occurred to me while doing development for a project. I need to add `https://api.mapbox.com/mapbox-gl-js/v0.25.1/mapbox-gl.css` to the frontend module, but I could not add it with `addCss`. It would resolve the file path to the local theme. The `addJs` method however has these 3 lines that checks for 'http' in the string. The `addCss` method however, did not have these lines. So that fixes the issue 👍

### DIFF
--- a/src/Frontend/Core/Engine/Header.php
+++ b/src/Frontend/Core/Engine/Header.php
@@ -132,7 +132,11 @@ class Header extends FrontendBaseObject
         $file = (string) $file;
         $minify = (bool) $minify;
         $addTimestamp = (bool) $addTimestamp;
-        $file = Theme::getPath($file);
+
+        // get file path
+        if (mb_substr($file, 0, 4) != 'http') {
+            $file = Theme::getPath($file);
+        }
 
         // no minifying when debugging
         if ($this->getContainer()->getParameter('kernel.debug')) {


src/Frontend/Core/Engine/Header.php
```diff
@@ -132,7 +132,11 @@ public function addCSS($file, $minify = true, $addTimestamp = null)
         $file = (string) $file;
         $minify = (bool) $minify;
         $addTimestamp = (bool) $addTimestamp;
-        $file = Theme::getPath($file);
+
+        // get file path
+        if (mb_substr($file, 0, 4) != 'http') {
```
carakas: wouldn't is be better to use `if (mb_strpos($file, 'http') !== 0) {`

StijnVrolijk: I ran extensive tests and found there is no performance difference between the two methods. Also, looking at the semantics, there is no major difference, it's a potato, potato story. Do you have any arguments as to why yours is better, @carakas ?

carakas: I find it better matching with what you try to test
because if you read the code it is check if the string starts with http
so I think strpos is better suited for that

carakas: But I don't consider this to be a blocking issue if @jessedobbelaere wouldn't have time to change it now :)

StijnVrolijk: I disagree with your argument. To me, `strpos` is 'is the http string present at the start of the string' while `substr` says 'are the 4 first letters of the string http'? Personally I find the latter a tad more logical and since @jessedobbelaere already implemented it like this, I'd just like to merge it :)

carakas: ok

PR #1768, opened 2016-10-09
Implement the symfony translator interface on the language classes

carakas:
## Type

- Feature

## Pull request description

You can now use the trans function of the translator service to translate strings with our system

Instead of `Language::lbl('MyLabel')` you can now do `$this->get'translator')->trans('lbl.MyLabel')`

If we can't translate it it will fallback to the default translation of symfony

If symfony also can't translate it we will return our normal fallback?-.

This was needed to be able to translate the error messages from symfony forms automatically but also makes it easier to move away from using static classes in our code to use the translations.

it also fixes the bug that the error messages when there was missing data for content blocks wasn't translated, see 935ffa1

### DIFF
--- a/src/Common/Language.php
+++ b/src/Common/Language.php
@@ -114,18 +114,24 @@ final class Language extends IdentityTranslator
      */
     public function trans($id, array $parameters = [], $domain = null, $locale = null)
     {
-        if (strpos($id, '.') === false) {
-            parent::trans($id, $parameters, $domain, $locale);
+        // we are not in a normal fork context
+        if (!defined('APPLICATION')) {
+            return parent::trans($id, $parameters, $domain, $locale);
         }
-        list($action, $string) = explode('.', $id, 2);
 
         $possibleActions = ['lbl', 'err', 'msg'];
         if (self::get() === FrontendLanguage::class) {
             $possibleActions[] = 'act';
         }
 
+        if (!preg_match('/(' . implode('|', $possibleActions) . ')./', $id)) {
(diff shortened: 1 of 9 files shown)


src/Common/Language.php
```diff
@@ -97,4 +112,45 @@ public static function msg($key)
...
         return self::callLanguageFunction('msg', [$key]);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function trans($id, array $parameters = [], $domain = null, $locale = null)
+    {
+        // we are not in a normal fork context
+        if (!defined('APPLICATION')) {
```
StijnVrolijk: This one can be removed since APPLICATION isn't used anymore

carakas: Fixed in 5416e7f

src/Common/Language.php
```diff
@@ -25,8 +25,8 @@ public static function get()
         $application = 'Backend';
 
         if (Model::has('request')
-            && Model::get('request')->attributes->has('application')
-            && Model::get('request')->attributes->get('application') === 'Frontend'
+            && Model::get('request')->attributes->has('_route')
+            && strpos(Model::get('request')->attributes->get('_route'), 'frontend') === 0
```
StijnVrolijk: are you 100% sure 'frontend' is always lowercased?

carakas: but to be 100% sure for future changes I made the check case-insensitive 9e0d919

PR #1771, opened 2016-10-10
Fix add link on the blog index page when filtering on a category

### LINKED ISSUE #1770: macro.buttonIcon bug with variables
## Type

- Non critical bug

# Problem description

In src/Backend/Modules/Blog/Actions/Index.php the following doesn't work
```twig
          {{ macro.buttonIcon( geturl('add', null, '&category=#{filterCategory.id}'), 'plus-square', 'lbl.Add'|trans|ucfirst) }}
```
#{filterCategory.id} is not working. It just doesn't show the id.
But if you do {{filterCategory.id}} outside the twig extension, it does work perfectly.

carakas:
## Type

- Non critical bugfix

fixes #1770 

## Pull request description

The id of the category was not passed correctly to the add action

### DIFF
--- a/src/Backend/Modules/Blog/Layout/Templates/Index.html.twig
+++ b/src/Backend/Modules/Blog/Layout/Templates/Index.html.twig
@@ -5,7 +5,7 @@
     <div class="btn-group" role="group">
       {% if showBlogAdd %}
         {% if filterCategory %}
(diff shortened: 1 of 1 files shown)


jeroendesloovere:
That works!

PR #1778, opened 2016-10-11
Fix timezone issues with doctrine

### LINKED ISSUE #1466: Fork CMS + Doctrine + \DateTime
### Example

I want to save $publishOn to "11/05/2016 20:00", this will be saved in the database as "2016-05-11 18:00:00" (Which is correct. Note the two hour difference, because of UTC+1 (Europe/Brussels) + summer time)

If you then want to show "H:i" in your Backend Edit.php action:
```php
date('H:i', $blogArticle->getPublishOn()->getTimestamp());
```
> This shows "18:00" in the website. Which is false!!! Because It should show "20:00"...!!!

### Actual problem

SpoonDate + SpoonForm

### Problems

**1) Doctrine always uses default timezone**
Doctrine always uses the default timezone that is in your setup ( "Europe/Brussels" in my case).

When saving
* A \DateTime, the timezone is NOT saved in the database.

So when loading:
* The \DateTime from the database, your default timezone ("Europe/Brussels") will be used. That is not what we want, it has to be "UTC".

**2) Fork CMS is actually not very good for international \DateTime**

Since we actually set ```date_default_timezone_set('Europe/Brussels');``` in the [`/src/Common/Core/Init.php`](https://github.com/forkcms/forkcms/blob/master/src/Common/Core/Init.php#L47).

### Solution recommended by Doctrine

> Always saving the timezone in your entity. [Read more](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/cookbook/working-with-datetime.html)

```php
<?php
namespace Shipping;

/**
 * @Entity
 */
class Event
{
    /** @Column(type="datetime") */
    private $created;

    /** @Column(type="string") */
    private $timezone;

    /**
     * @var bool
     */
    private $localized = false;

    public function __construct(\DateTime $createDate)
    {
        $this->localized = true;
        $this->created = $createDate;
        $this->timezone = $createDate->getTimeZone()->getName();
    }

    public function getCreated()
    {
        if (!$this->localized) {
            $this->created->setTimeZone(new \DateTimeZone($this->timezone));
        }
        return $this->created;
    }
}
```

### Another working solution

1) At [`/src/Common/Core/Init.php`](https://github.com/forkcms/forkcms/blob/master/src/Common/Core/Init.php#L47) setting `date_default_timezone_set('UTC')` without an if statement

2) Then you need to set the timezone in the backend Edit action:
```php
$this->frm->addTime(
    'publish_on_time',
    date(
        'H:i',
        $this->blogArticle->getPublishOn()->setTimeZone(
            new \DateTimeZone('Europe/Brussels')
        )->getTimestamp()
    )
);
```

### Other solutions?

* When defining Doctrine, maby we can set it to use timezone "UTC" instead of letting Doctrine get the default timezone.

carakas:
## Type

- Non critical bugfix

fixes #1466 

## Pull request description

we overwrite the default datetime and datetimez dbal types of doctrine so everything is stored in UTC like we are used to in fork

### DIFF
--- a/app/config/doctrine.yml
+++ b/app/config/doctrine.yml
@@ -20,6 +20,8 @@ doctrine:
         types:
             enum_bool: Common\Doctrine\Type\EnumBoolType
             locale: Common\Doctrine\Type\LocaleType
+            datetime: Common\Doctrine\Type\UTCDateTimeType
+            datetimetz: Common\Doctrine\Type\UTCDateTimeType
 
     orm:
         auto_generate_proxy_classes: "%kernel.debug%"
--- a/app/config/doctrine.yml
+++ b/app/config/doctrine.yml
@@ -22,6 +22,7 @@ doctrine:
             locale: Common\Doctrine\Type\LocaleType
             datetime: Common\Doctrine\Type\UTCDateTimeType
             datetimetz: Common\Doctrine\Type\UTCDateTimeType
+            time: Common\Doctrine\Type\UTCDateTimeType
 
     orm:
         auto_generate_proxy_classes: "%kernel.debug%"
--- a/src/Common/Doctrine/Type/UTCDateTimeType.php
+++ b/src/Common/Doctrine/Type/UTCDateTimeType.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Common\Doctrine\Type;
+
+use DateTime;
+use DateTimeZone;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\ConversionException;
+use Doctrine\DBAL\Types\DateTimeType;
+
+class UTCDateTimeType extends DateTimeType
+{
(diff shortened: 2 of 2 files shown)


app/config/doctrine.yml
```diff
@@ -20,6 +20,8 @@ doctrine:
         types:
             enum_bool: Common\Doctrine\Type\EnumBoolType
             locale: Common\Doctrine\Type\LocaleType
+            datetime: Common\Doctrine\Type\UTCDateTimeType
+            datetimetz: Common\Doctrine\Type\UTCDateTimeType
```
jeroendesloovere: You are overwriting "datetime" and "datimetz", but why not "time"?
http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/known-vendor-issues.html

carakas: nice catch :) fixed in 42215d9

PR #1779, opened 2016-10-11
Remove server root from loading dirs

jonasdekeukelaere:
## Type
- Critical bugfix
## Pull request description

Fix the error you get when open_basedir is active on a server.

`Warning: file_exists(): open_basedir restriction in effect. File(/) is not within the allowed path(s)`

### DIFF
--- a/src/Frontend/Core/Engine/TwigTemplate.php
+++ b/src/Frontend/Core/Engine/TwigTemplate.php
@@ -147,7 +147,6 @@ class TwigTemplate extends BaseTwigTemplate
                 $this->themePath,
                 FRONTEND_MODULES_PATH,
                 FRONTEND_PATH,
-                '/',
             ),
             function ($folder) use ($filesystem) {
                 return $filesystem->exists($folder);


carakas:
@Thijzer wouter told me you had added that slash because twig sometimes got absolute paths
do you know if this was in the front or the backend?

@jonasdekeukelaere can you run through fork completely to see if something is broken that the tests didn't pick up?

jonasdekeukelaere:
I had to put a project on FTP earlier today and it didn't work, @tijsverkoyen said to delete it and push it to fork.

Test what exactly?

carakas:
to see if twig can load all templates

jonasdekeukelaere:
Looks fine at first sight. Not sure why template wouldn't work without access to the server root.

tijsverkoyen:
I don't see any valid reason to "load" templates from the "/"-folder of the server. / is the start of the server, so in 99.99999999% of the cases it is not even in your application-folder.

jonasdekeukelaere:
Hmm, it isn't as simple as thought. When you want a frontend template in the backend you need to use FRONTEND_MODULES_PATH, but this fails on a server with open_basedir on.

jonasdekeukelaere:
FRONTEND_MODULES_PATH you can just remove if you send from the frontend, in the backend again, you'll need it.

carakas:
then there is the question if you need to be able to use frontend templates in the backend and vice versa

carakas:
@jonasdekeukelaere we would like to release this week, do you have an idea for a fix for this?

jonasdekeukelaere:
No idea, and no time to figure it out at the moment

jonasdekeukelaere:
Note to self: 2 issues to be fixed:
- Loading templates from **Core** (Backend or Frontend) without BACKEND_CODE_PATH or FRONTEND_CORE_PATH
- Loading Frontend (Core) template from Backend

carakas:
@jonasdekeukelaere can't you just add the paths of the frontend to the backend as possibilities after those of the backend and vice versa?

jonasdekeukelaere:
I don't get what you mean.
But the problem is that if you assign a frontend template to the mailer, the only way of the template to know it is a frontend path, is to compare it to FRONTEND_CORE_PATH, which we want gone, as it is a absolute path.

carakas:
I tried using frontend templates in the backend without the slash and didn't have troubles, so I'll just merge this.
If I did miss something they will create an issue and then we have a usecase

PR #1782, opened 2016-10-15
Fixes format template modifiers

### LINKED ISSUE #1781: Backend formatdate and formatdatetime throwing error
## Type
- Non critical bug
## Problem description

> Backend formatdate and formatdatetime throwing error
## Steps to reproduce

> When using `{{ item.createdOn|formatdate }}` or `{{ item.createdOn|formatdatetime }}` an error is thrown.

```
Attempted to call an undefined method named "formatDate" of class "Backend\Core\Engine\TemplateModifiers".
Did you mean to call "formatTime"?
500 Internal Server Error - UndefinedMethodException
```

`{{ item.createdOn|formattime }}` is working!

> That it works

jeroendesloovere:
Fixes #1781


src/Backend/Cache/Analytics/Index_1426692958178.txt
```diff
@@ -0,0 +1 @@
+done
```
carakas: This shouldn't be here

src/Backend/Cache/Analytics/1425942000_1426633200.xml
```diff
@@ -0,0 +1,238 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
carakas: this shouldn't be here

composer.lock
```diff
@@ -1170,6 +1170,90 @@
             "time": "2015-12-19 20:16:43"
         },
         {
+            "name": "league/flysystem",
```
carakas: wasn't this already added?

src/Backend/Core/Engine/TemplateModifiers.php
```diff
@@ -92,6 +92,40 @@ public static function formatNumber($number, $decimals = null)
...
+     *
+     * @return string
+     */
+    public static function formatDateTime($var)
+    {
+        // get setting
+        $format = Authentication::getUser()->getSetting('datetime_format');
+
+        // format the date
+        return \SpoonDate::getDate($format, (int) $var, BackendLanguage::getInterfaceLanguage());
```
carakas: maybe add a use statement for this

composer.lock
carakas: you actually messed up composer install
Your requirements could not be resolved to an installable set of packages.

carakas:
Are they also available in the frontend? because they are also loaded there

src/Frontend/Core/Engine/TemplateModifiers.php
```diff
@@ -11,7 +11,6 @@
 
 use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Core\Engine\Block\Widget as FrontendBlockWidget;
-use Frontend\Core\Language\Language;
```
carakas: Wrong implementation, the other one will throw deprecation warnings so you should create an alias for it and use that alias

carakas:
@jeroendesloovere can you look at the feedback and the failed ci build

src/Frontend/Core/Engine/TemplateModifiers.php
jeroendesloovere: There is no Language left in this class, so I removed this line...

jeroendesloovere: I'm using `LANGUAGE`

carakas: you should use `Frontend\Core\Language\Locale`

jeroendesloovere: Done.

PR #1784, opened 2016-10-16
Meta entity

carakas:
## Type

- Feature

## Pull request description

This PR is based on #1768 since it was needed to get the error messages working

This adds the option to use meta as an entity and in symfony form

additionally this also
- made the jquery function doMeta configurable
- made the locale serializable
- added a method to compare to locale more easily
- made it possible to set the meta settings of an action by passing the meta entity


app/config/doctrine.yml
```diff
@@ -37,3 +44,9 @@ services:
         class: Common\Doctrine\Entity\CreateSchema
         arguments:
             - "@doctrine.orm.entity_manager"
+
+    repository.meta:
```
jeroendesloovere: Just thinking out loud here:
* 'common.repository.meta', 'common.repository.module_extra', ...?
* Or 'fork.repository.meta', 'fork.repository.module_extra' (to be consistent with 'fork.settings', ...)

src/Backend/Core/Engine/Meta.php
```diff
@@ -625,4 +570,14 @@ public function getData()
...
     }
+
+    /**
+     * @return MetaEntity
+     */
+    public function getMetaEntity()
+    {
+        $this->validate();
+
+        return MetaEntity::fromBackendMeta($this);
```
jeroendesloovere: Wouldn't it be better to create a factory for this?
$this->commonMetaFactory->createFromBackend($this);
Which then uses ->create() of the MetaEntity

carakas: For this time I disagree since this class is now deprecated, so it is just a temporary thing to make your life more easy if you still use SpoonForm but want to use the entity

app/config/doctrine.yml
tijsverkoyen: I agree, should use fork as "namespace"

src/Backend/Core/Ajax/GenerateUrl.php
```diff
@@ -19,42 +19,24 @@
...
 
-        // meta object
-        $this->meta = new BackendMeta($frm, $metaId, $baseFieldName, $custom);
-
-        // set callback for generating an unique URL
-        $this->meta->setURLCallback($className, $methodName, $parameters);
-
         // fetch generated meta url
-        $URL = urldecode($this->meta->generateURL($URL));
+        $URL = urldecode($this->get('repository.meta')->generateURL($URL, $className, $methodName, $parameters));
```
tijsverkoyen: Why is `urldecode` needed?

tijsverkoyen: `$URL` => `$url`

src/Backend/Core/Engine/Meta.php
```diff
@@ -507,9 +452,9 @@ public function save($update = false)
      * REMARK: this method has to be public and static
      * REMARK: if you specify arguments they will be appended
      *
-     * @param string $className  Name of the class to use.
+     * @param string $className Name of the class to use.
      * @param string $methodName Name of the method to use.
-     * @param array  $parameters Parameters to parse, they will be passed after ours.
+     * @param array $parameters Parameters to parse, they will be passed after ours.
      */
     public function setURLCallback($className, $methodName, $parameters = array())
```
tijsverkoyen: Why not a real callable?

src/Common/Doctrine/Repository/MetaRepository.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param string $URL The base-url to start from.
+     * @param string $class The Fully Qualified Class Name or service name
+     * @param string $method The method that needs to be called
+     * @param array $parameters The parameters for the callback
+     *
+     * @throws Exception When the function does not exist
+     *
+     * @return string
+     */
+    public function generateURL($URL, $class, $method, array $parameters = [])
```
tijsverkoyen: Ik zou hier een echte callback gebruiken

app/config/doctrine.yml
carakas: fixed in 994036e and aa2a6ac

src/Backend/Core/Ajax/GenerateUrl.php
carakas: I don't know, but it was in the original code so because I didn't know the impact I left it there

src/Backend/Core/Engine/Meta.php
carakas: this is the original code so we can't just change it here because that would not be BC

src/Common/Doctrine/Repository/MetaRepository.php
carakas: I also would prefer that but internally there is a check if the class is the name of a service, that wouldn't work then any more
do you have a suggestion how we could solve that?

tijsverkoyen: not really

carakas: I think it will be best to leave it like this for now unless somebody else has a solution

PR #1794, opened 2016-10-19
fix example mailmotor templates

### LINKED ISSUE #1793: Mailmotor - templates still TPL format
## Type

- Critical bug

## Problem description

I ticked the Critical bug as without a solution to these, the mailmotor isn't functional.
Let me describe.

I use mailmotor with campaignmonitor (createsend) and am migrating to version 4.2.x (coming from 3.x.x), so I knew I had to adapt my own template to be functional.
While testing with the default and blank template included in the Fork release, I got an error "500 internal error". Checking the logs stated that the .html.twig files weren't present. And indeed, the templates are still the old ones in .tpl.

My knowledge is very basic (not to say absent) and simply changing the file format doesn't work, never the less the old template format is very basic html coding. and I guess I won't be the only one facing this.

## Steps to reproduce

Link createsend account to mailmotor, and try to create a mailing using one of the included templates. (with some bad luck you need to overrule the x-frame-options to see the iframe before getting the "internal error" to pop up).

Convert the tpl templates to html.twig.

Tnx in advance!

carakas:
## Type

- Critical bugfix

fixes #1793 

## Pull request description

Mailmotor was broken
This fixes it
Also it fixes that you couldn't configure a page to be available via iframe

### DIFF
--- a/src/Backend/Modules/Mailmotor/Templates/en/empty/template.html.twig
+++ b/src/Backend/Modules/Mailmotor/Templates/en/empty/template.html.twig
@@ -3,7 +3,7 @@
     <td>
       <!-- editor -->
         <textarea id="contentHtml" name="content_html" cols="62" rows="5" class="inputEditorNewsletter">
-          {{ contentHtml }}
+          {{ contentHtml|raw }}
         </textarea>
       <!-- /editor -->
     </td>
--- a/src/Backend/Modules/Mailmotor/Templates/en/empty/template.html.twig
+++ b/src/Backend/Modules/Mailmotor/Templates/en/empty/template.html.twig
@@ -3,7 +3,7 @@
     <td>
       <!-- editor -->
         <textarea id="contentHtml" name="content_html" cols="62" rows="5" class="inputEditorNewsletter">
-          {$contentHtml}
+          {{ contentHtml }}
         </textarea>
       <!-- /editor -->
     </td>
@@ -14,7 +14,7 @@
   </tr>
   <tr>
     <td>
-      <p><small>Ne voulez-vous plus recevoir ce bulletin d'email? <a id="unsubscribeURL" href="#">D&eacute;sinscrivez-vous</a> sur notre site web.</small></p>
+      <p><small>Do you no longer wish to receive this newsletter? <a id="unsubscribeURL" href="#">Unsubscribe</a> on our website.</small></p>
     </td>
   </tr>
 </table>
--- a/src/Backend/Modules/Mailmotor/Actions/EditMailingIframe.php
+++ b/src/Backend/Modules/Mailmotor/Actions/EditMailingIframe.php
@@ -9,6 +9,7 @@ namespace Backend\Modules\Mailmotor\Actions;
  * file that was distributed with this source code.
  */
 
+use Symfony\Component\HttpFoundation\Response;
 use TijsVerkoyen\CssToInlineStyles\CssToInlineStyles;
 use Backend\Core\Engine\Base\ActionEdit as BackendBaseActionEdit;
 use Backend\Core\Engine\Model as BackendModel;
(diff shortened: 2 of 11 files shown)


src/Backend/Modules/Mailmotor/Templates/en/empty/template.html.twig
```diff
@@ -14,7 +14,7 @@
   </tr>
   <tr>
     <td>
-      <p><small>Wens je deze nieuwsbrief niet meer te ontvangen? <a id="unsubscribeURL" href="#">Schrijf je uit</a> op de website.</small></p>
+      <p><small>Do you no longer wish to receive this newsletter? <a id="unsubscribeURL" href="#">Unsubscribe</a> on our website.</small></p>
```
StijnVrolijk: nice catch, doesn't really belong in the commit, though ;)

Lennerd320i:
Hi, this has been implemented in the 4.3.x version, but still has a minor issue (besides I had to change manually the x-frame-option).
When creating a mailing, within the EditMailingIframe, an error exists, creating 2 issues:
1) jquery.js:5 Uncaught Error: jQuery should be loaded before CKEditor jQuery adapter.(…)
2) Uncaught TypeError: $(...).ckeditor is not a function(…)

carakas:
@Lennerd320i can you create a PR with your fixes?

Lennerd320i:
I haven't "found" the solution yet :)

carakas:
The module will soon completely be replaced with a new version, in that version your problems should be fixed anyway

Lennerd320i:
I know, but in the meantime I'm paying $29 per month for the createsend account, and can't send any newsletters.

carakas:
on what step do you get this error?

Lennerd320i:
Within step 3.
I've been comparing the 3.x.x scripts with the 4.x.x scripts, and can't see a logic solution, except that in the 3.x.x scripts a whole bunch of JS files was loaded, but when trying to load these manually, I get other errors (exceptions).

carakas:
I'll look into creating a fix for it :)

Lennerd320i:
tnx @carakas and sorry for the waste of time!

carakas:
No problem, I also don't really know when the new module will be finished so it is worth fixing this.

I'll post a link to the pr with the fix in here within 24 hours

Lennerd320i:
Great, thanks!

carakas:
@Lennerd320i This should work https://github.com/forkcms/forkcms/pull/1840
If you still bump into any issues, feel free to let me know

Lennerd320i:
Thanks, I will test it asap and come back to you with feedback.

carakas:
@Lennerd320i have you already been able to test it?

Lennerd320i:
@carakas I just implemented the file changes for the 8 files.

In the backend, when using template "fork cms", I get an error 500 within step 3:
it is pointing towards the edit_mailing_iframe part and gives an error on the "iframebox" (<iframe id="contentBox" src="/private/nl/mailmotor/edit_mailing_iframe?token=vpmebdpxnk&amp;id=3" height="100%" width="100%"></iframe>) (I didn't edit the error).

When using another template (the 2 other available stock within fork) I get within step 3 this error:
Uncaught Error: Query should be loaded before CKEditor jQuery adapter.
&
Uncaught jsData is not available

This causes me to be able to enter plain text, but no CKEditor available.

I can continue to step 4 when using G. Chrome, but not with IE (and yes I'm one of the rare users still working quite everytime with IE).

Tnx so far!

kr,
Lennerd

carakas:
I'll try to look into it by the end of the week, but I'm not sure if I'll have time

Lennerd320i:
No problem at all @carakas , I do appreciate the effort done so far!

PR #1796, opened 2016-10-19
Officially deprecate our own events in favour of symfony events

carakas:
## Type

- Deprecation

Like asked in https://github.com/forkcms/forkcms/pull/1125


carakas:
The methods I added the trigger_error in aren't used
The other once I added @deprecated
The actual symfony event will be added when they are rewritten for doctrine and symfony forms

Issue #1714, opened 2016-08-31
Symfony integration


WouterSioen:
- [x] Replace SpoonCookie/SpoonSession with Request/Response object #399
- [x] Refactor Fork hooks using EventDispatcher #405
- [ ] Refactor locale module using Translation #408
- [ ] Refactor Fork and Pages with Routing component #409
- [ ] Replace SpoonThumbnail/SpoonFile #708
- [x] Refactor out all exit statements #1105
- [ ] Refactor out constants #1106
- [x] Refactor API #1142
- [x] Refactor out SpoonFilter::getGetValue and SpoonFilter::getPostValue #1347
## Replace SpoonCookie/SpoonSession with Request/Response object #399

We need to pass a response object to every action/widget/... that can manipulate the sessions and cookies.
## Refactor Fork hooks using EventDispatcher #405

step 1: delete all triggerEvents + deprecate the triggerEvent method
step 2: add events for everything
## Refactor Fork and Pages with Routing component #409

We already use the router for our applications. We can probably use the ChainedRouter to do the further routing in the frontend based on our Pages module.
## Replace SpoonThumbnail/SpoonFile #708

There's probably a package for this.
## Refactor out all exit statements #1105

Replace them with good clear exceptions. This will also help us test more stuff.

Done in #1797
## Refactor out constants #1106

Container parameters are the way to go
## Remove API #1142

The API needs to be deprecated and altered by an easy way to add new API endpoints. This will also fix the usage of SpoonTemplate. 

We can use a combination of bundles to achieve these results (https://github.com/nelmio/NelmioApiDocBundle, https://github.com/FriendsOfSymfony/FOSRestBundle, ...)
## Refactor out SpoonFilter::getGetValue and SpoonFilter::getPostValue #1347

We should use the request object for this.

myspulin:
Hi, I do some development on the last version API and I wonder why is class BaseApi deprecated. In the comments there was mentioned bundle NelmioApiDocBundle or FOSRestBundle is it already decided which one will be used? Is there any estimation when will be implemented? Thanks for info.

myspulin:
@jessedobbelaere, thanks for update, it make sense to remove legacy code first I cant wait for new version. Regarding FOSRestbundle I will try the Slack and ask the others. Thank you!

carakas:
@myspulin you can find a demo implementation of the FOSRestBundle here https://github.com/forkcms/forkcms/compare/master...justcarakas:rest-api-challange

[...]

PR #1797, opened 2016-10-19
1714 refactor out exit statements

carakas:
## Type

- Enhancement

is part of #1714 but doesn't resolve it yet

## Pull request description

Refactors away the exit statements


src/Backend/Core/Ajax/Templates.php
```diff
@@ -35,38 +36,34 @@ public function execute()
...
-        header('Content-type: text/javascript');
-
-        // output the templates
-        if (!empty($templates)) {
-            echo 'CKEDITOR.addTemplates(\'default\', { imagesPath: \'/\', templates:' . "\n";
-            echo json_encode($templates) . "\n";
-            echo '});';
-        }
-        exit;
+        throw new RedirectException(
```
carakas: note to self, refactor this with the method getContent

src/Backend/Modules/Locale/Actions/Export.php
```diff
@@ -138,7 +140,6 @@ public function execute()
         parent::execute();
```
carakas: this can be removed

src/Backend/Modules/Locale/Actions/ExportAnalyse.php
```diff
@@ -59,7 +61,6 @@ public function execute()
         parent::execute();
```
carakas: this can be removed

PR #1801, opened 2016-10-22
FIX send mail by a Cronjob / URL object not available

wouterverstuyf:
## Type
- Non critical bugfix

When sending an email from a cronjob (by Common/Mailer) the system throws an error, because the URL object is not known in the Cronjob.
Checking on the existence of the object fixes the error.

## Pull request description

Implemented class_exists().
It works, but if there is a better way let me know!

### DIFF
--- a/src/Backend/Core/Engine/TwigTemplate.php
+++ b/src/Backend/Core/Engine/TwigTemplate.php
@@ -149,7 +149,7 @@ class TwigTemplate extends BaseTwigTemplate
         );
 
         // check on url object
-        if(class_exists('url')) {
+        if (class_exists('url')) {
 
             $url = Model::get('url');
 
@@ -391,7 +391,7 @@ class TwigTemplate extends BaseTwigTemplate
         $this->assign('timestamp', time());
 
         // check on url object
-        if(class_exists('url')) {
+        if (class_exists('url')) {
 
             // assign body ID
             $url = Model::get('url');
--- a/src/Backend/Core/Engine/TwigTemplate.php
+++ b/src/Backend/Core/Engine/TwigTemplate.php
@@ -148,28 +148,33 @@ class TwigTemplate extends BaseTwigTemplate
             Model::getContainer()->getParameter('site.multilanguage')
         );
 
-        $url = Model::get('url');
+        // check on url object
+        if(class_exists('url')) {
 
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Engine/TwigTemplate.php
```diff
@@ -148,28 +148,33 @@ private function parseConstants()
             Model::getContainer()->getParameter('site.multilanguage')
         );
 
-        $url = Model::get('url');
+        // check on url object
+        if (class_exists('url')) {
```
carakas: you need to check if the container has the 'url' service here.
It isn't a class name but a service name
The correct implementation here would be `Model::getContainer()->has('url')`

src/Backend/Core/Engine/TwigTemplate.php
```diff
@@ -385,22 +390,27 @@ private function parseVars()
         // assign current timestamp
         $this->assign('timestamp', time());
 
-        // assign body ID
-        $url = Model::get('url');
-        if ($url instanceof Url) {
-            $this->assign('bodyID', \SpoonFilter::toCamelCase($url->getModule(), '_', true));
+        // check on url object
+        if (class_exists('url')) {
```
carakas: same issue here

src/Backend/Core/Engine/TwigTemplate.php
```diff
@@ -148,28 +148,33 @@ private function parseConstants()
             Model::getContainer()->getParameter('site.multilanguage')
         );
 
-        $url = Model::get('url');
+        // check on url object
+        if (class_exists('url')) {
```
carakas: This empty line should be removed, see the results of the Style CI

src/Backend/Core/Engine/TwigTemplate.php
```diff
@@ -148,28 +148,33 @@ private function parseConstants()
...
+                        BACKEND_PATH . '/' . $url->getModule()
+                    );
+                } else {
+                    $this->assign(
+                        'BACKEND_MODULE_PATH',
+                        BACKEND_MODULES_PATH . '/' . $url->getModule()
+                    );
+                }
             }
+
```
carakas: This empty line should be removed, see the results of the Style CI

PR #1811, opened 2016-10-26
Make sure icons have no margin right when no text is next to it

StijnVrolijk:
## Type

- Non critical bugfix

Recently I changed the icon's padding-right to a margin-right to provide better support for animations. I oversaw the use in icon-only buttons in the page module which now had a margin of 6px to the right where previously the padding would be nullified. I now did the same with the margin


src/Backend/Core/Layout/Sass/components/_pages.scss
```diff
@@ -221,6 +221,7 @@
 
         .fa {
           padding-right: 0;
+          margin-right: 0;
```
carakas: where do you check that there is no text in it?

Katrienvh:
👍🏼

src/Backend/Core/Layout/Sass/components/_pages.scss
StijnVrolijk: i don't explicitly but the full selector is   
```
.templatePositionCurrentType:hover .buttonHolder .btn .fa {}
```
`.templatePositionCurrentType` doesn't contain any buttons with text and never will

PR #1813, opened 2016-10-27
Fix incorrect use of $this->header->addJS

StijnVrolijk:
## Type
- Critical bugfix

You got a filemtime(): start failed error on the backend location index and edit actions, plus the maps didn't work, obviously.
## Pull request description

Fixed the way the LocationMapStyles.js map was included in the backend, this was a poor copy of the frontend inclusions, this caused our module to be '1'


StijnVrolijk:
@jeroendesloovere the 'true' in `$this->header->addJS()` is the `$module` parameter, causing the module to be `1`, which the system can't find, obviously.

The 'true' you're thinking about is in your method to get the path to the JS file, which I didn't alter in any way

PR #1819, opened 2016-10-29
Feature page auth

ciryk:
## Type
- Feature

## Pull request description

Gives access to a new tab 'Authentication' within the pages module. Which let's you secure pages for logged in users or specific profile groups. It will remove the menu item from the navigation and disables access to that page directly if you don't have the access to it.

Documentation PR: https://github.com/forkcms/documentation/pull/65


.idea/copyright/profiles_settings.xml
```diff
@@ -0,0 +1,3 @@
+<component name="CopyrightManager">
```
carakas: this file shouldn't be in here
it is best to add `.idea` to your global htaccess

.idea/fork.iml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
carakas: this file shouldn't be in here
it is best to add `.idea` to your global htaccess

.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
carakas: this file shouldn't be in here
it is best to add `.idea` to your global htaccess

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
carakas: this file shouldn't be in here
it is best to add `.idea` to your global htaccess

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -253,6 +253,35 @@ private function loadForm()
         $this->frm->addImage('image');
         $this->frm->addCheckbox('remove_image');
 
+        // page auth related fields
+        // check if profiles module is installed
+        if (BackendModel::isModuleInstalled('Profiles')) {
+            // check data for auth_required
+            $auth_required = false;
```
carakas: we use camelCase for naming our variables

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -253,6 +253,35 @@ private function loadForm()
...
+            if (isset($this->record['data']['auth_required']) && $this->record['data']['auth_required']) {
+                $auth_required = true;
+            }
+            // add checkbox for auth_required
+            $this->frm->addCheckbox('auth_required', $auth_required);
```
carakas: you can simplify this,
in the if you check something and if it is true you pass true to the checkbox
You can use that check directly so the if won't be needed

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -253,6 +253,35 @@ private function loadForm()
...
+            // get all groups and parse them in key value pair
+            $groupItems = \Backend\Modules\Profiles\Engine\Model::getGroups();
```
carakas: there is already a usestatement for this, you can use `BackendModel`

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -253,6 +253,35 @@ private function loadForm()
...
+            if (!empty($groupItems)) {
+                $groups = array();
+                foreach ($groupItems as $key => $item) {
+                    $groups[] = array('label' => $item, 'value' => $key);
+                }
+                // set checked values
+                $checkValues = array();
```
carakas: why `checkValues`? I think `checkedGroups` or `enabledGroups` or something like it would be a better name

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -253,6 +253,35 @@ private function loadForm()
...
+                if (!empty(isset($this->record['data']['auth_groups']))) {
```
carakas: why do you do an empty check on the boolean returned by isset?

carakas: I also would use `is_array` instead of !empty since you loop over the array in the line below this one, if it is empty it just won't loop

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -556,6 +585,9 @@ protected function parse()
 
         // parse the tree
         $this->tpl->assign('tree', BackendPagesModel::getTreeHTML());
+
+        // assign if profiles module is installed
+        $this->tpl->assign('authentication', BackendModel::isModuleInstalled('Profiles'));
```
carakas: I would name this something like `showAuthenticationTab` since it better describes the purpose, or `profilesModuleIsInstalled`

src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -448,6 +450,16 @@ jsBackend.pages.extras =
 		}
 	},
 
+	// show the groups for authentication
+	showGroups: function(e)
+	{
+		// save element to variable
+		var element = $(this);
```
carakas: you don't use this variable

src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -448,6 +450,16 @@ jsBackend.pages.extras =
...
+		var list = $('#groupList');
```
carakas: It would be safer to use something like `.js-authentication-groups'
That way if we do a rewrite we won't accidentally break the js since we know js is listening for something here

src/Frontend/Core/Engine/Page.php
```diff
@@ -110,6 +111,35 @@ public function load()
             $this->record = Model::getPage(404);
         }
 
+
+        // authentication
+        if (BackendModel::isModuleInstalled('Profiles') && isset($this->record['data']['auth_required'])) {
+            $data = $this->record['data'];
+            // is auth required and is profile logged in
+            if ($data['auth_required']) {
+                if (!\Frontend\Modules\Profiles\Engine\Authentication::isLoggedIn()) {
```
carakas: You should add a use statement for this

src/Frontend/Core/Engine/Page.php
```diff
@@ -110,6 +111,35 @@ public function load()
...
+                    throw new RedirectException(
+                        'Redirect',
+                        new RedirectResponse(Navigation::getURLForBlock('Profiles', 'Login') . '?queryString=' . $queryString)
+                    );
+                }
+                // specific groups for auth?
+                if (!empty($data['auth_groups'])) {
+                    $inGroup = false;
+                    foreach ($data['auth_groups'] as $group) {
+                        if (\Frontend\Modules\Profiles\Engine\Authentication::getProfile()->isInGroup($group)) {
```
carakas: idem

.idea/copyright/profiles_settings.xml
carakas: thx @jessedobbelaere  🍰

src/Backend/Modules/Pages/Actions/Edit.php
ciryk: How would you do this? Here we have to check for two things, if it exists and if it is true?

.htaccess
```diff
@@ -91,4 +91,4 @@ FileETag MTime Size
 
 	# browser issues
 	mod_gzip_item_exclude reqheader "User-agent: Mozilla/4.0[678]"
-</IfModule>
+</IfModule>
```
carakas: you need to have a new line at the end of the file just like it was

src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -454,10 +454,8 @@ jsBackend.pages.extras =
 	showGroups: function(e)
 	{
 		// save element to variable
-		var element = $(this);
-		var list = $('#groupList');
-
-		list.toggle();
+		var authenticationGroups = $('.js-authentication-groups');
```
carakas: why don't you use it directly without the variable?
```
$('.js-authentication-groups').toggle();
```

PR #1824, opened 2016-11-04
Prevent scroll to top after click a cookievar button

Katrienvh:
## Type

- Non critical bugfix

## Pull request description

when you click on a button in the cookiebar then window scrolls to the top


src/Frontend/Core/Js/frontend.js
```diff
@@ -85,11 +85,13 @@ jsFrontend.cookieBar =
 		}
 
 		$cookieBar.on('click', '#cookieBarAgree', function(e) {
+            e.preventDefault();
```
carakas: The indentation is different

carakas: @Katrienvh can you fix this?

PR #1825, opened 2016-11-04
Fix issues with invalid backend credentials

carakas:
## Type

- Non critical bugfix

## Pull request description

Check if the login credentials are valid


carakas:
@StijnVrolijk had lost his laptop and reviewed it on mine

PR #1826, opened 2016-11-06
Return a clone of the file when there are changes

carakas:
## Type

- Enhancement

## Pull request description

This will make it easier to let doctrine detect the changes
SInce it is a value object you had to add a changed on to make doctrine update the entity
By returning a clone when there is a change the object reference will have changed causing doctrine to detect it


carakas:
a better way would be to create a form type that handles it for you

PR #1827, opened 2016-11-07
Removed duplicate for "SITE_MULTILANGUAGE"

jeroendesloovere:
Because a few lines above this is already set, because it's a parameter.


carakas:
@jeroendesloovere did you test this?

jeroendesloovere:
Yes, it was something that I came around when creating https://github.com/forkcms/forkcms/pull/1828

PR #1828, opened 2016-11-07
Change "showModuleAction" variables to "allowModuleAction"

### LINKED ISSUE #1777: Can't overwrite {% if showBlogDelete %}
## Type
- Non critical bug
## Problem description

> If for any reason you want to overwrite the default "showBlogDelete", this seems to be not possible anymore. Which I think was possible in older fork cms 3.9.\* versions (which used SpoonTemplate).
## Steps to reproduce

> You can try to add the following in your parse() method in the Edit.php action class of the Blog or ContentBlocks module:

``` php
$this->tpl->assign('showBlogDelete', false);
```

> Expected behavior is that the delete button is gone. But it is still available...

jeroendesloovere:
Fixes #1777

### DIFF
--- a/src/Backend/Modules/Blog/Layout/Templates/EditCategory.html.twig
+++ b/src/Backend/Modules/Blog/Layout/Templates/EditCategory.html.twig
@@ -50,7 +50,7 @@
     <div class="col-md-12">
       <div class="btn-toolbar">
         <div class="btn-group pull-left" role="group">
-          {% if showBlogDeleteCategory %}
+          {% if showBlogDeleteCategory and allowBlogDeleteCategory %}
             <button type="button" class="btn btn-danger" data-toggle="modal" data-target="#confirmDelete">
               <span class="fa fa-trash-o"></span>
               {{ 'lbl.Delete'|trans|ucfirst }}
@@ -63,7 +63,7 @@
           </button>
         </div>
       </div>
-      {% if showBlogDeleteCategory %}
+      {% if showBlogDeleteCategory and allowBlogDeleteCategory %}
         <div class="modal fade" id="confirmDelete" tabindex="-1" role="dialog" aria-hidden="true">
           <div class="modal-dialog">
             <div class="modal-content">
--- a/src/Backend/Modules/Pages/Actions/Edit.php
+++ b/src/Backend/Modules/Pages/Actions/Edit.php
@@ -541,6 +541,9 @@ class Edit extends BackendBaseActionEdit
             $showDelete = false;
         }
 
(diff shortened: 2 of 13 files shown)


carakas:
@jeroendesloovere can you fix the errors that showed up on the CI

jeroendesloovere:
@carakas those errors aren't mine?

carakas:
@jeroendesloovere I think they are, maybe your code has side effects.

Because master doesn't have these errors

jeroendesloovere:
Discovered a side effect when clicking to "Modules".
schermafbeelding 2016-11-07 om 13 20 09

jeroendesloovere:
When you visit Dashboard, the BaseTwigTemplate is instantiated three times.
So that makes my code useless.

jeroendesloovere:
This should not be possible. Since they should be run in a sandbox.
I'm investigating this further.

jeroendesloovere:
The problem lies at the following rule:
https://github.com/forkcms/forkcms/blob/master/src/Backend/Core/Engine/TwigTemplate.php#L39

We can solve this by removing the if statement and the function parameter and just saying:
```
Model::getContainer()->set('template', $this);
```
But don't know if that gives further problems...
Since template is begin used at more places:
schermafbeelding 2016-11-07 om 13 51 43

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -541,9 +541,6 @@ protected function parse()
             $showDelete = false;
         }
 
-        // show delete button
-        $this->tpl->assign('showPagesDelete', $showDelete);
```
carakas: this isn't the fix, you should use a different template variable name for this. you are changing the intended implementation

src/Backend/Modules/Blog/Layout/Templates/EditCategory.html.twig
```diff
@@ -63,7 +63,7 @@
           </button>
         </div>
       </div>
-      {% if showBlogDeleteCategory %}
+      {% if showBlogDeleteCategory and allowBlogDeleteCategory %}
```
carakas: isn't the showBlogDeleteCategory variable already checked in the BackendBlogModel?

please read the code very well when making changes like this

jeroendesloovere: @carakas This is something different.
See: https://github.com/forkcms/forkcms/pull/1828/files#diff-8f7f5362034515c7d4353de51b90b63aR83

Issue #1783, opened 2016-10-16
Page detail buttons look strange in dutch


carakas:
## Type
- Non Critical Bug
## Problem description

<img width="900" alt="screen shot 2016-10-16 at 13 49 55" src="https://cloud.githubusercontent.com/assets/3634654/19417229/94091c5a-93a7-11e6-84af-e327f92d931e.png">

these buttons look a bit odd
## Steps to reproduce

Set the backend language to dutch
## Expected behavior / Proposed solutions

nicer spacing between the buttons, preferable still next to each other or otherwise floated to the right

PR #1829, opened 2016-11-07
Fix styling when page detail buttons stand on two lines

Katrienvh:
## Type

- Non critical bugfix

#1783 

## Pull request description

solves #1783


carakas:
@Katrienvh can you add a screenshot of the proposed fix?

PR #1830, opened 2016-11-07
Fixing Backend sandbox to stand alone

jeroendesloovere:
I've discovered that the Backend uses previous template ($this->get('template)) for every widget that it creates. This isn't correct behavior for sandboxed templates. They should be empty at start.

Note: The same occurs in Frontend... When you try generating a widget, all previously assigned variables are gone, poefff, foetsie!!!

More information:

**This doesn't work**
```php
    /**
     * Parse
     */
    protected function parse()
    {
        // Assign MediaGallery
        $this->tpl->assign(
            'mediaGallery',
            ($this->mediaGallery) ? $this->mediaGallery : false
        );

        // Assign widget
        $this->tpl->assign(
            'widget',
            // We can create widget for the MediaGroup id
            $this->get('media_library.helper.frontend')->parseWidget(
                $this->mediaGallery->getAction(),
                $this->mediaGroup->getId()
            )
        );
    }
```
> Because we define "widget" (which generated a widget) after "mediaGallery", the widget sandbox deletes everything from the parent sandbox (= this is a BUG). Which will mean that "mediaGallery" is not parsed to the template.

Temporary fix:
Moving the "mediaGallery" assign below the "widget" assign fixes everything...

Another problem:
When you want to assign "two generated widgets" new problems occur, since the first widget will now be removed (because of this sandbox error).

Temporary fix for that:
```php
$this->tpl->assignArray(
    array(
        array(
            'firstWidget',
            // We can create widget for the MediaGroup id
            $this->get('media_library.helper.frontend')->parseWidget(
                $this->mediaGallery->getAction(),
                $this->mediaGroup->getId()
            ),
            'secondWidget',
            // We can create widget for the MediaGroup id
            $this->get('media_library.helper.frontend')->parseWidget(
                $this->mediaGallery->getAction(),
                $this->mediaGroup->getId()
            )
        ) 
    )
)
```

### DIFF
--- a/src/Backend/Core/Engine/TwigTemplate.php
+++ b/src/Backend/Core/Engine/TwigTemplate.php
@@ -30,14 +30,11 @@ class TwigTemplate extends BaseTwigTemplate
 {
     /**
      * The constructor will store the instance in the reference, preset some settings and map the custom modifiers.
-     *
-     * @param bool $addToReference Should the instance be added into the reference.
      */
-    public function __construct($addToReference = true)
+    public function __construct()
     {
-        if ($addToReference) {
-            Model::getContainer()->set('template', $this);
-        }
+        // We must set the template for later use
+        Model::getContainer()->set('template', $this);
 
         $this->forkSettings = Model::get('fork.settings');
         $this->debugMode = Model::getContainer()->getParameter('kernel.debug');
--- a/src/Backend/Core/Engine/Base/Action.php
+++ b/src/Backend/Core/Engine/Base/Action.php
@@ -62,7 +62,7 @@ class Action extends Object
         parent::__construct($kernel);
 
         // get objects from the reference so they are accessible from the action-object
-        $this->tpl = $this->getContainer()->get('template');
+        $this->tpl = new TwigTemplate();
         $this->URL = $this->getContainer()->get('url');
         $this->header = $this->getContainer()->get('header');
 
--- a/src/Backend/Core/Engine/Base/Widget.php
+++ b/src/Backend/Core/Engine/Base/Widget.php
@@ -71,7 +71,7 @@ class Widget extends \KernelLoader
     {
         parent::__construct($kernel);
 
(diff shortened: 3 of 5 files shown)


jeroendesloovere:
Discovered a side effect when clicking to "Modules".
schermafbeelding 2016-11-07 om 13 20 09
=> Fixed

When you visit Dashboard, the BaseTwigTemplate is instantiated three times (probably once at start and then two times for dashboard widgets).
So that makes my code useless.

This should not be possible. Since they should be run in a sandbox.
I'm investigating this further.

The problem lies at the following rule:
https://github.com/forkcms/forkcms/blob/master/src/Backend/Core/Engine/TwigTemplate.php#L39

We can solve this by removing the if statement and the function parameter and just saying:
```
Model::getContainer()->set('template', $this);
```
But don't know if that gives further problems...
Since template is being used at more places:
schermafbeelding 2016-11-07 om 13 51 43

src/Backend/Core/Engine/TwigTemplate.php
```diff
@@ -30,14 +30,11 @@ class TwigTemplate extends BaseTwigTemplate
 {
     /**
      * The constructor will store the instance in the reference, preset some settings and map the custom modifiers.
-     *
-     * @param bool $addToReference Should the instance be added into the reference.
      */
-    public function __construct($addToReference = true)
+    public function __construct()
     {
-        if ($addToReference) {
```
carakas: I already told you this couldn't be removed

carakas:
ci fails

carakas:
@jeroendesloovere can you look at the feedback and the failed ci build

src/Backend/Core/Engine/TwigTemplate.php
jeroendesloovere: Hmmm, I should investigate further then, because removing this, I fixed everything.

carakas:
@jeroendesloovere poke

PR #1840, opened 2016-11-18
Mailmotor should not have js

carakas:
## Type

- Non critical bugfix

## Pull request description

Some additional fixes for mailmotor


carakas:
see comment on #1794

carakas:
merged this as reference for old forks that might need to be fixed before it is replaced with the new module

PR #1846, opened 2016-11-23
Fix serialize bug in meta entity SEOFollow and SEOIndex

carakas:
## Type

- Critical bugfix

## Pull request description

The meta entity saved the SEOFollow and SEOIndex value objects wrong in the database

For backwards compatibility I added a regex replace that fixes the issue when loading the meta entity.

After saving it once the database should be ok.

### DIFF
--- a/src/Common/Doctrine/ValueObject/SEOFollow.php
+++ b/src/Common/Doctrine/ValueObject/SEOFollow.php
@@ -2,6 +2,8 @@
 
 namespace Common\Doctrine\ValueObject;
 
+use Serializable;
+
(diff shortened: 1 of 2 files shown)


src/Common/Doctrine/ValueObject/SEOFollow.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Common\Doctrine\ValueObject;
 
+use Serializable;
```
StijnVrolijk: Shouldn't this be implemented?

PR #1847, opened 2016-11-24
Fix get url for block fixes

carakas:
## Type

- Non critical bugfix

## Pull request description

the backend didn't take the data into account at all and in the frontend it didn't work properly if the action was null

### DIFF
--- a/src/Backend/Core/Engine/Model.php
+++ b/src/Backend/Core/Engine/Model.php
@@ -610,13 +610,16 @@ class Model extends \Common\Core\Model
     /**
      * Get the URL for a give module & action combination
      *
-     * @param string $module   The module to get the URL for.
-     * @param string $action   The action to get the URL for.
-     * @param string $language The language to use, if not provided we will use the working language.
+     * @param string $module   The module wherefore the URL should be build.
+     * @param string $action   The specific action wherefore the URL should be build.
+     * @param string $language The language wherein the URL should be retrieved,
+     *                         if not provided we will load the language that was provided in the URL.
+     * @param array $data      An array with keys and values that partially or fully match the data of the block.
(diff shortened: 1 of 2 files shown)


src/Backend/Core/Engine/Model.php
```diff
@@ -625,22 +628,52 @@ public static function getURLForBlock($module, $action = null, $language = null)
...
                         continue;
                     }
 
                     // loop extras
                     foreach ($properties['extra_blocks'] as $extra) {
-                        if ($extra['module'] == $module && $extra['action'] == $action) {
+                        // direct link?
+                        if ($extra['module'] == $module && $extra['action'] == $action  && $extra['action'] !== null) {
+                            // if there is data check if all the requested data matches the extra data
+                            if (isset($extra['data']) && $data !== null
```
StijnVrolijk: shouldnt `&& $data !== null` be on the next line?

src/Frontend/Core/Engine/Navigation.php
```diff
@@ -514,15 +515,24 @@ public static function getURLForBlock($module, $action = null, $language = null,
 
                         if ($extra['module'] == $module && $extra['action'] == null) {
                             // if there is data check if all the requested data matches the extra data
-                            if (isset($extra['data']) && $data !== null
-                                && array_intersect_assoc($data, (array) $extra['data']) !== $data) {
-                                // It is the correct module but has the wrong data
-                                continue;
+                            if (isset($extra['data']) && $data !== null) {
+                                if (array_intersect_assoc($data, (array) $extra['data']) !== $data) {
```
StijnVrolijk: well never mind :) cancel my last comment, Siri!

PR #1854, opened 2016-11-28
Update Formbuilder mailing form: Form.html.twig

### LINKED ISSUE #1852: Form Builder - enter is being displayed as <br />
## Type

- Non critical bug

## Problem description

When someone adds text in the "textarea" part of a form, and the form is sent, every "enter" is displayed within the mail as <br />.

## Steps to reproduce

Create a form with a textarea, go through the front end, try e.g. test ENTER test and send it. In the mail sent, the enter will be html wise translated to <br />.

translate the <br /> stored within the database back.

Maybe it's related to the Original issue with ampersand within issue #1649 ?

P.S.: I had this issue as from the 4.x version, and tested it now with both 4.2.3 and 4.3.1.

kr,

Lennerd

Lennerd320i:
## Type

- Non critical bugfix

fixes #1852 
html translations like "< br / >" stored in the DB will be sent in the mail as "enter"

## Pull request description

add |raw within the message content so html translations like "< br / >" are translated to "enter".
Issue mentioned within #1852 and solution proposed by @carakas

### DIFF
--- a/src/Frontend/Modules/FormBuilder/Layout/Templates/Mails/Form.html.twig
+++ b/src/Frontend/Modules/FormBuilder/Layout/Templates/Mails/Form.html.twig
@@ -11,6 +11,6 @@
 
   <h3>{{ 'lbl.Content'|trans|ucfirst }}</h3>
   {% for field in fields %}
-    <p><strong>{{ field.label }}:</strong><br /> {{ field.value }}</p>
+    <p><strong>{{ field.label }}:</strong><br /> {{ field.value|raw }}</p>
   {% endfor %}
 {% endblock %}


bart-webleads:
Hmmm, removing the nl2br in FormBuilderSubmittedMailSubscriber->getEmailFields should and adding it in the template should to the trick. Wil test when time permits...

bart-webleads:
I have fixed it in my current project.
I also have added a "file" upload field type. This need some more testing.
When I've tested it I will create a PR including the file field type.

bart-webleads:
At the moment the files are stored in frontend/files. The filename has a uniq id. This will be hard to guess.

bart-webleads:
Yep. That should be the best way.

PR #1862, opened 2016-11-30
Fix pagination last and first url and label.

### LINKED ISSUE #1861: Last and first pagination url and label are not displayed
## Type

- Non critical bug

## Problem description
The value of url and label for first and last item in pagination are null.

## Steps to reproduce

Create a log of blog posts to make pagination displayed.

I will provide a pull request.

GhaziTriki:
## Type

- Non critical bugfix

Fixes #1861 

## Pull request description

Many thanks to Marouene Agrebi for finding this issue and suggesting the fix for it.

### DIFF
--- a/src/Frontend/Core/Layout/Templates/Pagination.html.twig
+++ b/src/Frontend/Core/Layout/Templates/Pagination.html.twig
@@ -16,7 +16,7 @@
             {% if pagination.first %}
               {% for pagination in pagination.first %}
                 <li>
-                <a href="{{ pagination.first.url }}" rel="nofollow" title="{{ 'lbl.GoToPage'|trans|ucfirst }} {{ pagination.first.label }}">{{ pagination.first.label }}</a>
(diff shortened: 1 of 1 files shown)


src/Frontend/Core/Layout/Templates/Pagination.html.twig
```diff
@@ -16,7 +16,7 @@
             {% if pagination.first %}
               {% for pagination in pagination.first %}
```
StijnVrolijk: Aren't these for loops the cause of the problem? There can only be one first and last item.
I think you can just drop the loops and use `{{ pagination.first.url }}` for example.

Can you try this @GhaziTriki ?

carakas: in the bootstrap theme it is like @GhaziTriki wrote, so I think this implementation is correct

StijnVrolijk: Yeah, you are correct. The structure is just a bit weird :p

PR #1863, opened 2016-11-30
Make sure that house number is always handled and shown before the street name

### LINKED ISSUE #1860: Google maps is geolocating incorrectly in some countries
## Type

- Non critical bug

## Problem description

ForkCMS puts the street number before the street name for geolocating addresses. The address number should come first.

## Steps to reproduce

Create a location with the following parameters:
- Street name: Rue de Annaba
- Street number: 16
- ZIP: 2074
- City: Mourouj

The geolocated address is incorrect.

Put the street number before the street name when geolocating.

GhaziTriki:
## Type

- Non critical bugfix

Fixes #1860 

## Pull request description

Street number comes before street name.


jeroendesloovere:
In Belgium (Flanders) street number comes behind street,
but in Belgium (Wallonie) street number comes before street...

carakas:
I'll close this one since it will fix your problem but create an issue for the people that need it in this format, we should think of a better solution and include that in fork 5

Issue #1866, opened 2016-12-02
<link rel="alternate" ...> is missing in Fork CMS


jeroendesloovere:
## Type

- Enhancement

## Problem description

> Fork CMS could actually automatically create <link rel="alternate" ...> meta tags because the "ID" is the same for pages if you copy/paste from one language to another.

<img width="871" alt="schermafbeelding 2016-12-02 om 09 07 53" src="https://cloud.githubusercontent.com/assets/588616/20826952/d5ae1a16-b86e-11e6-8f8a-1d9ae332316f.png">

> You can see that the ID is the same, but the language is different.

## Steps to reproduce

1. Install a new Fork CMS with *multiple languages*
2. Add pages in one language, then use the `copy` action from the Pages module
3. The new pages in the other languages will have the same ID.
> Which makes this perfectly possible to Fork CMS to create <link rel="alternate"> meta tags.

## Expected behavior / Proposed solutions

Fork CMS should automatically create the appropriate tags.
```html
<link rel="canonical" href="http://demo.fork-cms.com/nl/over-ons">
<link rel="alternate" hreflang="nl" href="http://demo.fork-cms.com/nl/over-ons">
<link rel="alternate" hreflang="fr" href="http://demo.fork-cms.com/fr/a-propos">
...
```

Note: the only page where Fork CMS does this correctly is on the Home page.

PR #1868, opened 2016-12-02
Create link meta tags for alternate languages

jeroendesloovere:
Fix for #1866

### DIFF
--- a/src/Frontend/Core/Engine/Page.php
+++ b/src/Frontend/Core/Engine/Page.php
@@ -490,23 +490,40 @@ class Page extends FrontendBaseObject
             );
         }
 
-        // get languages
-        $activeLanguages = Language::getActiveLanguages();
+        // Multi language is activated
+        if ($this->getContainer()->getParameter('site.multilanguage')) {
+            $links = array();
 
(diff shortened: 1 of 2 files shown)


src/Frontend/Core/Engine/Page.php
```diff
@@ -490,6 +490,25 @@ protected function processPage()
...
+            $url = Navigation::getURL($this->pageId, $language);
+
+            // Convert relative to absolute url
+            if (substr($url, 0, 1) == '/') {
+                $url = SITE_URL . $url;
+            }
+
+            // Add hreflang
+            $this->header->addLink(
+                array('rel' => 'alternate', 'hreflang' => $language, 'href' => $url)
```
carakas: you shouldn't add it if the link isn't found

carakas:
@jeroendesloovere can you fix the error that popped up in the tests?

jeroendesloovere:
Nope, can't fix it. Seems like a core fork cms bug.

carakas:
replaced by #1885 since it is based on this one but with passing tests

Issue #1838, opened 2016-11-13
Better detection for nginx in install


GhaziTriki:
## Type

- Enhancement

## Problem description

![image](https://cloud.githubusercontent.com/assets/4991088/20245694/f41f5b12-a9a6-11e6-823c-1ba366b83051.png)

Install is configured to detect Apache2 only.

## Steps to reproduce

Run a ForkCMS install on a nginx server.

## Expected behavior / Proposed solutions

Install should be able to detect nginx using `$_SERVER["SERVER_SOFTWARE"]`

carakas:
@GhaziTriki can you work on a fix for this since you already have the correct environment for this?

GhaziTriki:
@carakas I will take a look at it :)

carakas:
Going to close this one for now

PR #1872, opened 2016-12-05
Check rewrite for nginx in installer 

GhaziTriki:
## Type

- Enhancement

#1838 

## Pull request description

Detect if the web server is nginx or not.


carakas:
@GhaziTriki can you fix the tests

GhaziTriki:
@carakas I could net understand the test error. Is it related to my update? Is it the related to the fact I am not doing `isset` for `$_SERVER['SERVER_SOFTWARE']`?

carakas:
you can test it by running tests locally and see what the error is by getting the content where the statuscode is 500

GhaziTriki:
@carakas it seems like capitilising the method name did not fix the issue 😭

carakas:
The test runs as far as I can see, I think the issue is that your check triggers an unexpected result in the tests.
You assume `$_SERVER['SERVER_SOFTWARE']` is set, I think it might not be set when running the tests, so maybe check for that :)

src/ForkCMS/Bundle/InstallerBundle/Service/RequirementsChecker.php
```diff
@@ -278,13 +278,21 @@ protected function checkAvailableFiles()
...
      */
-    protected function checkApacheRewrites()
+    protected function checkServerRewrites()
     {
-        $this->checkRequirement(
-            'modRewrite',
-            (bool) (getenv('MOD_REWRITE') || getenv('REDIRECT_MOD_REWRITE')),
-            self::STATUS_WARNING
-        );
+        if (!preg_match('/^nginx/i', $_SERVER['SERVER_SOFTWARE'])) {
```
carakas: Try this
```
if (array_key_exists('SERVER_SOFTWARE', $_SERVER) && stripos($_SERVER['SERVER_SOFTWARE'], 'nginx') === 0) {
    $this->checkRequirement(
        'modRewrite',
        (bool) (getenv('MOD_REWRITE') || getenv('REDIRECT_MOD_REWRITE')),
        self::STATUS_WARNING
    );

    return;
}

$this->checkRequirement(
    'modRewrite',
    true,
    self::STATUS_WARNING
);
```

jessedobbelaere:
Now the test complains that a 302 doesnt match a 200. Strange because the test has `$client->followRedirect();` which should ignore 302's ?

src/ForkCMS/Bundle/InstallerBundle/Service/RequirementsChecker.php
```diff
@@ -278,13 +278,21 @@ protected function checkAvailableFiles()
...
-            (bool) (getenv('MOD_REWRITE') || getenv('REDIRECT_MOD_REWRITE')),
-            self::STATUS_WARNING
-        );
+        if (array_key_exists('SERVER_SOFTWARE', $_SERVER) && stripos($_SERVER['SERVER_SOFTWARE'], 'nginx') !== 0) {
+            $this->checkRequirement(
+                'modRewrite',
+                (bool) (getenv('MOD_REWRITE') || getenv('REDIRECT_MOD_REWRITE')),
+                self::STATUS_WARNING
+            );
+        } else {
```
carakas: you don't need this else if you add a return in the if like in the code example I provided :)

src/ForkCMS/Bundle/InstallerBundle/Service/RequirementsChecker.php
GhaziTriki: But do you agree that the condition should be `stripos($_SERVER['SERVER_SOFTWARE'], 'nginx') === false` instead of `stripos($_SERVER['SERVER_SOFTWARE'], 'nginx') === 0`

GhaziTriki:
Tests are still failing. I have tested the condition and works correctly in the browser. Not sure why tests are still unhappy.

StijnVrolijk:
@GhaziTriki are you still working on this?

carakas:
Going to close this one since there hasn't been progress in a very long time

PR #1876, opened 2016-12-09
Bugfix for SpoonFormException: The field "auth_required" does not exist.

bart-webleads:
## Type

Critical bugfix

## Pull request description

Fix for SpoonFormException: The field "auth_required" does not exist. You will get this exception when the profiles module is not installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+4.4.1 (2017-xx-yy)
+------------------
+
+Bugfixes:
+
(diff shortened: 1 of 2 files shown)


jeroendesloovere:
Duplicate for https://github.com/forkcms/forkcms/pull/1874

bart-webleads:
Yes duplicate, but with better solution ;-)

carakas:
It isn't a better solution, because the field is only added if the profiles module is installed. So it is more logical to check it like that. That way if something is added at a later Point it is clear what the code does

PR #1885, opened 2016-12-14
Add missing alternate tags and fix redirect loop on 404 page

### LINKED ISSUE #1866: <link rel="alternate" ...> is missing in Fork CMS

carakas:
## Type

- Bug Fix

fixes #1866

## Pull request description

Adds missing alternate tags to pages when we have them in an other language.

Also fixes the redirect loop on a 404

### DIFF
--- a/src/Frontend/Core/Engine/Header.php
+++ b/src/Frontend/Core/Engine/Header.php
@@ -223,7 +223,7 @@ class Header extends FrontendBaseObject
         $uniqueKeys = (array) $uniqueKeys;
 
         if ($uniqueKeys == null) {
-            $uniqueKeys = array('rel', 'type', 'title');
+            $uniqueKeys = array('rel', 'hreflang', 'type', 'title');
         }
 
         // stop if the content is empty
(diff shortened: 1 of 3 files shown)


jeroendesloovere:
👍

PR #1905, opened 2016-12-20
Backend: Changed css for .modal to overflow:auto

### LINKED ISSUE #1902: Backend css bug
## Type

- Non critical bug

## Problem description

> In the Backend: when having a modal that is higher then your screen size (f.e.: on a laptop). You can't scroll because `overflow: hidden` is being used instead of `overflow: auto` for `.modal`.

> Can a frontender fix this please?

jeroendesloovere:
Fixes #1902


carakas:
in branch v5.0.0-dev  this will be fixed by using a newer version of bootstrap

.modal-open .modal has overflow-y: auto

jeroendesloovere:
Hi, just tested this in v5.0.0-dev and that isn't a solution.

PR #1908, opened 2016-12-21
Dont index content for search

freshface:
Dont index content for search if authentication for the page is true.

## Type

- Enhancement

Dont index the page content of authentication for the page is true, else it will show up inde search results.

### DIFF
--- a/src/Backend/Modules/Pages/Actions/Edit.php
+++ b/src/Backend/Modules/Pages/Actions/Edit.php
@@ -262,6 +262,13 @@ class Edit extends BackendBaseActionEdit
                 'auth_required',
                 isset($this->record['data']['auth_required']) && $this->record['data']['auth_required']
             );
+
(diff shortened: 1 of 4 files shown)


carakas:
@freshface code style checks failed, can you look at this: https://styleci.io/analyses/qv07Wr

carakas:
Apart from that, I think it should be an option on the authentication page to remove it from the search and not something you can't do anything about

freshface:
Ok, how do we fix it?
On a "per page" base or as a general setting for the pages module "Don't search  index pages which are authenticated"

carakas:
I would do it per page in the authentication tab

carakas:
@freshface any update on this?

freshface:
I am checking the feedback atm.

freshface:
I changed it.

PR #1913, opened 2016-12-22
Make it language aware when needed.

freshface:
## Type

- Enhancement

When you get a tag or tags you can make it language aware.

### DIFF
--- a/src/Frontend/Modules/Tags/Engine/Model.php
+++ b/src/Frontend/Modules/Tags/Engine/Model.php
@@ -12,7 +12,7 @@ namespace Frontend\Modules\Tags\Engine;
 use Frontend\Core\Engine\Exception as FrontendException;
 use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Core\Engine\Navigation as FrontendNavigation;
-use Backend\Core\Language\Locale;
+use Frontend\Core\Language\Locale;
 
 /**
  * In this file we store all generic functions that we will be using in the tags module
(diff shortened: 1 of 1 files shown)


carakas:
@freshface can you look at the feedback of @jeroendesloovere

freshface:
I changed it.

src/Frontend/Modules/Tags/Engine/Model.php
```diff
@@ -12,6 +12,7 @@
 use Frontend\Core\Engine\Exception as FrontendException;
 use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Core\Engine\Navigation as FrontendNavigation;
+use Backend\Core\Language\Locale;
```
carakas: you need to use the frontend version :)

src/Frontend/Modules/Tags/Engine/Model.php
```diff
@@ -92,11 +93,14 @@ public static function getAll()
...
      * @return array
      */
-    public static function getForItem($module, $otherId)
+    public static function getForItem($module, $otherId, $language = null)
     {
         $module = (string) $module;
         $otherId = (int) $otherId;
 
+        // redefine language
+        $language = ($language !== null) ? (string) $language : Locale::workingLocale();
```
carakas: this method will also need to change to the frontend equivalent

carakas:
@freshface some new feedback

freshface:
I fixed it.

PR #1918, opened 2016-12-28
Yarn

tijsverkoyen:
## Type

- Enhancement
- Feature

## Pull request description

Install needed assets with [`yarn`](https://yarnpkg.com), and using `gulp` to generate the css based on the scss-files.

Yarn is a package manager which has a `lock`-file per default, so everybody should get the correct versions, this in favor of `npm`, which does not have a locking mechanism.

The `gulp build`-task will move the required assets into the correct folders and will regenerate the CSS-files for the backend.

At this moment the "moved" assets are included in the repo for two reasons: 

1. We don't have an automated way to include the assets when releasing
2. We want users to be able to install Fork CMS without any knowledge of Sass/Gulp/...

With this pull-request we should improve the disk-space usage by a fair bit, as the bower_components-folder is removed and only the required files are included.

Also we don't rely on `bower` anymore.


jessedobbelaere:
yay for yarn! yay for a smaller fork cms.

tijsverkoyen:
Note to self: check if jquery-ui-images are loaded

jessedobbelaere:
Nice PR. Big improvement! Looks good 👍 

Some remarks, not related to this PR:
* I was wondering why do we use jquery-migrate?
* Typeahead is only used in the search Module afaik? Or do we use it for tags etc? Maybe remove it from the core and keep it in the module folder-only
* In the future, we can maybe let Travis upload the artifact/compiled forkcms to Github releases (http://docs.travis-ci.com/user/deployment/releases/) ? Or how do other github projects use compiled binaries for their releases

src/Backend/Core/Layout/Sass/screen.scss
```diff
@@ -1,8 +1,8 @@
 @charset 'UTF-8';
 
 // Font Awesome
-$fa-font-path: '../../../../../bower_components/fontawesome/fonts';
-@import '../../../../../bower_components/fontawesome/scss/font-awesome';
+$fa-font-path: '/fonts';
```
Katrienvh: i should move this variable to _bootstrap-variables.scss

gulpfile.js
```diff
@@ -10,11 +14,34 @@ gulp.task("build:backend:assets:copy-js-vendors", function() {
...
 
+gulp.task("build:backend:sass:generate-css", function() {
+  return gulp.src([
+    "./src/Backend/Core/Layout/Sass/screen.scss",
+    "./src/Backend/Core/Layout/Sass/debug.scss",
+  ])
+      .pipe(sourcemaps.init())
+      .pipe(sass({
+        includePaths: [
+          "./node_modules/"
```
Katrienvh: Maybe you can also add the path for the bootstrap sass files, so you only need to import the file name in the _bootstrap-imports.scss file, the file will be clearer it think

tijsverkoyen: Personally, I found this more confusing, as you have to "know" there are multiple include-paths.

src/Backend/Core/Layout/Sass/screen.scss
tijsverkoyen: we can't move this, as the `font-awesome/scss/font-awesome`-file is loaded before our variables-file has been imported.

tijsverkoyen:
@jessedobbelaere 

> Currently we have folders "css" and "js" in our root folder. Can we already create a web folder, and put them there? It's more close to what Symfony does for assets

We could, but in Symfony the `web`-folder is the document root, so if we add a `web`-folder our urls won't match the urls that will be used in Symfony, so I don't see any reason to do this now.

> I was wondering why do we use jquery-migrate? And should we start using jquery 3.0?
I just migrated the used libraries, so I didn't make any changes to which libraries we use, or don't use.

> Typeahead is only used in the search Module afaik? Or do we use it for tags etc? Maybe remove it from the core and keep it in the module folder-only
See above

> In the future, we can maybe let Travis upload the artifact/compiled forkcms to Github releases (http://docs.travis-ci.com/user/deployment/releases/) ? Or how do other github projects use compiled binaries for their releases
Not relevant for this PR, maybe create an issue?

src/Backend/Core/Layout/Sass/screen.scss
Katrienvh: and can't we move the import of font awesome after the imports of the bootstrap files ?

tijsverkoyen: Uhu, but in that case the name "_bootstrap-variables.scss" isn't correct anymore

tijsverkoyen: Maybe we should wait with this for now, as it is not related to the PR, and create a separate PR for this.

Katrienvh: True

PR #1921, opened 2017-01-03
Add Composer merge plugin

jessedobbelaere:
## Type

- Feature

When you install a Fork CMS module that uses some external php dependencies/libraries, we either have to include the external php files in the module folder (oldschool), or add instructions in the Readme file to let the user do `composer require vendor/package` in the root forkcms directory, before he can use the module in his project. This could be improved...

## Pull request description

What if we add a composer.json file in our src/Backend/Modules/modulename folder that has the external dependencies. We need our main composer.json to parse this and install those dependencies. Thanks to https://github.com/wikimedia/composer-merge-plugin this is possible. 

To test this:

* Use this branch
* Do a `composer install` to install wikimedia/composer-merge-plugin
* Create e.g. a `src/Backend/Modules/Blog/composer.json` file with the following example packages:

```
{
    "require": {
        "tinify/tinify": "^1.3",
        "aws/aws-sdk-php": "3.20.12"
    }
}
```

* Do a `composer update` in  your forkcms root
* The composer-merge-plugin will find your module composer.json file and merge the contents into the lockfile and install the dependencies into the vendor/ folder. You can use your module! 

Note that `composer update` will also update your other dependencies minor versions (normally). If we already have a lockfile in our project, then wikimedia/composer-merge-plugin won't do anything I noticed when doing a normal composer install. Only if you do composer update, it will update the lockfile with the merge of the new dependencies from the module's composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
         "simple-bus/doctrine-orm-bridge": "^4.0",
         "mailmotor/mailchimp-bundle": "^2.0",
         "mailmotor/campaignmonitor-bundle": "^1.0",
-        "mailmotor/mailmotor-bundle": "^2.0"
+        "mailmotor/mailmotor-bundle": "^2.0",
+        "wikimedia/composer-merge-plugin": "^1.3"
     },
     "require-dev": {
         "jdorn/sql-formatter": "1.2.17",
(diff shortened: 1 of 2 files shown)


freshface:
Tried it out and works perfect!

StijnVrolijk:
We will implement a different way of doing this when the PSR-4 PR is merged.

PR #1923, opened 2017-01-03
Require php 7

carakas:
## Type

- Enhancement

## Pull request description

Since php 5.6 only gets security updates it is the perfect time to require php 7 when fork 5 is released

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: php
 
 matrix:
   include:
-    - php: 5.5
-    - php: 5.6
     - php: 7.0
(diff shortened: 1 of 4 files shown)


jessedobbelaere:
Hurray!

But let's hope most shared hosting offers php7 by default nowadays. I can imagine the amount of slack questions coming from people new to fork cms, asking why it doesn't want to install. Or we should create a list of "recommended hostings".

carakas:
It's something long term, this MR, but I already created it so we don't forget it

PR #1926, opened 2017-01-07
Fix stylesheetparser plugin + upgrade ckeditor 4.5 > 4.6 + improvements

jessedobbelaere:
## Type

- Critical bugfix

In CKEditor, the styles dropdown contains a lot of bootstrap rubbish and none of our frontend styles that we push to `contentsCss` in backend.js

## Pull request description
* Upgraded CKEditor from 4.5 to 4.6: http://ckeditor.com/blog/CKEditor-4.6-released
* Removed the devtools plugin
* Removed the div area plugin (http://ckeditor.com/addon/divarea). This was the cause of all problems! This plugin removes the iframe and turns it into a div area. But our stylesheetparser plugin takes a look at all the loaded stylesheets in the document and creates a list of css classes. In an iframe, the ckeditor iframe document only has the stylesheets we push (screen.css, ...), but when it's a div the stylesheetparser plugin takes a look at all the stylesheets in the document, so also the bootstrap stylesheets etc. ==> We get a lot of bootstrap rubbish styles and no frontend css files are loaded.


jessedobbelaere:
Can I also remove some of these plugins as they're not used in fork cms?

* Ajax Data Loading: provides additional api's to work with Ajax/XML (do we use this?)
* Document Properties: manipulate the metadata of a HTML Document when in full-page mode (??)
* Extra form attributes: lets you specify a class & id in forms
* Notification Aggregator: allows you to aggregate multiple tasks into a single notification (??)
* Placeholder: lets you create and edit placeholders (non-editable text fragments).
* UI Color Picker: provides a color picker dialog to select and preview the editor user interface color.

carakas:
@jessedobbelaere will you still check the ACF? http://docs.ckeditor.com/#!/guide/dev_acf

jessedobbelaere:
yes, hopefully this weekend 🤗

jessedobbelaere:
I looked into this but 1) cannot get ACF working. Config seems to be ignored.
Anyway we would need `pasteFilter` as we should only filter on paste (http://docs.ckeditor.com/#!/api/CKEDITOR.config-cfg-pasteFilter). If I want to change the html code using the Source button and add a div or inline style or anything, it shouldn't be stripped away.

`pasteFilter: 'plain-text',`

But when I paste something with headers, images and comic sans ms (inline style) copied from Word, then it keeps the formatting so doesnt seem to be working either.

I would just merge with the current `pasteFromWordRemoveFontStyles: true,` and look at this later. As this ticket solves some other issues like the styles dropdown not working

carakas:
Ok

PR #1928, opened 2017-01-09
Transform numeric to number because it is stored in the database that way

mathiashelin:
## Type

- Enhancement

## Pull request description

In js, we'd like to use type 'number' as it is more correct than 'numeric'. We need to change this in formbuilder, because in the database the fields are stored as type 'numeric'.

### DIFF
--- a/src/Frontend/Modules/FormBuilder/Widgets/Form.php
+++ b/src/Frontend/Modules/FormBuilder/Widgets/Form.php
@@ -321,6 +321,7 @@ class Form extends FrontendBaseWidget
     private function setCustomHTML5ErrorMessages(array $item, SpoonFormAttributes $formField)
     {
         foreach ($item['validations'] as $validation) {
+            // @deprecated
             // we need to change it here since it is saved like this in the database and we can't change that for now.
             if ($validation['type'] === 'numeric') {
                 $validation['type'] = 'number';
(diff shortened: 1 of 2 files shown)


StijnVrolijk:
Why is 'number' more correct than 'numeric'?

mathiashelin:
Because type="numeric" doesn't exist in html.

src/Frontend/Modules/FormBuilder/Widgets/Form.php
```diff
@@ -322,6 +321,11 @@ private function loadForm()
     private function setCustomHTML5ErrorMessages(array $item, SpoonFormAttributes $formField)
     {
         foreach ($item['validations'] as $validation) {
+            // we need to change it here since it is saved like this in the database and we can't change that for now.
+            if ($validation['type'] === 'numeric') {
```
tijsverkoyen: Wouldn't it be more correct to store the correct value?

carakas: you can change that in the backend but it needs to be implemented in a backwards compatible way.

src/Frontend/Modules/FormBuilder/Widgets/Form.php
```diff
@@ -322,6 +321,11 @@ private function loadForm()
     private function setCustomHTML5ErrorMessages(array $item, SpoonFormAttributes $formField)
     {
         foreach ($item['validations'] as $validation) {
+            // we need to change it here since it is saved like this in the database and we can't change that for now.
```
carakas: I would maybe add a @TODO here so we can find this when we prepare for fork 5

src/Frontend/Modules/FormBuilder/Widgets/Form.php
tijsverkoyen: Well, it should.
Maybe we should also add a `@deprecated`-line, so this code can be removed.

Issue #1520, opened 2016-06-10
underscore variable names assigned to twig and created in twig.


WouterSioen:
The twig coding standards state this:

> Use lower cased and underscored variable names:

source: http://twig.sensiolabs.org/doc/coding_standards.html

PR #1930, opened 2017-01-10
[WIP] Bootstrap per default

tijsverkoyen:
## Type

- Enhancement
- Feature

## Pull request description

Moving to a frontend that will use Bootstrap per default

All templates are revised and have:

* consistent markup.
* correct usage of the `<section>`, `<nav>`, ... tags.
* all widgets and blocks have a class: `module-$module-name`.
* all widgets have a class to identify the widget: `widget-$module-name-$widget-name`.
* all blocks have a class to identify the block: `block-$module-name-$widget-name`.
* in the templates Twig-blocks are added, see below.
* removed useless classes and id's

# Themes

Your themes can be much simpler from now on. 
Since we started using the Twig-block, you can extend these templates.

The default markup for the TagCloud-widget is:

    {% if widgetTagsTagCloud %}
      <aside class="module-tags widget-tags-tag-cloud">
        {% block widgetHeading %}
          <header class="widget-heading" role="banner">
            <h3>{{ 'lbl.Tags'|trans|ucfirst }}</h3>
          </header>
        {% endblock %}
    
        {% block widgetBody %}
          <div class="widget-body">
            <p>
              {% for tag in widgetTagsTagCloud %}
                <a href="{{ tag.url }}" rel="tag">
                  {{ tag.name }}
                </a>
                <span class="badge hidden-phone">{{ tag.number }}</span>{% if not loop.last %},{% endif %}
              {% endfor %}
            </p>
          </div>
        {% endblock %}
      </aside>
    {% endif %}

As you can see there is a `widgetHeading`-block. So if your theme doesn't need a header, your code in your theme would be:

    {% extends "@ForkFrontendModules/Tags/Layout/Widgets/TagCloud.html.twig" %}
    
    {% block widgetHeading %}
    {% endblock %}

As you can see we are extending the default-template, and overrule the `widgetHeading`-block.

## Questions

Below are some questions, which I think should be decided by the community.

1. When "no theme" is selected, should we load the Bootstrap-css and js? Or should "no theme" be a theme where no styles are loaded?
2. In the location-module you can set the width of the map, is this relevant? Shouldn't this be 100% and be limited by the wrapping container?

## Todo

* [x] Review all the templates and make them consistent:
    * [x] Use `page-header` when relevant.
    * [x] Remove useless classes and id's.
    * [x] Categories and tags should be parsed in the same way across modules.
    * [x] Use correct `section`, `article`, `aside`, `nav`-tags in the default templates.
    * [x] Add the correct `role`-attributes.
* [x] Review the mail templates
* [x] Rewrite module-specific css in sass, and generate the css-files with `gulp`.
* [x] Add a `gulp serve`-task for the frontend.
* [x] Check the classes that are used inside JS and prefix them with `js-`.
* [x] Check all occurrences of `|capitalize` and check if they should be replaced with `|ucfirst`.
* [x] Check all loops in the templates and check if they use `loop.first|last` instead of `varname.last|first`.
* [x] Add a ton of blocks in each template, so the "extending" templates can contain only the relevant changes.
* [x] Remove triton + make a bootstrap basic theme


carakas:
- I would leave no-theme in place, that way there is the option to create a site without bootstrap if one wishes to do so
- to your second question about the map width I think you have a point but I'll leave that up to the frontenders

tijsverkoyen:
@carakas Re 1: There is no such thing as "without Bootstrap" as the markup is already Bootstrap

carakas:
wouldn't it be better then to have the module templates contain dumps of the available data, and that the actual template is in the theme? But that might be confusing, so in that case I might opt-in to let the no-theme be bootstrap

tijsverkoyen:
dumps, really? In that case it renders the default-templates useless... In my opinion the default templates should contain the most correct markup, and the themes should extend/overwrite only the parts that should be changed.
If we use dumps, each theme has to include *all* templates for each module *ever*.

tijsverkoyen:
At SumoCoders a lively discussion has emerged when we talked about the `js-` prefix for the classes/id's used by JavaScript.

In our opinion there are 2 options:

1. Prefix all the classes/id's that are used by JavaScript with `js-` so the person who styles everything knows he should leave these classes.

2. Using data-attributes instead of classes/id's to bind to. Eg:

```html
<button
  data-role="amount-button" 
  data-direction="decrease" 
  class="btn btn-default" 
  type="button"
>
  -
</button>
<input type="text" class="form-control" value="1" data-role="order-amount">
<button 
  data-role="amount-button" 
  data-direction="increase" 
  class="btn btn-default" 
  type="button"
>
  +
</button>
```

In JS you bind on `$('[data-role=amount-button]')`.

ciryk:
I would keep the no-theme, but set the bootstrap theme as default to not scare non-developers away. For the Location module i've put the width to 100% several times, but this might not always be the case. Best way in my opinion would be able to put percentages/px in the location module in the backend instead of just px now. I will check the code this evening.

matthiasmullie:
@tijsverkoyen An argument against data attributes in this case: performance.

Searching for ids/classes has always been natively supported by browsers. Searching by attribute (or attribute/value) has not been. IIRC, jQuery used to just iterate elements based on element/id/class, and then drop elements that did not fit the rest of the criteria (e.g. some data attribute in the selector). It was slower.

Browsers now natively support this stuff with querySelectorAll, but it's still less efficient. I haven't looked at the browsers' source code for this, but have ran a quick test. Run this on the google.com homepage (which has a very small number of nodes!):

```js
var time = (new Date()).getTime(), i;
for (i = 0; i < 100000; i++) {
	document.querySelectorAll('[role="search"]');
}
console.log((new Date()).getTime() - time);
```
This takes over a second on my machine.

Then run it with `document.querySelectorAll('.tsf');`. This takes about 300ms on my machine
Then try `document.querySelectorAll('#tsf');`. That's about 50ms.

All of these select the exact same node, but the attribute-based one was *a lot* slower.

If you're going to do have a lot of selectors this way, or have a document with a lot of nodes, I'd suggest to steer clear of using (only) attributes here.

StijnVrolijk:
@matthiasmullie I'm getting very different (read: less mind-blowing) results. I made a JSFiddle testing the same thing you did but using jQuery (which we would use in a real life situation) and found the results not to be as different as in your tests.

https://jsfiddle.net/97tc2dyL/
```
attribute: 446
class: 317
id: 45
```

As you can see there is still a difference between all of them but the difference is a lot smaller. So small in fact that using data attributes everywhere in a real life application (in Fork) in comparison to classes would result in a process time difference lower than 1ms. To me, that's completely negligible and I will say DX > performance in this case.

matthiasmullie:
I used `querySelectorAll` because that's the best case you can hope for: native browser support. For jQuery to work with certain older browsers, they have to work around the lack of `querySelectorAll` and the difference will be even larger.

A more realworld example: select all the ".appListFull" nodes on http://www.fork-cms.com/extensions/apps.
That's not even a complex/heavy page, but it's perhaps a better example than the few optimized nodes on google.com.

```php
$('.appListFull').attr('some', 'test');

var time = (new Date()).getTime(), i;
for (i = 0; i < 1000; i++) {
	$('[some="test"]');
}
console.log('attribute: ' + ((new Date()).getTime() - time));

time = (new Date()).getTime(), i;
for (i = 0; i < 1000; i++) {
	$('.appListFull');
}
console.log('class: ' + ((new Date()).getTime() - time));
```

I've only ran it a 1000 times now, a very reasonable amount of times. The result is around 40ms vs 20ms just about every time I run it.

I just grepped the ForkCMS source code; there are 2707 (`$(`) - 387 (`$(this)`) + 1395 (`.find`) selectors used in ForkCMS (there's probably even a few more, but I think the use of stuff like `.siblings(selector)` is negligible. So that's around 3000 selectors already being used. And I'm sure a bunch of them are used in a loop, so the final result is likely going to be a lot higher than 3000.
Those are, of course, not all run on the same page (although probably most), but it goes to show that 1000 selectors is not at all an exaggerated number. Even there, on a pretty decent machine, I lose 20ms. On cheap/old/3rd world phones, this is going to be even more.

You can still argue that 20ms is not worth ruining dev experience (although I personally don't think either of the ideas make any difference in dev experience), but the difference is not negligible or <1ms.

src/Frontend/Modules/Search/Layout/Sass/Search.scss
```diff
@@ -0,0 +1,206 @@
...
+  &:focus {
+    border-color: darken($border-color, 10%);
+    box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075), 0 0 6px lighten($border-color, 20%);
+  }
+}
+
+// coloring based on state
+.has-warning .twitter-typeahead {
+  .tt-input, .tt-hint {
+    @include typeahead-form-control($state-warning-text);
```
Katrienvh: Why you use the state text color and not the state border color?

src/Frontend/Modules/Search/Layout/Sass/Search.scss
```diff
@@ -0,0 +1,206 @@
...
+  }
+}
+
+.has-error .twitter-typeahead {
+  .tt-input, .tt-hint {
+    @include typeahead-form-control($state-danger-text);
```
Katrienvh: Why you use the state text color and not the state border color?

src/Frontend/Modules/Search/Layout/Sass/Search.scss
```diff
@@ -0,0 +1,206 @@
...
+  }
+}
+
+.has-success .twitter-typeahead {
+  .tt-input, .tt-hint {
+    @include typeahead-form-control($state-success-text);
```
Katrienvh: Why you use the state text color and not the state border color?

tijsverkoyen:
We had a discussion on [Slack](https://forkcms.slack.com/archives/general/p1484146008000251). The final outcome is we will use `data-role` for the selectors, more details will be provided in the commit-message.

tijsverkoyen:
Note to self check #1520

src/Frontend/Modules/Search/Layout/Sass/Search.scss
tijsverkoyen: I used the same variables as the less-version. And I think they use these vars, because the focus-state calculates it value based on this variable

src/Frontend/Modules/Search/Layout/Sass/Search.scss
tijsverkoyen: I used the same variables as the less-version. And I think they use these vars, because the focus-state calculates it value based on this variable

src/Frontend/Modules/Search/Layout/Sass/Search.scss
tijsverkoyen: I used the same variables as the less-version. And I think they use these vars, because the focus-state calculates it value based on this variable

jeroendesloovere:
@tijsverkoyen Keep up the good work 👍

carakas:
@tijsverkoyen you will have to rebase from 5.0.0-dev to fix the tests because of changes on travis

src/Backend/Modules/Location/Installer/Installer.php
```diff
@@ -63,6 +63,6 @@ public function install()
         $this->setNavigation($navigationModulesId, 'Location', 'location/settings');
 
         // add extra's
-        $this->insertExtra('Location', ModuleExtraType::widget(), 'Location', null, 'a:1:{s:3:"url";s:34:"/private/location/index?token=true";}', 'N');
```
carakas: there is also a widget, so we need both

src/Backend/Core/Installer/ModuleInstaller.php
```diff
@@ -313,6 +313,11 @@ protected function getTemplateId($template, $theme = null)
             $theme = $this->getSetting('Core', 'theme');
         }
 
+        // if the theme is still null we should fallback to the core
+        if ($theme === null) {
+            $theme = 'core';
```
carakas: @tijsverkoyen is the casing correct here? a bit further you use Core

carakas:
going to merge this, we will fix the failing tests in a later stage

PR #1940, opened 2017-01-14
Remove annotations from documentation in comments

carakas:
When validating the value object this gives an error since it tries to parse them as an annotation

## Type

- Critical bugfix

## Pull request description

When validating the value object this gives an error since it tries to parse them as an annotation


StijnVrolijk:
Wouldn't it be a better idea to just say `PrePersist()`, `PreUpdate()` and so on?
If you type that in a decent editor it'll typehint on it anyways and `ORM` is an alias (although often used in official documentation). I think using `[at] ORM` will just confuse people

carakas:
@StijnVrolijk fixed in e9ffc77e1

PR #1941, opened 2017-01-16
Remove the deprecated code

tijsverkoyen:
## Type

- Enhancement

## Pull request description

Removes all the deprecated code

## Todo

* [x] [`SPOON_*`-constants](https://github.com/forkcms/forkcms/blob/5.0.0-dev/app/Kernel.php#L91) should be removed. But `SPOON_DEBUG` is still used. This should be checked for each constant.
* [x] [`Backend\Core\Engine\Meta->generateURL()`](https://github.com/forkcms/forkcms/blob/5.0.0-dev/src/Backend/Core/Engine/Meta.php#L124)
* [x] [`Backend\Core\Engine\Meta->save()`](https://github.com/forkcms/forkcms/blob/5.0.0-dev/src/Backend/Core/Engine/Meta.php#L425) should be removed, but a decent Upgrade-guide should be added.
* [x] [`Common\Core\Twig\BaseTwigTemplate->startGlobals()`](https://github.com/forkcms/forkcms/blob/5.0.0-dev/src/Common/Core/Twig/BaseTwigTemplate.php#L137) should be reviewed, or at least marked deprecated. 
* [x] [`Common\Core\Twig\Extensions\TwigFilters`](https://github.com/forkcms/forkcms/blob/5.0.0-dev/src/Common/Core/Twig/Extensions/TwigFilters.php#L106) should be reviewed, as it is not clear which alternatives should be used.

PR #1942, opened 2017-01-16
Scripts cleanup

tijsverkoyen:
## Type

- Enhancement
- Feature

## Pull request description

In the tools-folder a lot of scripts exists, some of them are kinda useless, or not something you want in Fork. So I removed the scripts that are not Fork-related:

* `check_code`, I don't think anybody uses this. Also this is something that should be documented instead of a script
* `stats`, generating stats about the code is so 2014, and we have Scrutinizer for this.
* `batch_resize`, this should not be something that has to be maintained by Fork.

I rewrote the "relevant" scripts into a Symfony Console Command.


jeroendesloovere:
Love this PR!

tijsverkoyen:
@carakas those two were marked deprecated, and are removed in #1941

PR #1949, opened 2017-01-19
Improved file and image type for symfony form

carakas:
## Type

- Feature

## Pull request description

Make your life easier and use these file/image types ;)

They do a lot of things automagically for you. See Documentation PR for more info

### DIFF
--- a/src/Backend/Core/Layout/Templates/FormLayout.html.twig
+++ b/src/Backend/Core/Layout/Templates/FormLayout.html.twig
@@ -242,6 +242,28 @@
   {{ form_rest(form) }}
 {%- endblock meta_widget -%}
 
+{% block image_widget %}
+  {% spaceless %}
+    <div class="form-group">
+      {{ form_widget(form.file) }}
+      {% if help_text_label %}
+        <div class="help-text">{{ help_text_label|trans|format(help_text_argument) }}</div>
+      {% endif %}
+      {{ form_errors(form.file) }}
+    </div>
+    {% set show_preview = show_preview and data is not null and data.fileName is not empty %}
+    {% if show_preview or show_remove_image %}
+      <div class="form-group">
+        {% if show_preview and preview_url %}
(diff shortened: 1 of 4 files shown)


src/Backend/Core/Layout/Templates/FormLayout.html.twig
```diff
@@ -242,6 +242,54 @@
   {{ form_rest(form) }}
 {%- endblock meta_widget -%}
 
+{% block image_widget %}
```
jeroendesloovere: Rename to "fork_image_widget"

src/Backend/Form/Type/ImageType.php
```diff
@@ -0,0 +1,279 @@
...
+            ]
+        );
+    }
+
+    /**
+     * @return string
+     */
+    public function getBlockPrefix()
+    {
+        return 'image';
```
jeroendesloovere: Rename to 'fork_image'

src/Backend/Form/Type/ImageType.php
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /**
+     * @return string
+     */
+    public function getParent()
+    {
+        if (!$this instanceof self) {
+            return 'image';
```
jeroendesloovere: Don't know if this should be 'fork_image' instead, but the line below is 'file', maby that one should be 'fork_file' also?

src/Backend/Form/Type/ImageType.php
carakas: fixed in 290c35926

src/Backend/Form/Type/ImageType.php
carakas: fixed in d68ce22d7
and no :) it has to be file because we don't want the custom stuff of fork_file to interfere with fork_image :) file is the default of symfony, there is non for image that's why it originally wasn't prefixed with fork

PR #1950, opened 2017-01-19
Feature mailmotor one page subscribe widget

cirykpopeye:
## Type

- Critical bugfix
- Feature

- When a faulty list id, or no users in a list it was throwing a foreach error, solved by checking with is_array()

## Pull request description

- One page submission on MailChimp implemented in the widget, instead of linking to the action.


src/Frontend/Modules/Mailmotor/Layout/Widgets/Subscribe.html.twig
```diff
@@ -4,16 +4,15 @@
       <h3 class="title">{{ 'lbl.SubscribeToNewsletter'|trans|capitalize }}</h3>
     </div>
     <div class="bd">
-      <form action="{{ geturlforblock('Mailmotor', 'Subscribe' ) }}" method="post">
-        <input type="hidden" name="form" value="mailMotorSubscribeForm" />
+      {{ form_start(form) }}
+        {#<input type="hidden" name="form" value="mailMotorSubscribeForm" />#}
```
carakas: just remove it, don't leave code in comments

cirykpopeye: Removed. Forgot it was there, thanks!

src/Frontend/Modules/Mailmotor/Form/SubscribeType.php
```diff
@@ -102,12 +106,14 @@ public function getInterests()
             $mailMotorInterests = $this->subscriber->getInterests();
 
             // Has interests
-            if (!empty($mailMotorInterests)) {
+            if (!empty($mailMotorInterests) && is_array($mailMotorInterests)) {
```
jeroendesloovere: Normally, we don't encourage to do two bugfixes in one PR.

carakas: But if you get errors along the way of fixing your initial bug, it is normal to also fix them

jeroendesloovere:
I'm going to test this now and see if this a good solution

jeroendesloovere:
Just tested this and it's not a required solution.

carakas:
@jeroendesloovere the downside of that is that people have to fill in things on 2 pages, and for something as simple as this that is overkill afaik

cirykpopeye:
Maybe we should have two seperate widgets for that then, or check if there are any interests to select?

jeroendesloovere:
We just added a new bug. Can you look at the screenshot I provided above.
Then "submit" button is above the "interests".
@cirykpopeye can you fix this?

tijsverkoyen:
@jeroendesloovere I'm not a usability-expert. But if you have a widget that states "Subscribe to the newsletter", with only the emailaddress field, that implies you only have to enter you email address. So entering extra details in the "next" step will have less conversion as a result.

src/Frontend/Modules/Mailmotor/Layout/Widgets/Subscribe.html.twig
```diff
@@ -8,6 +8,9 @@
         <p>
           {{ form_row(form.email) }}
         </p>
+
+        {{ form_rest(form) }}
```
carakas: that won't work since it will also take the button with it

cirykpopeye: Yes indeed. Fixed.

src/Frontend/Modules/Mailmotor/Layout/Widgets/Subscribe.html.twig
```diff
@@ -9,7 +9,7 @@
           {{ form_row(form.email) }}
         </p>
 
-        {{ form_rest(form) }}
+        {{ form_row(form.interests) }}
```
carakas: There are cases when the interests aren't added, this will now give an error in those cases.
You need to check if the field exists

src/Frontend/Modules/Mailmotor/Form/SubscribeType.php
carakas: but it would indeed have been better to do it in a separate mr

cirykpopeye: I fixed it as I couldn't continue testing the feature. But indeed.

PR #1951, opened 2017-01-20
930 create a new page under an existing selected one

### LINKED ISSUE #930: Create a new page under an existing selected one
When we "Add page" after selecting an existing page in the tree we would expect the new page to be created a child of the selected item. Currently it creates it as a single page which has to be moved to its desired page.

carakas:
## Type

- Feature

fixes #930 

## Pull request description

screen shot 2017-01-20 at 14 25 15


jeroendesloovere:
Jihaa

Issue #1965, opened 2017-02-01
Redirect bug in parsewidget


jeroendesloovere:
## Type

- Non Critical bug

## Problem description

> When adding a form from the FormBuilder module to a page using `parsewidget`, you can't submit the form. That throws an error

## Steps to reproduce

1. Create a form in formbuilder
2. Open up a template, f.e.: `src/Themes/triton/Core/Layout/Templates/Default.html.twig`
3. Add `{{ parsewidget('FormBuilder', 'Form', 1) }}`
4. Visit the website page, fill in the form and click "submit"
=> BAM, error
<img width="1219" alt="schermafbeelding 2017-02-01 om 10 41 47" src="https://cloud.githubusercontent.com/assets/588616/22501897/0bc3bac6-e86b-11e6-92ae-6cc299b79852.png">

## Expected behavior / Proposed solutions

> Fix for this nasty bug

carakas:
changed to non critical, since parsewidget is a hack and should be avoided when possible

PR #1966, opened 2017-02-01
Fix in parseWidget when a widget throws a RedirectException

jeroendesloovere:
Fixes https://github.com/forkcms/forkcms/issues/1965

Solves the Exception being thrown. The redirect will work.
But... it first shows the same page, with a redirect text... (see screenshot 2)

schermafbeelding 2017-02-01 om 11 01 48
schermafbeelding 2017-02-01 om 11 02 09
schermafbeelding 2017-02-01 om 11 01 51

### DIFF
--- a/src/Frontend/Core/Engine/TemplateModifiers.php
+++ b/src/Frontend/Core/Engine/TemplateModifiers.php
@@ -9,6 +9,7 @@ namespace Frontend\Core\Engine;
  * file that was distributed with this source code.
  */
 
+use Common\Exception\RedirectException;
 use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Core\Engine\Block\Widget as FrontendBlockWidget;
 use Frontend\Core\Language\Locale;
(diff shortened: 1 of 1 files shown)


src/Frontend/Core/Engine/TemplateModifiers.php
```diff
@@ -406,6 +407,8 @@ public static function parseWidget($module, $action, $id = null)
             FrontendModel::getContainer()->set('parseWidget', null);
 
             return $content;
+        } catch (RedirectException $e) {
+            return $e->getResponse();
```
carakas: does this work?
Can you describe what it does now?

jeroendesloovere: Yes, it works. But not perfectly (since there is probably not better fix?).
Check this page above to find out what it does...

carakas:
Ti

src/Frontend/Core/Engine/TemplateModifiers.php
carakas: I'm sorry, but I'm not merging this.
You will need to find another solution

carakas: I'm sorry, but I'm not merging this.
You will need to find another solution

jeroendesloovere:
Ok, can somebody else look for a solution. Don't have time to search for a better solution.

carakas:
in that case i'll close this one

PR #1969, opened 2017-02-02
Datagrid macro

StijnVrolijk:
## Type

- Enhancement
- Feature

## Pull request description

This is just a macro for a simple dataGrid. It takes the dataGrid itself and an optional 'noItems' message parameter. Not entirely useful in the core modules because of all the custom stuff but when creating your own modules with a simple CRUD, you don't want to copy/paste the dataGrid code every time, do you?

### DIFF
--- a/src/Backend/Core/Layout/Templates/macros.html.twig
+++ b/src/Backend/Core/Layout/Templates/macros.html.twig
@@ -8,3 +8,23 @@
 {% macro required() %}
   <abbr data-toggle="tooltip" title="{{ 'lbl.RequiredField'|trans|ucfirst }}">*</abbr>
 {% endmacro %}
+
+{% macro dataGrid(dataGrid, noItemsMessage) %}
+  <div class="row">
+    <div class="col-md-12">
+      {% if dataGrid %}
+        <div class="table-responsive">
+          {{ dataGrid|raw }}
+        </div>
+      {% else %}
+        <p>
+          {% if noItemsMessage %}
+            {{ noItemsMessage }}
+          {% else %}
+            {{ 'msg.NoItems'|trans|raw }}
+          {% endif %}
+        </p>
+      {% endif %}
+    </div>
+  </div>
+{% endmacro %}
--- a/src/Backend/Modules/Blog/Layout/Templates/Index.html.twig
+++ b/src/Backend/Modules/Blog/Layout/Templates/Index.html.twig
@@ -43,53 +43,31 @@
   {% endif %}
 
   {% if dgRecent %}
-    <div class="row">
-      <div class="col-md-12">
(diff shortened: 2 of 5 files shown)


src/Backend/Core/Layout/Templates/macros.html.twig
```diff
@@ -8,3 +8,23 @@
 {% macro required() %}
   <abbr data-toggle="tooltip" title="{{ 'lbl.RequiredField'|trans|ucfirst }}">*</abbr>
 {% endmacro %}
+
+{% macro dataGrid(dataGrid, noItemsMessage) %}
+  <div class="row">
+    <div class="col-md-12">
```
carakas: what is the point of first doing row and then col-md-12?
I think you can leave both out

src/Backend/Modules/Blog/Layout/Templates/Index.html.twig
```diff
@@ -43,53 +43,31 @@
...
-          </div>
-          <div class="table-responsive">
-            {{ dgRecent|raw }}
-          </div>
-        </div>
+    <div class="content-block content-block-pb">
+      <div class="content-title">
+        <p>{{ 'lbl.RecentlyEdited'|trans|ucfirst }}</p>
       </div>
+      {{ macro.dataGrid(dgRecent, 'msg.NoItems'|trans|format(geturl('add')) }}
```
carakas: add action should be capitalised, also in all the following parts

src/Backend/Core/Layout/Templates/macros.html.twig
StijnVrolijk: nope, when used in an empty content block, the structure will be `.container>.row>.col-md-12` which is bootstrap's standard way of working. If you use this macro in some other markup   leaving the row and col out might cause some layout issues

carakas: fair point, hadn't thought of that :)

src/Backend/Modules/Blog/Layout/Templates/Index.html.twig
StijnVrolijk: well there isn't a single place in the backend which uses capitalised parameters in the `geturl()` modifier. This line of code in the BackendModel::createURLForAction() will convert everything anyways:

```
$action = mb_strtolower(preg_replace('/([a-z])([A-Z])/', '$1_$2', $action));
```

`add` stays `add`
`Add` becomes `add`
`AddWhatever` becomes `add_whatever`
and so on..

carakas: ok

jeroendesloovere:
Nice one!

PR #1976, opened 2017-02-08
Profiles/Forgot password: show message if code is invalid

wouterverstuyf:
## Type

- Enhancement

> When requesting a new password, sometimes people want to use the same link twice.
> A second time clicking on the link results in a 404 error.

## Pull request description

> Maybe it's better to show a message, that tells the user the code is invalid or already used.

### DIFF
--- a/src/Frontend/Modules/Profiles/Actions/ResetPassword.php
+++ b/src/Frontend/Modules/Profiles/Actions/ResetPassword.php
@@ -55,7 +55,6 @@ class ResetPassword extends FrontendBaseBlock
 
                 // validate
                 $this->validateForm();
-
             } elseif ($this->URL->getParameter('sent') != 'true') {
                 // redirect
                 $this->redirect(


StijnVrolijk:
I personally think this is a bad idea.
Displaying a message saying the code is incorrect encourages people to write scripts to test other codes. As soon as they find one they can reset a person's password.

I think if this is ever implemented, it should contain some form of brute force protection

carakas:
I agree with @StijnVrolijk 

apart from that I'm also not really a fan of the if else party

I'm not really a fan of all the if else going on here

```php
if ($this->URL->getParameter('sent') !== 'true') {
    $this->frm->parse($this->tpl);
    return;
}

// hide form
$this->tpl->assign('resetPasswordHideForm', true);

if ($this->URL->getParameter('code') !== 'invalid') {
    $this->tpl->assign('resetPasswordSuccess', true);

    return;
}

$this->tpl->assign('resetPasswordCodeInvalid', true);
```

is way cleaner IMHO

StijnVrolijk:
@carakas to be fair it matches Fork's code styles in that file... :p

wouterverstuyf:
@StijnVrolijk 
I agree...
Is it more secure redirecting to another page where the message is displayed (instead of the 404-page)? Therefore you have an extra page off course... 
A 404 is fine for me, but a customer noted it... Maybe users that click double on the link should not get more than a 404 :)

StijnVrolijk:
Come to think of it.. if you know the URL of the code checking page, you could just do `curl -i http://www.fork-cms.com/en/reset-password/e8r834bc32948d` for example.. if you get a 301 (or 302 i don't know which exactly it is) you know the code is wrong, if you get a 200 you've hit the jackpot..

I think I've just uncovered a security flaw... thanks @wouterverstuyf 
Using your solution you always get a 200, so the `curl` bruteforcing will be a bit harder..

@carakas what do you reckon we do?

carakas:
rewrite the profiles/user module with the multiuserbundle :p

carakas:
the 404 also only comes after they already used the link to reset the password.
you can open the page twice, the first one that is submitted will pass the other will give an error in the current implementation

We can make bruteforce less likely by limiting the time a password reset string is valid and add the same bruteforce protection as we have on the login form of the backend

carakas:
@StijnVrolijk about the code style in the file https://speakerdeck.com/donis/make-legacy-great-again-at-php-benelux-2017?slide=25

wouterverstuyf:
@StijnVrolijk 
Ok, got it... so my less secure solution is actually more secure? Win win, right ;)
The counter like @carakas mentioned would be perfect I think, it's more secure and we can still show a message if a code is invalid, for example 5 times...

carakas:
@wouterverstuyf can you look at the login of the backend and add the implementation here as well?

and please do the code refactoring :) cleaning up the code one bit at a time

StijnVrolijk:
@carakas I agree, but you can't expect everyone to be knowledgable about object calisthenics or even be a fan of it :p 

@wouterverstuyf yep, good job 😉 now make it a great job by adding the bruteforce protection 👍 
If you feel like doing it, some examples for bruteforce prevention can be found here:

# of attempts check starts here: https://github.com/forkcms/forkcms/blob/master/src/Backend/Modules/Authentication/Actions/Index.php#L134

rapid fire prevention start here: https://github.com/forkcms/forkcms/blob/master/src/Frontend/Modules/Blog/Actions/Detail.php#L305

Just be sure to comply to @carakas 's coding standards :p here is an article to get you started on object calisthenics: http://williamdurand.fr/2013/06/03/object-calisthenics/
It's a java (i think) article but it's easily translated to PHP. If you have some specific questions, you know where to find us ;)

carakas:
I'm happy your code looks better than your markdown @StijnVrolijk

StijnVrolijk:
Oh, and please don't use Spoon stuff, that stuff gives me the creeps

wouterverstuyf:
Hmz, I think I can give it a try... I will try doing that the following days, but give me some time. Afterwords you guys can clean up my mess ;)

carakas:
just to make it clear, stijn edited my last comment :p

carakas:
I'm currently working on a reimplementation of the profiles module based on symfony. It uses expiring tokens :)

carakas:
I'll try to finish it for fork 5

WouterSioen:
@carakas are you using this for your rework of the profiles module or can this PR be deleted?

carakas:
@WouterSioen I'm not using this one

PR #1979, opened 2017-02-13
Redirect all traffic to https

jeroendesloovere:
## Type

- Enhancement

## Pull request description

- "all traffic" now gets redirected to https 
- You don't even need to add a website url, just uncomment and it works.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -36,7 +36,7 @@ FileETag MTime Size
 
 	# redirect all trafic to https
 	# RewriteCond %{HTTPS} !=on
-    # RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
+	# RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
 
 	# one url to rule them all
 	# RewriteCond %{HTTP_HOST} !^www\.<domain>\.be [NC]
--- a/.htaccess
+++ b/.htaccess
@@ -35,6 +35,7 @@ FileETag MTime Size
 	RewriteRule src/Backend/Cache/Locale/.*\.php - [F]
 
 	# redirect all trafic to https
+	# Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
 	# RewriteCond %{HTTPS} != on
 	# RewriteCond %{HTTP_HOST} !.*\.dev [NC]
 	# RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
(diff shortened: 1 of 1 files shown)


carakas:
@jeroendesloovere the code redirects all trafic on port 80 to http, so your code offers no improvements

jeroendesloovere:
My code is an improvement in two ways:
- "all traffic" now gets redirected to https 
- You don't even need to add a website url, just uncomment and it works.

carakas:
then why didn't you clarify that in the description :)
also you removed the code that prevents the code from working on .dev

.htaccess
```diff
@@ -35,9 +35,9 @@ FileETag MTime Size
 	RewriteRule src/Backend/Cache/Locale/.*\.php - [F]
 
 	# redirect all trafic to https
-	# RewriteCond %{SERVER_PORT} 80
+	# RewriteCond %{HTTPS} != on
```
jessedobbelaere: Can be written like `RewriteCond %{HTTPS} off`

.htaccess
```diff
@@ -35,9 +35,9 @@ FileETag MTime Size
...
 	# RewriteCond %{HTTP_HOST} !.*\.dev [NC]
-	# RewriteRule ^(.*)$ https://www.<domain>.be/$1 [R=301,L]
+	# RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
```
jessedobbelaere: Add this recommended tweak  for security reasons:

```
# Enable HSTS
Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
```

jessedobbelaere:
Added tiny remark based on Bramus his post on https://www.bram.us/2014/12/06/migrating-your-wordpress-website-from-http-to-https/

.htaccess
carakas: @jessedobbelaere is that something we can enable by default or should it only be enabled with https

jessedobbelaere: Better comment it out too. And enable it when enabling https

> **The only secure way to redirect http to https is to use HSTS (Header Strict-Transport-Security) .** The apache redirect is insecure because an attacker can intercept it and rewrite it. Unfortunately, for older browser and browser how didn't preload HSTS, it's your only option. HSTS is the only reliable protection against SSLTrip
> It tells the browsers to only use https on that domain and all sub-domains for one year (all http request will be rewrite as https request without network interaction)

> HTTP Strict Transport Security (HSTS) is an opt-in security enhancement that is specified by a web application through the use of a special response header. Once a supported browser receives this header that browser will prevent any communications from being sent over HTTP to the specified domain and will instead send all communications over HTTPS. It also prevents HTTPS click through prompts on browsers.
> https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet

.htaccess
tijsverkoyen: This does not work on all servers

PR #1982, opened 2017-02-17
composer.lock updated

jeroendesloovere:


carakas:
I would wait with merging since we discovered another bug today

PR #1983, opened 2017-02-20
Don't use the deprecated Language class

StijnVrolijk:
Use the relocated Language class instead

## Type

- Critical bugfix


src/Frontend/Core/Engine/TemplateModifiers.php
```diff
@@ -11,6 +11,7 @@
 
 use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Core\Engine\Block\Widget as FrontendBlockWidget;
+use Frontend\Core\Language\Language;
```
carakas: doesn't this give conflcts because the class name is already taken in this namespace?

StijnVrolijk: I think that's only an issue if you explicitly use that class. If it's implicitly used (like in this case) it's no problem. I'm using this code in production right now :)

StijnVrolijk:
Rebase master

PR #1990, opened 2017-02-25
frontend.js should also be minified

freshface:
## Type

- Enhancement

Minifies frontend.js


carakas:
@tijsverkoyen a long time ago you changed it in https://github.com/freshface/forkcms/commit/31dc7e85e4d1a2cfbcfd8d71501fc126a3f3c543

probably too long to remember why?

tijsverkoyen:
I think because at that time locale was parsed into the file. But at this moment we use `jsFrontend.locale.loc`

carakas:
ok, in that case it seems like a 👍

PR #1991, opened 2017-02-27
Fix some button issues

Katrienvh:
## Type
- Non critical bugfix

The margins and paddings on button icons were a mess. Some had padding, some margin, some both.


src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -145,9 +145,9 @@ jsBackend.pages.extras =
                             '<span class="templateTitle">' + title + '</span>' +
                             '<span class="templateDescription">' + description + '</span>' +
                             '<div class="btn-group buttonHolder">' +
-                                '<a href="#" class="btn btn-default btn-xs toggleVisibility"><span class="fa fa-' + (visible ? 'eye' : 'eye-slash') + '"></span></a>' +
-                                '<a href="' + (editLink ? editLink : '#') + '" class="' + (extraId == 0 ? 'showEditor ' : '') + 'btn btn-primary btn-xs' + '"' + (extraId != 0 && editLink ? ' target="_blank"' : '') + (extraId != 0 && editLink ? '' : ' onclick="return false;"') + ((extraId != 0 && editLink) || extraId == 0 ? '' : 'style="display: none;" ') + '><span class="fa fa-pencil"></span></a>' +
-                                '<a href="#" class="deleteBlock btn btn-danger btn-xs"><span class="fa fa-trash-o"></span></a>' +
+                                '<a href="#" class="btn btn-default btn-icon-only btn-xs toggleVisibility"><span class="fa fa-' + (visible ? 'eye' : 'eye-slash') + '"></span></a>' +
```
carakas: are these the only places this is used?

PR #1992, opened 2017-02-28
Nicer installer requirements page.

freshface:
## Type

- Enhancement

Installer with requirements is nicer.

Before:

screen shot 2017-02-28 at 14 35 25

After: 

screen shot 2017-02-28 at 14 35 28

### DIFF
--- a/src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/step1.html.twig
+++ b/src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/step1.html.twig
@@ -24,40 +24,40 @@
             <div id="requirementsInformation">
               <div class="list-group">
                 <div class="list-group-item list-group-item-{{ checker.errors.phpVersion }}">
-                  <h3 class="list-group-item-heading">PHP version</h3>
+                  <h4 class="list-group-item-heading">PHP version</h4>
                   <p class="list-group-item-text">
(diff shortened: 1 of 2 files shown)


src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/step1.html.twig
```diff
@@ -36,59 +36,59 @@
               <h3>PHP Extensions</h3>
               <div class="list-group">
                 <div class="list-group-item list-group-item-{{ checker.errors.extensionCURL }}">
-                  <h4 class="list-group-item-heading">cURL</h4>
+                  <h3 class="list-group-item-heading">cURL</h3>
```
carakas: these are h4's since they fall under the h3 of PHP Extensions

StijnVrolijk: I agree

StijnVrolijk:
I like it, just change the h3's

src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/step1.html.twig
freshface: I like them smaller, but the h3 is styled smaller then the h4 which is not correct.
I will fix that also.

PR #1993, opened 2017-03-02
Add a cache clearing button to the backend

StijnVrolijk:
## Type

- Feature

## Pull request description

Simply adds a new Settings action called 'Tools' which contains a button to clear the cache. Clears the cache upon clicking it... fairly simple stuff but I think it could help with people having cache issues but aren't able to clear it using the console commands.

Note I haven't set any rights in the installer, so by default this will only be visible for the God user. If you want other users to see it, you'll have to set the permissions manually.


jeroendesloovere:
Don't know if the "settings module" is the right place to define the "console command" `forkcms.console.cache.clear`.
This functionality is something "core"...
Isn't there a better place for this?

StijnVrolijk:
@jeroendesloovere Unless we create some kind of 'CoreBundle' I can't find a better location.

The reason I moved it from `app/console` to the Settings module is so I can use the command in Fork's kernel as well instead of the kernel created in `app/console`. If I didn't move it, I'd have to assign the command manually which doesn't sound like the responsibility of an AJAX action.

Also, the settings module is and always will be a core module so it seems like a safe place to put it in my opinion.

carakas:
Maybe it is better to then just add a console.yml to the main config and add it there?

StijnVrolijk:
Also, do you guys reckon I should put the `msg.ClearCache` message in a `.help-block` or just leave it as is? 

Currently:
screen shot 2017-03-02 at 13 44 18

Help-block:
screen shot 2017-03-02 at 13 43 58

carakas:
help block

jeroendesloovere:
Help-block

PR #2005, opened 2017-03-13
Mailmotor double opt in setting

jeroendesloovere:
Fixes #1910

PR #2016, opened 2017-03-22
Revert "Mailmotor double opt in setting"

carakas:
Reverts forkcms/forkcms#2005


carakas:
was merged in wrong branch

PR #2018, opened 2017-03-24
Parse custom meta tags for a blogpost in the frontend

sweet-greg:
## Type

- Non critical bug

## Pull request description

This PR adds support for custom meta tags for the blog module. Seems this was activated in the backend but never parsed in the frontend

### DIFF
--- a/src/Frontend/Modules/Blog/Actions/Detail.php
+++ b/src/Frontend/Modules/Blog/Actions/Detail.php
@@ -208,7 +208,7 @@ class Detail extends FrontendBaseBlock
             $this->record['meta_keywords'],
             ($this->record['meta_keywords_overwrite'] == 'Y')
         );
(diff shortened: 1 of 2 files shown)


src/Frontend/Modules/Blog/Actions/Detail.php
```diff
@@ -208,6 +208,7 @@ private function parse()
             $this->record['meta_keywords'],
             ($this->record['meta_keywords_overwrite'] == 'Y')
         );
+		$this->header->setMetaCustom($this->record['meta_custom']);
```
jeroendesloovere: Seems like something gone wrong here with the indentation. Can you fix this @sweet-greg?

carakas: Ah yes, I missed it, it is using tabs. Nice catch jeroen

PR #2019, opened 2017-03-27
JstChecking

sholakolalekan:
## Type

- Critical bugfix
- Non critical bugfix
- Enhancement
- Feature

> List the hashes of the issues that this pull request resolves if their are issues for it.
> Use the following format: fixes #[issue_number]

## Pull request description

> Describe what your pull request will fix / add / …


carakas:
this would make it harder to install fork out of the box, also you added some text to the readme that wasn't correct

Apart from that I don't see the benefit.

carakas:
Thank you for your pull request but this would make it harder to install fork out of the box, also you added some text to the readme that wasn't correct as fork isn't a fork of fork to work with other engines

Apart from that I don't see the benefit. We are moving away from spoon database to doctrine then this isn't needed at all since doctrine will handle everything

Also, I have no idea what Aria is but I'm rather sure that it isn't as well used as InnoDB or MyIsam

As a closing note, I hope this won't keep you from further contributions to the core

carakas:
also, only noticed just now, you created a PR for the repository of somebody else? Can you elaborate why you did that?

carakas:
@sholakolalekan also, only noticed just now, you created a PR for the repository of somebody else? Can you elaborate why you did that?

jeroendesloovere:
@carakas Ok, we are moving to doctrine, but that doesn't mean we should ignore people using other database software?

carakas:
@jeroendesloovere SpoonDB only supports mysql
Aria is aparently something specific to MariaDB that would lock all mysql users out
The source of this PR is a repository from an other person that is is 7 commits ahead, 3910 commits behind forkcms:master.
And as stated in the commit message you would need to add a my.cnf to get it working, that alone makes it a no go.

If you want to alter fork to work like that be my guest in your own branch. But this is not something that we can do in the official repository imho

carakas:
Also the fact that, if we would want to merge it, there need to happen changes on that branch which is not possible because @sholakolalekan is not the owner of that branch

sholakolalekan:
My apologies I am new on the github hence trying to see how things work. My presumption was that a pull request creates a shortcourt to the project on my dashboard for a later indepth look.

carakas:
Ah no problem
What you are looking for is forking

PR #2023, opened 2017-04-10
Make sure SpoonSession is initialised before starting symfony session

StijnVrolijk:
## Type

- Critical bugfix

Symfony sessions did not work if they where started before the spoon sessions

## Pull request description

In most cases we will only access the symfony sessions after SpoonSession has started. But in the cases where that is not the case the symfony sessions didn't work because of some stuff happening This will fix that issue by making sure that when we start the symfony sessions the SpoonSessions also have started.

### DIFF
--- a/app/Kernel.php
+++ b/app/Kernel.php
@@ -52,6 +52,8 @@ abstract class Kernel extends BaseKernel implements KernelInterface
         // boot if it hasn't booted yet
         $this->boot();
 
+        $this->fixSymfonySessions();
+
         return $this->getHttpKernel()->handle($request, $type, $catch);
     }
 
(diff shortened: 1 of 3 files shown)


app/Kernel.php
```diff
@@ -71,6 +73,21 @@ public function boot()
...
 
     /**
+     * We need to manually initialize SpoonSession for the Symfony sessions to work properly
+     */
+    private function fixSymfonySessions()
+    {
+        SpoonSession::start();
+
+        /** @var \Symfony\Component\HttpFoundation\Session\Session $session */
+        $session = $this->getContainer()->get('session');
```
carakas: I would wrap this method still in a check where you check if the service exists, I think that in cases like a console command it doesnt exist, maybe in those cases also skipp strarting SpoonSession

tijsverkoyen: Also, starting a session, means a cookies, which sucks for caching.

tijsverkoyen: In my opinion, the session should only be started if it is needed.

carakas: @tijsverkoyen this should be fixed in https://github.com/forkcms/forkcms/pull/2023/commits/3513b16b6ca4b30bb38a87bbe4194f98ff73b032

PR #2029, opened 2017-04-25
Replaced : self with : entity-name

jeroendesloovere:
## Type

- Enhancement


carakas:
what is the reason behind this?  if it is scruteniser it won't get merged because it is a bug in scruteniser and they will fix it in the next release or something (already reported it)

jeroendesloovere:
I've noticed that PHPStorm sometimes does try to find the class "self" instead of "MediaFolder", ...
And says, class not exists...

carakas:
I don't think that is reason not to use self

PR #2031, opened 2017-04-27
Bugfix in users

jeroendesloovere:
## Type

- Non critical bugfix

## Summary

We have 
- current_login
- last_login

But f.e. when you are GOD, you want to see the actual last login. The backend shows "last_login", but actually we want to see the utmost last login and that is actually "current_login".

## Pull request description

> For a client, I needed to look for the last login time. So I clicked to "settings" > "users", clicked on that user and I saw the date "26 mei 2016" (in database usersetting = "last_login"). Then I logged in with that user (I had the password) and saw, "last login: 23 januari 2017" (in database usersetting = "current_login"). Then I realised this is a bug. Because I was unable to see the actual last login, but only the "previous login"...

## Solution

To fix any confusion, I just add both dates in the overview...

Dates in screenshot are my testsite, so not from the example above.
schermafbeelding 2017-04-27 om 10 48 19

What I've done:
We now see "current_login" and "last_login" in the same row "last logins"


jeroendesloovere:
WTF are all those merges doing here...

carakas:
very simple, you started on the 5.0.0-dev branch and tried to merge it in master

PR #2037, opened 2017-05-17
Fix saving groups

### LINKED ISSUE #2036: Profiles > Group authentication settings broken
## Type

- Non critical bug

## Problem description

When adding some groups through the profiles module you can limit the access to pages to these groups. This seems to be broken.
When you go to the authentication tab when editing a page, the selected groups don't get saved. Also they are default visible, and should only become visible when you enable the checkbox "Users most be logged..."

## Steps to reproduce

Go to the demo, add some groups, edit a page, and go to the authentication tab. Try to save some settings...

The checkboxes must be saved
The checkboxes should only be visible if the checkbox to limit the page access is active

freshface:
## Type

- Non critical bugfix

Fixes #2036 

## Description

If auth is required this field should be hidden. The logic was not correct. The field will be visible when the checkbox to authenticate the page is ticked.

### DIFF
--- a/src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
+++ b/src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
@@ -261,7 +261,7 @@
                     <label for="authRequired" class="control-label">{% form_field auth_required %} {{ 'msg.AuthRequired'|trans|ucfirst }}</label>
                   </li>
                   {% if authGroups %}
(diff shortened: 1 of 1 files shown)


freshface:
@carakas nope.

Thijzer:
don't know the full context but I'm not sure why you wish to hide instead of not render.
```php
{% if authGroups and not item.data.auth_required %}
                  <div class="js-authentication-groups"
```

carakas:
@Thijzer because there is a checkbox on that page and if you check that checkbox it needs to become visible

PR #2038, opened 2017-05-18
Dockerize Fork CMS

jessedobbelaere:
## Type
- Feature

## Pull request description
I feel like we should add support for Docker to make it easy for people to try out Fork CMS. The time that we installed php and mysql locally on our mac, or in vagrant, or with MAMP is long gone. Having a `docker-compose.yml` file in the root will make it easy for people to spin up an instance of Fork CMS. If you look at other web projects nowadays, most of them have a Dockerfile so you can easily spin it up and try it out and not having to worry that you miss a php extension or setting, ...

The goal would be to keep this docker integration up to date so you can use it for your own local development. 

If your team or yourself have project-specific things to add, you can always extend the docker-compose file like this: 
```
docker-compose -f docker-compose.yml -f docker-compose.redis.yml up -d
```
Or you can add a `docker-compose.override.yml` in the root directory where you can override an environment variable (passwords!) or volumes for the existing services. You just do a regular `docker-compose up -d` and the override gets applied. Read more about docker-compose.override.yml [here](https://docs.docker.com/compose/extends/). Will document this too. 

**Should not be used in production, will be mainly focused on local-development**

## Todo's
- [ ] Create documentation article with basic commands how to work with docker, how to quickly spin up a local environment and how to easily setup your docker container in PHPStorm (configure remote interpreter, add phpunit and xdebug, ...).
- [x] Make sure mysql data is mapped with a volume to the host so database is persisted whenever docker container goes down. This seems to cause issues with docker-machine (https://github.com/docker-library/mysql/issues/99)


Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM php:7.1-apache
+MAINTAINER Fork CMS <info@fork-cms.com>
+
+# Enable Apache mod_rewrite
+RUN a2enmod rewrite
+
+# Install the dependencies and PHP extensions needed
+RUN apt-get update && \
```
tijsverkoyen: I'm not really a fan of these onliners, I prefer separate RUN's per extension.

jessedobbelaere: Each RUN command creates a separate docker cache layer. So doing a separate RUN for each command creates a lot of cache layers (and extra space). 

Also, apt-get update and apt-get install commands should always be combined. 
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get

So not sure if we should split it up?

jessedobbelaere:
Also to mention: @carakas suggests moving a docker integration to it's own repo. There are pro's and cons that we can discuss and come to a conclusion (or in an upcoming core meeting).

Personally I think that the docker-compose file should be in the same repo as the code. That way, changes to the php version or a new extension or update are tightly coupled with the repository code and the version it's on.

WouterSioen:
I would keep it in the same repo. It's just one small file so an extra repo seems too much to me.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+    docker-php-ext-install gd mysqli opcache mcrypt intl mbstring pdo pdo_mysql zip
+
+# Custom php.ini settings
+COPY docker/php/php.ini /usr/local/etc/php/
+
+# Enable Xdebug
+COPY docker/php/xdebug.ini xdebug.ini
+RUN pecl install xdebug-2.5.0 && \
+        docker-php-ext-enable xdebug && \
+        cat xdebug.ini  >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
```
WouterSioen: Shouldn't there be some custom xdebug settings in our php.ini? If I remember correctly, Symfony and twig get over the maximum stack trace size with the default xdebug settings, causing white screens sometimes.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+    php -- --install-dir=/usr/bin/ --filename=composer
+COPY composer.json .
+COPY composer.lock .
+RUN composer install --no-scripts --no-interaction --no-autoloader
+
+# Bundle source code into container. Certain files
+COPY . /var/www/html
+
+# Dump the autoloader and execute post-install scripts.
+RUN composer dump-autoload --optimize && \
```
WouterSioen: Can't you just add `--optimize` to the `composer install`? Dumping the autoloader is done by default when installing.
This would also probably enable you to remobe the post install scripts line.

WouterSioen:
All config files should be added in the .htaccess

WouterSioen:
Don't forget to add this to the documentation. This would also be a nice feature for a short blogpost.

Good job 👏🏻

Dockerfile
tijsverkoyen: I would combine the update and install commands, but I would split them in separate RUNS per package.
At SumoCoders we use: https://gist.github.com/tijsverkoyen/fa096eb123f2b85ca140e25ccc2e98ff

I find it more easy to read and see which dependency there are for each package.

tijsverkoyen:
I won't move the Dockerfile to a separate repo, it should be tied together with the version at that moment.

Dockerfile
jessedobbelaere: Hmm but there's no source code yet at the time I do composer install. I only install vendors which should be fast and cached by a docker layer (they don't change that much).

Then I copy the source code and dump the autoload and run post-install cmd's.

I can simplify it by first copying the source code to the container and then doing a composer install if thats easier? Not sure if the cache layer will add a lot of speed improvements

jessedobbelaere:
Added some fixes based on review feedback. 
Saw a good example in the api-platform repo too: https://github.com/api-platform/api-platform (dockerfile, docker-compose.yml etc)

carakas:
@jessedobbelaere Can you also update the docs and add an entry for this and link to it where needed?

WouterSioen:
:+1:

.htaccess
```diff
@@ -66,6 +66,10 @@ deny from env=stayout
 	RewriteRule src/Frontend/Cache/CompiledTemplates - [F]
 	RewriteRule src/Frontend/Cache/Search - [F]
 	RewriteRule src/Backend/Cache/CompiledTemplates - [F]
+	RewriteRule docker/.* - [F]
+	RewriteRule Dockerfile - [F]
+	RewriteRule \.dockerignore - [F]
+	RewriteRule docker-compose\.yml - [F]
```
carakas: docker-compose.override.yml is still missing here

jessedobbelaere: 👍  fixed. will take up the documentation part too soon and do some more testing

carakas:
@jessedobbelaere any update on this?

jessedobbelaere:
Worked on it a bit more (also because I need it as my local dev setup!). Added some improvements and will add a documentation page soon

Video: http://d.pr/v/c1N9N6 of it working



Btw, the reason to first copy composer.json and lock and do composer install, then copy the entire app sourcecode, and then do composer autoloader is because of layer caching. 

You can see a huge improvement in build time when doing it like this. Imagine you build your docker container. Then you make some code changes and you let the container build again. The first build will do the composer install, copy over the code and then do the dump autoloader. The second time it should not install dependencies again because nothing changed in that part (layer cached). It'll say cached and jump immediately to the copying of the code (that likely changed and invalidates the docker cache) and dumping the autoloader. 

e.g. 

```
Step 12/19 : RUN curl -sS https://getcomposer.org/installer |     php -- --install-dir=/usr/bin/ --filename=composer
 ---> Using cache
 ---> 71b2d85e7ecf
Step 13/19 : WORKDIR /var/www/html
 ---> Using cache
 ---> a07f43bfbdc4
Step 14/19 : COPY composer.json ./
 ---> Using cache
 ---> 68c82ce9b7bc
Step 15/19 : COPY composer.lock ./
 ---> Using cache
 ---> b6b8a1fa4539
Step 16/19 : RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest &&     composer clear-cache
 ---> Using cache
 ---> 4149b96219e2
Step 17/19 : COPY . /var/www/html
 ---> 7ff4e667df55
Removing intermediate container dc1870e7248b
Step 18/19 : RUN composer dump-autoload --optimize --classmap-authoritative --no-dev
 ---> Running in 6dbc39acd6d8
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Generating optimized autoload files
 ---> 277d53e68518
```

You can see that building is superfast as the only thing that is not cached is the copying of the project files (they probably changed and autoloader dump.

Read more: https://sentinelstand.com/article/composer-install-in-dockerfile-without-breaking-cache

jessedobbelaere:
Everything should be working. Only todo left is to continue writing documentation (docker quick-guide, phpstorm setup for docker/phpunit/... How to run commands etc)

carakas:
@jessedobbelaere the 2 classes you couldn't find should be solved in fork 5 since they have now a namespace and composer should handle the loading for you

jessedobbelaere:
Awesome :-) !

jeroendesloovere:
@jessedobbelaere Can you create a PR for `composer test` or add this in your docker setup. So we know immediately docker can work with `composer test`...

carakas:
@jessedobbelaere @jeroendesloovere a separate PR for that please :)

jeroendesloovere:
@jessedobbelaere, how far done is this PR?

jessedobbelaere:
@jeroendesloovere Need to continue to write some docs :p

carakas:
composer test has been added btw @jessedobbelaere   @jeroendesloovere 
Any idea when you can finish up the docs?

docker/db/scripts/00_init_test_db.sh
```diff
@@ -0,0 +1,17 @@
...
+##############################################################
+# Initialize a second database used for integration testing  #
+##############################################################
+MYSQL_DATABASE_TEST="${MYSQL_DATABASE}_test"
+
+echo "Creating database ${MYSQL_DATABASE_TEST}..."
+mysql -uroot -p"${MYSQL_ROOT_PASSWORD}" -e "create database ${MYSQL_DATABASE_TEST};"
+
+echo "Importing fixtures into database ${MYSQL_DATABASE_TEST}..."
+mysql -uroot -p"${MYSQL_ROOT_PASSWORD}" ${MYSQL_DATABASE_TEST} < ../test_db.sql
```
carakas: you'll need to update this path since it has moved

.dockerignore
```diff
@@ -0,0 +1,35 @@
...
+.travis.yml
+build/
+composer.phar
+docker-compose.override.yml
+docker-compose.yml
+docker/db/data/
+var/
+vendor/
+.idea
+app/cache/*
```
carakas: this is now van/cache

.dockerignore
```diff
@@ -0,0 +1,35 @@
...
+app/logs/*
```
carakas: this is now van/logs

.gitignore
```diff
@@ -20,3 +20,7 @@
 /bundles/*
 /css/*.css
 /js/*.js
+
+# Docker
+/data/*
```
carakas: would it be possible to move as much of the docker stuff into var/docker ?

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+            # Make sure to change these if you're planning on using Docker for purposes other than local development!
+            - MYSQL_DATABASE=forkcms
+            - MYSQL_ROOT_PASSWORD=kingtriton
+            - MYSQL_USER=forkcms
+            - MYSQL_PASSWORD=forkcms
+        volumes:
+            # Uncomment the next line if you want to keep mysql in a data container
+            # - db-data:/var/lib/mysql:rw
+            # By default, use a bind-mounted host directory instead. It's harder to accidentally lose all your db data!
+            - ./docker/db/data:/var/lib/mysql:rw
```
carakas: can you change this to var/docker

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+            - ./tools/test_db.sql:/test_db.sql:ro
```
carakas: this is now located in the tests directory

docker/db/scripts/00_init_test_db.sh
jessedobbelaere: This is the local path, from the docker-compose.yml volume, so thats the only place that didnt need to change :)

.gitignore
jessedobbelaere: done 👍

carakas:
There is minimal documentation, more documentation can always be added later

PR #2039, opened 2017-05-22
Bugfixes in 5.0.0 dev

jeroendesloovere:
## Type

- Critical bugfix

## Pull request description

- Fixes locale:import for module
- Fixes AssetCollection by adding a sequence
- "Fork" is the default theme, changed this because it caused errors when creating pages
- Added "orderAssets()" method in AssetCollection
- AssetCollectionTest tests expanded
- Fixed adding page without parentId

### DIFF
--- a/src/Backend/Modules/Pages/Engine/Model.php
+++ b/src/Backend/Modules/Pages/Engine/Model.php
@@ -918,8 +918,12 @@ class Model
         ];
     }
 
-    public static function getURL(string $url, int $id = null, int $parentId = 0, bool $isAction = false): string
+    public static function getURL(string $url, int $id = null, int $parentId = null, bool $isAction = false): string
     {
+        if ($parentId === null) {
+            $parentId = 0;
+        }
+
         $parentIds = [$parentId];
 
         // 0, 1, 2, 3, 4 are all top levels, so we should place them on the same level
--- a/src/Common/Core/Header/Asset.php
+++ b/src/Common/Core/Header/Asset.php
@@ -13,11 +13,15 @@ final class Asset
     /** @var bool */
     private $addTimestamp;
 
+    /** @var int */
+    private $sequence;
+
     public function __construct(string $file, bool $addTimestamp = false, Priority $priority = null)
     {
         $this->file = $file;
         $this->addTimestamp = $addTimestamp;
         $this->priority = $priority ?? Priority::standard();
+        $this->sequence = 0;
     }
 
     public function getFile(): string
@@ -30,12 +34,22 @@ final class Asset
         return $this->priority;
     }
 
+    public function getSequence(): int
+    {
+        return $this->sequence;
+    }
+
     public function forCacheUrl(string $cacheUrl): self
     {
         $cacheAsset = clone $this;
         $cacheAsset->file = $cacheUrl;
 
         return $cacheAsset;
+    }
+
+    public function setSequence(int $sequence): void
+    {
+        $this->sequence = $sequence;
     }
 
     public function __toString(): string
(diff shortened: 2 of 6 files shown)


src/Backend/Modules/Pages/Engine/Model.php
```diff
@@ -918,8 +918,12 @@ public static function getTypes(): array
...
     }
 
-    public static function getURL(string $url, int $id = null, int $parentId = 0, bool $isAction = false): string
+    public static function getURL(string $url, int $id = null, int $parentId = null, bool $isAction = false): string
     {
+        if ($parentId === null) {
+            $parentId = 0;
+        }
+
         $parentIds = [$parentId];
```
carakas: you can remove the code above if you change this to
`$parentIds = [$parentId ?? 0];`

src/Common/Core/Header/AssetCollection.php
```diff
@@ -32,22 +32,37 @@ public function add(Asset $asset, bool $minify = true): void
...
     }
 
-    public function parse(BaseTwigTemplate $template, string $key): void
+    private function orderAssets(): void
     {
         usort(
             $this->assets,
             function (Asset $asset1, Asset $asset2) {
-                return $asset1->getPriority()->compare($asset2->getPriority());
+                $comparison = $asset1->getPriority()->compare($asset2->getPriority());
```
carakas: I would move the comparison code that is here completely to the Asset class, that way you don't need to expose the createdOn property that is only used to sort to the outside world and will reduce the complexity since it is an implementation detail

src/Common/Doctrine/ValueObject/AbstractFile.php
```diff
@@ -38,12 +38,12 @@
      */
     protected $namePrefix;
 
-    protected function __construct(string $fileName)
+    protected function __construct(?string $fileName)
```
carakas: why would you ever want to create a file with the filename null ?

jeroendesloovere: When "file" is "not" required.

src/Backend/Modules/Pages/Engine/Model.php
jeroendesloovere: Fixed by https://github.com/jeroendesloovere/forkcms/commit/34b6b8e52c0a44b5dba57db30519f9a899eee1a2

src/Common/Core/Header/AssetCollection.php
jeroendesloovere: Fixed in https://github.com/jeroendesloovere/forkcms/commit/b1c006cd6b07b2b723df6f42f00921af2e0e16c7

src/Common/Core/Header/Asset.php
```diff
@@ -26,6 +26,17 @@ public function __construct(string $file, bool $addTimestamp = false, Priority $
         $this->createdOn = new DateTimeImmutable();
     }
 
+    public function compare(Asset $asset)
+    {
+        $comparison = $this->priority->compare($asset->getPriority());
+
+        if ($comparison === 0) {
+            $comparison = $this->createdOn <=> $asset->getCreatedOn();
```
carakas: since it is the same class you can remove the getter and just use `$asset->createdOn` because it is the same class you can access the private properties

src/Common/Doctrine/ValueObject/AbstractFile.php
carakas: ok

src/Common/Core/Header/Asset.php
jeroendesloovere: Waaaa, that sounds amazing 👍

PR #2043, opened 2017-05-31
Fix html5 file input revalidation not working

carakas:
## Type

- Critical bugfix

## Pull request description

Selecting a new file doesn't trigger the input event but it does trigger the change event.
Because the input event wasn't triggered the custom event wasn't reset and therefore keeping the form in a permanent invalid state

### DIFF
--- a/src/Frontend/Core/Js/jquery/jquery.frontend.js
+++ b/src/Frontend/Core/Js/jquery/jquery.frontend.js
@@ -343,8 +343,7 @@
         var $input = $(this),
             errorMessage = '',
             type = '',
-            defaults =
-            {
+            defaults = {
                 required: jsFrontend.locale.err('FieldIsRequired'),
                 email: 	  jsFrontend.locale.err('EmailIsInvalid'),
                 date: 	  jsFrontend.locale.err('DateIsInvalid'),
@@ -369,7 +368,7 @@
 
             e.target.setCustomValidity(errorMessage);
 
-            $input.on('input', function(e) {
+            $input.on('input change', function(e) {
                 e.target.setCustomValidity('');
             });
         });
--- a/src/Frontend/Core/Js/jquery/jquery.frontend.js
+++ b/src/Frontend/Core/Js/jquery/jquery.frontend.js
@@ -345,16 +345,15 @@
             type = '',
             defaults = {
                 required: jsFrontend.locale.err('FieldIsRequired'),
-                email: 	  jsFrontend.locale.err('EmailIsInvalid'),
-                date: 	  jsFrontend.locale.err('DateIsInvalid'),
-                number:   jsFrontend.locale.err('NumberIsInvalid'),
-                value:    jsFrontend.locale.err('InvalidValue')
+                email: jsFrontend.locale.err('EmailIsInvalid'),
+                date: jsFrontend.locale.err('DateIsInvalid'),
+                number: jsFrontend.locale.err('NumberIsInvalid'),
+                value: jsFrontend.locale.err('InvalidValue')
             };
 
         options = $.extend(defaults, options);
 
         $input.on('invalid', function(e) {
-
             if ($input.context.validity.valueMissing) {
                 errorMessage = options.required;
             } else if (!$input.context.validity.valid) {
--- a/src/Frontend/Core/Js/jquery/jquery.frontend.js
+++ b/src/Frontend/Core/Js/jquery/jquery.frontend.js
@@ -337,41 +337,41 @@
  */
 (function($)
 {
-	$.fn.html5validation = function(options)
-	{
-		// define defaults
-		var $input = $(this),
-			errorMessage = '',
-			type = '',
-			defaults =
-			{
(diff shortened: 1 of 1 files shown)


WouterSioen:
👍 Looks good

src/Frontend/Core/Js/jquery/jquery.frontend.js
```diff
@@ -337,41 +337,40 @@
...
+
+        $input.on('invalid', function(e) {
+
+            if ($input.context.validity.valueMissing) {
+                errorMessage = options.required;
+            } else if (!$input.context.validity.valid) {
+                type = $input.context.type;
+                errorMessage = options.value;
+
+                if (options[type]) {
```
Thijzer: `options[type]` or `options[$input.context.type]` might be undefined here.

src/Frontend/Core/Js/jquery/jquery.frontend.js
```diff
@@ -337,41 +337,40 @@
...
+                    errorMessage = options[type];
```
Thijzer: suggestion : some helperMethod might help readability `translate(fieldType)` 
I toke a while to understand what was going on here.

```js
$input.on('invalid', function(e) {
  e.target.setCustomValidity(this.translate(this.retrieveErrorMessageFromInput($input)))
}
```

src/Frontend/Core/Js/jquery/jquery.frontend.js
```diff
@@ -337,41 +337,40 @@
...
+                }
+            }
+
+            e.target.setCustomValidity(errorMessage);
```
Thijzer: errorMessage might not be defined

src/Frontend/Core/Js/jquery/jquery.frontend.js
```diff
@@ -337,41 +337,40 @@
...
+
+            $input.on('input change', function(e) {
```
Thijzer: this event is trapped inside the `invalid` event so it's probably never called.

src/Frontend/Core/Js/jquery/jquery.frontend.js
carakas: isn't undefined falsy so it will still work?

src/Frontend/Core/Js/jquery/jquery.frontend.js
carakas: it is called, it is apart from the indentation fix the actual place where I changed something because it didn't work properly

src/Frontend/Core/Js/jquery/jquery.frontend.js
carakas: @mathiashelin what do you think? Personally I find it more clear and extendable in the current implementation, but that might just be taste

src/Frontend/Core/Js/jquery/jquery.frontend.js
carakas: undefined has the same effect as an empty string, in that case it will result in no custom error being shown

src/Frontend/Core/Js/jquery/jquery.frontend.js
mathiashelin: I think the current implementation is more clear and I guess the helperMethod would probably just do the same.

src/Frontend/Core/Js/jquery/jquery.frontend.js
```diff
@@ -337,41 +337,40 @@
...
-	};
+    $.fn.html5validation = function(options)
+    {
+        // define defaults
+        var $input = $(this),
+            errorMessage = '',
+            type = '',
+            defaults = {
+                required: jsFrontend.locale.err('FieldIsRequired'),
+                email: 	  jsFrontend.locale.err('EmailIsInvalid'),
```
jeroendesloovere: Don't love those extra spaces to divide "email:" from "jsFrontend.locale.err('EmailIsInvalid'),". Also present on all lines in this array...

src/Frontend/Core/Js/jquery/jquery.frontend.js
```diff
@@ -337,41 +337,40 @@
...
+                date: 	  jsFrontend.locale.err('DateIsInvalid'),
+                number:   jsFrontend.locale.err('NumberIsInvalid'),
+                value:    jsFrontend.locale.err('InvalidValue')
+            };
+
+        options = $.extend(defaults, options);
+
+        $input.on('invalid', function(e) {
+
```
jeroendesloovere: Can you remove this extra line?

src/Frontend/Core/Js/jquery/jquery.frontend.js
carakas: fixed in 099070048

src/Frontend/Core/Js/jquery/jquery.frontend.js
carakas: fixed in 0990700

PR #2048, opened 2017-06-02
Small label fix

bart-webleads:
## Type

- Enhancement

## Pull request description

The input is id "qWidget" and not "q_widget".


carakas:
@bart-webleads can you update the description by using the template we provide?

PR #2049, opened 2017-06-02
Fix ssl

OwenDeWaele:
## Type

- Enhancement

the http:  /  https: on external scripts

## Pull request description

Fix mixed content issues


jeroendesloovere:
Hi Owen, thanks for your PR.
Did you test if this works? Since // aren't absolute paths...?
My suggestion is using the "site.protocol" instead...

tijsverkoyen:
@jeroendesloovere if you don't specify a protocol it will use the protocol of the current page. So if you are using the site on https:, it will use https. (https://stackoverflow.com/questions/9161769/url-without-httphttps).

carakas:
@tijsverkoyen from what I see this PR only removed https and not http and of those services they all redirect http to https, so I don't really see a benefit of merging this one

jonasdekeukelaere:
No improvements, so closing this one.

Issue #2044, opened 2017-05-31
Some old default positions are missing in the new theme


carakas:
## Type

- Enhancement

## Problem description

Some core modules expect them to be there like the login widget or the search box

## Steps to reproduce

Install fork 5 and look in the backend at the pages

## Expected behavior / Proposed solutions

I think it would be best to add the missing position(s) back

jeroendesloovere:
Probably fixed by https://github.com/forkcms/forkcms/pull/2039

carakas:
@jeroendesloovere  nope :) Has nothing to do with that.

jeroendesloovere:
Ok, indeed, templates are empty
```
  <templates>
    <template label="Default" path="Core/Layout/Templates/Default.html.twig" image="false">
      <positions>
        <position name="main" />
      </positions>
      <format>
        [main]
      </format>
    </template>
  </templates>
```

jeroendesloovere:
Here is a screenshot what happens:
<img width="708" alt="schermafbeelding 2017-06-07 om 16 43 16" src="https://user-images.githubusercontent.com/588616/26884627-7de3c4c8-4ba0-11e7-9742-6b648909fea7.png">

jeroendesloovere:
Also when visiting the website, every page redirects you to 404.

jonasdekeukelaere:
The bug with the numbers in widget on the pages edit page has been fixed in #2063

PR #2053, opened 2017-06-08
Fix determining language for single language website

jonasdekeukelaere:
## Type
- Critical bugfix

After a clean install of a single language website, all the frontend pages give a 404, as described in https://github.com/forkcms/forkcms/issues/2044#issuecomment-306817211

## Pull request description
Will fix the determination of the language for a single language website


carakas:
Has been recreated here because the merge was lost https://github.com/forkcms/forkcms/pull/2054

PR #2060, opened 2017-06-08
Replace location address by one textarea

### LINKED ISSUE #1860: Google maps is geolocating incorrectly in some countries

jonasdekeukelaere:
So people can format the address how they like.

## Type
- Non critical bugfix

Fixes #1860

## Pull request description
Replaces the separate address text fields by one text area for the address.


jeroendesloovere:
👎 Con this PR

src/Backend/Modules/Location/Installer/Data/install.sql
```diff
@@ -3,10 +3,7 @@ CREATE TABLE IF NOT EXISTS `location` (
  `language` varchar(5) NOT NULL,
  `extra_id` int(11) NOT NULL,
  `title` varchar(255) NOT NULL,
- `street` varchar(255) NOT NULL,
- `number` varchar(255) NOT NULL,
- `zip` varchar(255) NOT NULL,
- `city` varchar(255) NOT NULL,
+ `address` text NOT NULL,
```
carakas: can you also do this change in the test database

carakas: I'm also missing an entry in the upgrade guide

jonasdekeukelaere: Fixed in 11b948d and 70951fd

jonasdekeukelaere:
Fixed in f456acb

src/Frontend/Modules/Location/Layout/Templates/Index.html.twig
```diff
@@ -25,9 +25,7 @@
 
             <div itemprop="name" class="hidden"><strong>{{ locationItem.title }}</strong></div>
             <address itemprop="address" itemscope itemtype="http://schema.org/PostalAddress">
-              <span itemprop="streetAddress">{{ locationItem.street }} {{ locationItem.number }}</span><br>
-              <span itemprop="postalCode">{{ locationItem.zip }}</span>
-              <span itemprop="addressLocality">{{ locationItem.city }}</span><br>
+              <span itemprop="streetAddress">{{ locationItem.address }}</span><br>
```
carakas: isn't streetAddress the wrong itemprop here?

src/Frontend/Modules/Location/Layout/Widgets/Location.html.twig
```diff
@@ -63,9 +63,7 @@
 
           <div itemprop="name" class="hidden"><strong>{{ widgetLocationItem.title }}</strong></div>
           <address itemprop="address" itemscope itemtype="http://schema.org/PostalAddress">
-            <span itemprop="streetAddress">{{ widgetLocationItem.street }} {{ widgetLocationItem.number }}</span><br>
-            <span itemprop="postalCode">{{ widgetLocationItem.zip }}</span>
-            <span itemprop="addressLocality">{{ widgetLocationItem.city }}</span><br>
+            <span itemprop="streetAddress">{{ widgetLocationItem.address }}</span><br>
```
carakas: idem

src/Frontend/Modules/Location/Layout/Templates/Index.html.twig
jonasdekeukelaere: It was the closest I could find, there is none for combined address.

jeroendesloovere: That is because it isn't advised to put the entire address in one address field.
Zip, city and country should always be separate.
schermafbeelding 2017-06-13 om 08 32 40

carakas: if that is the case then it looks to me that this is a bad solution to the problem at hand

carakas: Maybe we can add a setting where people can set the format they want to display with placeholders?
Something like this
```
[street] [number]
[postalCode] [locality]
[country]
```

jonasdekeukelaere: This looks like a big overhead for something very simple as an address. Maybe split it like addressLine, postcode, city, country?

tijsverkoyen:
I think there is a lot of useless fuzz over this PR. I would say to drop the current PR, leave the existing functionality as it is.
When there is more time, we can try to find a way that works in more use cases.

PR #2066, opened 2017-06-09
Password hashing

### LINKED ISSUE #1817: Move to password_hash for our passwords
## Type
- Enhancement
## Problem description

We currently have some custom combination of hashing algorithms to has our passwords

We should move to the password hashing methods of php.

We can use the method `password_needs_rehash` to detect if the password has already converted
We also van use `password_get_info` to detect the hashing algorithm, in our case it will return undefined so we can detect when we need to use our current system as a fallback before rehashing it with the password functions

jonasdekeukelaere:
## Type
- Enhancement

fixes #1817 

## Pull request description
Replaces the old sha1/md5 hashing of password to the PHP password_hash method.


jonasdekeukelaere:
WIP: frontend passwords needs to be changed as well.

src/Backend/Core/Engine/Authentication.php
```diff
@@ -313,16 +316,40 @@ public static function loginUser(string $login, string $password): bool
     {
         $db = BackendModel::get('database');
 
-        // fetch the encrypted password
-        $passwordEncrypted = static::getEncryptedPassword($login, $password);
+        // build fallback for migrating devs
+        // todo remove in next major release
+        $encryptedPassword = BackendUsersModel::getEncryptedPassword($login);
```
carakas: while this is a nice to have, won't this conflict if for instance php 5.2 or 5.3 would have a new implementation?

Also I know that the php implementation doesn't need the salt but didn't fork use a salt?

I think it would be better to just break it here and say that people will have to use the reset password (we might provide a command to trigger that)

Implementing this will make it way harder to switch internally to the symfony security component withing this mayor version and I'm not sure if the few people that will upgrade from 4 to 5 should stand in the way of this.

src/Frontend/Modules/Profiles/Actions/Register.php
```diff
@@ -118,10 +118,7 @@ private function validateForm(): void
 
                 // values
                 $values['email'] = $txtEmail->getValue();
-                $values['password'] = FrontendProfilesModel::getEncryptedString(
-                    $txtPassword->getValue(),
-                    $settings['salt']
```
carakas: the entire salt setting should be removed

src/Frontend/Modules/Profiles/Engine/Authentication.php
```diff
@@ -83,21 +83,38 @@ public static function cleanupOldSessions(): void
...
 
-        // encrypt password
-        $encryptedPassword = FrontendProfilesModel::getEncryptedString(
-            $password,
-            FrontendProfilesModel::getSetting($profileId, 'salt')
-        );
+            // check old password
+            $encryptedPasswordOldMethod = FrontendProfilesModel::getEncryptedString(
+                $password,
+                FrontendProfilesModel::getSetting($profileId, 'salt')
```
carakas: same comment as in the backend

src/Frontend/Modules/Profiles/Actions/Register.php
jonasdekeukelaere: In the profile module it is also used for reset password key and activation key, so can't be removed completely.

src/Backend/Core/Engine/Authentication.php
jonasdekeukelaere: In the original issue you said that this should be there. I don't understand what php 5.2 or 5.3 has to do with this?

This can just be removed in the next release, right?

carakas: I meant 7.2 or 7.3.
If we now include this fallback it needs to be carried on all the way till fork 6.
There is an implementation for password_hash in the security component from symfony meaning that by switching to password hash can be done in a feature release since the implmentation is something internal. But if we would add this fallback we would need to write a custom implementation for it to carry on our old implementation from fork.

The issue was originaly written for 4.5 in a way that it would be backwards compatible. Since this is a major version we can break that

src/Frontend/Modules/Profiles/Actions/Register.php
carakas: maybe those implementations should be changed then

src/Backend/Core/Engine/Authentication.php
jonasdekeukelaere: So, just remove the fallback and let the dev worry about the problems?

carakas: with an entry in the UPGRADE file yes, but maybe others have different opinions on this

jonasdekeukelaere: @tijsverkoyen @StijnVrolijk ?

tijsverkoyen: I agree with Carakas. But it should be added in the UPGRADE file.

jonasdekeukelaere: Fallback remove in 33481e4

jonasdekeukelaere:
@carakas fallback for old accounts/profiles has been removed
what do we do with the salt that is used in frontend for activation key and reset password?

carakas:
I would just generate random strings to generate those and remove the salt

jonasdekeukelaere:
@carakas It was only 1 instance, d20441b03, the other already use this implementation.

src/Frontend/Modules/Profiles/Actions/Register.php
jonasdekeukelaere: changes in d20441b039bc0a062d7770a1621138e261ffdd8e

PR #2068, opened 2017-06-09
Updated ContentBlocks Installer

jeroendesloovere:
## Type

- Enhancement

## Pull request description

> In the contentBlocks module, moved most the installer code to properly named methods.

### DIFF
--- a/src/Backend/Modules/ContentBlocks/Installer/Installer.php
+++ b/src/Backend/Modules/ContentBlocks/Installer/Installer.php
@@ -28,7 +28,7 @@ class Installer extends ModuleInstaller
         $this->configureRights();
     }
 
-    protected function configureBackendNavigation(): void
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/ContentBlocks/Installer/Installer.php
```diff
@@ -20,34 +20,43 @@ class Installer extends ModuleInstaller
...
+        $this->configureSettings();
+        $this->configureBackendNavigation();
+        $this->configureBackendRights();
+    }
 
-        // general settings
-        $this->setSetting($this->getModule(), 'max_num_revisions', 20);
+    private function configureBackendNavigation(): void
+    {
+        // Set navigation for "Modules"
```
carakas: I think this comment can be removed since it is rather vague and the method name already explains it

PR #2071, opened 2017-06-09
Updated "Extensions" installer

jeroendesloovere:
Changes:
- Moved functionality "Extensions" to separate methods
- Fixed bug: using Search “Form” instead of “form”
- Removed "Core" theme (since the only theme remaining in fork cms 5 is "Fork")

## Type

- Enhancement

## Pull request description

> In the Extensions module, restructured installer functionality into properly named methods.

### DIFF
--- a/src/Backend/Modules/Extensions/Installer/Installer.php
+++ b/src/Backend/Modules/Extensions/Installer/Installer.php
@@ -85,6 +85,7 @@ class Installer extends ModuleInstaller
     private function configureBackendRights(): void
     {
         $this->setModuleRights(1, $this->getModule());
+
         $this->configureBackendActionRightsForModules();
         $this->configureBackendActionRightsForTemplates();
         $this->configureBackendActionRightsForThemes();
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Extensions/Installer/Installer.php
```diff
@@ -126,39 +80,24 @@ public function install(): void
...
+        $this->configureBackendActionRightsForThemes();
     }
 
-    private function installForkTheme(): void
+    private function configureFrontendExtras(): void
     {
-        /*
-         * Fallback templates
-         */
+        $this->insertExtra('Search', ModuleExtraType::widget(), 'SearchForm', 'Form');
```
jonasdekeukelaere: Because of this the widget is added more than once in the database, is this necessary here?

jonasdekeukelaere: Nvm, was bug - https://github.com/forkcms/forkcms/pull/2088

PR #2079, opened 2017-06-09
Updated "Pages" installer

jeroendesloovere:
Changes:
- Refactored to cleaner methods
- Fixed duplicate code
- Removed “extra” sequence, because this is auto generated

## Type

- Enhancement

## Pull request description

> In the Pages module, restructured installer functionality into properly named methods.

### DIFF
--- a/src/Backend/Modules/Pages/Installer/Installer.php
+++ b/src/Backend/Modules/Pages/Installer/Installer.php
@@ -17,230 +17,68 @@ use Common\ModuleExtraType;
  */
 class Installer extends ModuleInstaller
 {
-    private function importData(): void
-    {
-        // insert required pages
-        $this->insertPages();
+    /** @var int */
+    private $searchFormWidgetId;
 
-        // install example data if requested
-        if ($this->installExample()) {
-            $this->installExampleData();
-        }
-    }
+    /** @var int */
+    private $searchWidgetId;
 
-    private function insertPages(): void
-    {
-        // get extra ids
-        $extras = [];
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Pages/Installer/Installer.php
```diff
@@ -356,28 +181,123 @@ private function installExampleData(): void
...
             }
         }
     }
 
-    private function setRights(): void
+    private function configureFrontendExtras(): void
     {
-        // module rights
-        $this->setModuleRights(1, $this->getModule());
+        $this->extraIds['search'] = $this->insertExtra('Search', ModuleExtraType::block(), 'Search');
```
jonasdekeukelaere: The problem here is that this also get added in the search installer, what results in having this widget more than once in the modules_extras table.

jonasdekeukelaere: Nvm, was bug - https://github.com/forkcms/forkcms/pull/2088

src/Backend/Modules/Pages/Installer/Installer.php
```diff
@@ -356,28 +181,123 @@ private function installExampleData(): void
...
 
-        // action rights
-        $this->setActionRights(1, $this->getModule(), 'GetInfo');
-        $this->setActionRights(1, $this->getModule(), 'Move');
+    private function configureFrontendPages(): void
+    {
+        // loop languages
+        foreach ($this->getLanguages() as $language) {
+            // check if pages already exist for this language
+            if (!$this->hasPage($language)) {
```
carakas: I would replace this with ```php
if ($this->hasPage($language) {
    continue;
}

... the code that is now inside the big if statement

PR #2083, opened 2017-06-09
Updated "Tags" installer

jeroendesloovere:
## Type

- Enhancement

## Pull request description

> In the Tags module, restructured installer functionality into properly named methods.

### DIFF
--- a/src/Backend/Modules/Tags/Installer/Installer.php
+++ b/src/Backend/Modules/Tags/Installer/Installer.php
@@ -65,7 +65,7 @@ class Installer extends ModuleInstaller
                 // insert contact page
                 $this->insertPage(
                     [
-                        'title' => 'Tags',
+                        'title' => $this->getModule(),
                         'type' => 'root',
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Tags/Installer/Installer.php
```diff
@@ -17,66 +17,85 @@
...
-            if (!(bool) $this->getDB()->getVar(
-                'SELECT 1
-                 FROM pages AS p
-                 INNER JOIN pages_blocks AS b ON b.revision_id = p.revision_id
-                 WHERE b.extra_id = ? AND p.language = ?
-                 LIMIT 1',
-                [$tagsID, $language]
-            )
-            ) {
+            if (!$this->hasTagsExtra($language)) {
```
carakas: same feedback as in the other PR's

jeroendesloovere: Fixed with 19514c8

PR #2088, opened 2017-06-12
Fix saving null values for module extra data

jonasdekeukelaere:
## Type
- Non critical bugfix

## Pull request description
Denies duplicate modules extras.

### DIFF
--- a/src/Backend/Core/Engine/Model.php
+++ b/src/Backend/Core/Engine/Model.php
@@ -389,7 +389,7 @@ class Model extends \Common\Core\Model
             array_filter(
                 $moduleExtras,
                 function (array $data) use ($key, $value) {
-                    $data = unserialize($data);
+                    $data = $data === null ? [] : unserialize($data);
 
                     return isset($data[$key]) && $data[$key] === $value;
                 }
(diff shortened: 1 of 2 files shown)


src/Backend/Core/Engine/Model.php
```diff
@@ -708,7 +708,7 @@ public static function insertExtra(
                 'type' => $type,
                 'label' => $label ?? $module, // if label is empty, fallback to module
                 'action' => $action ?? $module, // if action is empty, fallback to module
-                'data' => serialize($data),
+                'data' => $data ===  null ? null : serialize($data),
```
carakas: is this also checked when unserialising ?

jonasdekeukelaere: Fixed in 52a15bead

src/Backend/Core/Engine/Model.php
```diff
@@ -389,7 +389,7 @@ public static function getExtrasForData(string $module, string $key, string $val
             array_filter(
                 $moduleExtras,
                 function (array $data) use ($key, $value) {
-                    $data = unserialize($data);
+                    $data = $data === null ? [] : unserialize($data);
```
carakas: wouldn't it make more sense to return null if the data is null ?

jonasdekeukelaere: Didn't change anything to the logic, as the return value should be a bool ` return isset($data[$key]) && $data[$key] === $value;` (filter method)

carakas: ah k, sorry, missed that

PR #2093, opened 2017-06-13
Remove module as default for module extra on module install

jonasdekeukelaere:
As some modules have not always specific block for each action, this is breaking those module.
For example blog module, where detail etc is called through the translation actions.

## Type
- Critical bugfix

## Pull request description
Making modules with no specific block for each action work again, for example blog.

### DIFF
--- a/src/Backend/Core/Engine/Model.php
+++ b/src/Backend/Core/Engine/Model.php
@@ -707,7 +707,7 @@ class Model extends \Common\Core\Model
                 'module' => $module,
                 'type' => $type,
                 'label' => $label ?? $module, // if label is empty, fallback to module
-                'action' => $action ?? $module, // if action is empty, fallback to module
+                'action' => $action ?? null,
                 'data' => $data ===  null ? null : serialize($data),
                 'hidden' => $hidden ? 'Y' : 'N',
                 'sequence' => $sequence ?? self::getNextModuleExtraSequenceForModule($module),


src/Backend/Core/Engine/Model.php
```diff
@@ -707,7 +707,7 @@ public static function insertExtra(
                 'module' => $module,
                 'type' => $type,
                 'label' => $label ?? $module, // if label is empty, fallback to module
-                'action' => $action ?? $module, // if action is empty, fallback to module
+                'action' => $action ?? null,
```
carakas: Am I wrong to think  that this is then also an issue in the current master branch?
https://github.com/forkcms/forkcms/blob/master/src/Backend/Core/Engine/Model.php#L747

jonasdekeukelaere: I would think so, but can't reproduce on demo site :/

jonasdekeukelaere: ModuleInstaller on master branch uses his own insertExtra method: https://github.com/forkcms/forkcms/blob/master/src/Backend/Core/Installer/ModuleInstaller.php#L500 - that's why it's not an issue on installation.

carakas:
can you give some more details on where and how it goes wrong with the blog?

jonasdekeukelaere:
As Blog will be set as action for the Blog block, all pages will arrive on the Index action (because Blog action will not exists). Also url generation will give 404 for any action.

jonasdekeukelaere:
Also when adding/editing blog items, you'll get the message "Cannot generate URL. Create a page that has this module linked to it." - as the url generation can't find a block for detail - as the only block the blog module has would be the "Blog" block. (Not an general block for all actions with action NULL)

PR #2095, opened 2017-06-14
Replace &amp; by & as parameter query divider

jonasdekeukelaere:
## Type
- Critical bugfix

## Pull request description
Using &amp; will resolve in the following queryString: index?foo=test&amp;bar=test&amp;token=x - which will resolve in the following parameters: foo = test, amp;=bar = test, amp;token = x
Replacing it by & will resolve this issue.

### DIFF
--- a/src/Backend/Core/Engine/Model.php
+++ b/src/Backend/Core/Engine/Model.php
@@ -96,7 +96,7 @@ class Model extends \Common\Core\Model
             $parameters['sort'] = $queryParameterBag->get('sort');
         }
 
-        $queryString = '?' . http_build_query($parameters, null, '&');
+        $queryString = '?' . http_build_query($parameters);
 
         if (!$encodeSquareBrackets) {
             // we use things like [id] to parse database column data in so we need to unescape those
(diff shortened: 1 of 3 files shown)


src/Backend/Core/Engine/Model.php
```diff
@@ -96,7 +96,7 @@ public static function createURLForAction(
             $parameters['sort'] = $queryParameterBag->get('sort');
         }
 
-        $queryString = '?' . http_build_query($parameters, null, '&amp;');
+        $queryString = '?' . http_build_query($parameters, null, '&');
```
carakas: `&` is the default so I think you can just use `http_build_query($parameters)`

jonasdekeukelaere: Fixed ecf455d

jonasdekeukelaere:
@carakas Also found some more uses of &amp; instead of & in other classes.

carakas:
@jonasdekeukelaere did you test all the usecases that this doesn't give issues in the template?

jonasdekeukelaere:
I didn't test all use cases, but I tested it where it went wrong (contentblocks - report after editing/deleting was incorrect) and works. Also checked the rss (frontend) where the frontend fix is being used. In both cases the url is now correct.

PR #2091, opened 2017-06-13
Using Domain structure for ContentBlocks module

jeroendesloovere:
## Type

- Enhancement

## Pull request description

> The current ContentBlocks module in Fork CMS uses Doctrine and Symfony forms. That's amazing, but we are going a step further by dividing the logic in a folder called `Domain`. This is the way we are going to build all modules in Fork CMS. If you want to see some examples with the `Domain` folder, you can check out the MediaGalleries and MediaLibrary modules.

PR #2096, opened 2017-06-14
Domain structure content blocks

jeroendesloovere:
## Type

- Enhancement

> New PR for https://github.com/forkcms/forkcms/pull/2091

## Pull request description

> The current ContentBlocks module in Fork CMS uses Doctrine and Symfony forms. That's amazing, but we are going a step further by dividing the logic in a folder called `Domain`. This is the way we are going to build all modules in Fork CMS. If you want to see some examples with the `Domain` folder, you can check out the MediaGalleries and MediaLibrary modules.


jonasdekeukelaere:
Isn't moving everything into a Domain folder (with only Command, Event and Exception in it's own folder) making it more complicated to see what is what? Instead of clear folder names Command, DBALType, Entity, Form, Repository?
And why the extra ContentBlocks folder in the Domain folder?
Also why make the action class name unnecessary longer?
Those changes seem a bit unclear to me really.

jeroendesloovere:
Hi @jonasdekeukelaere,
Well, the Domain folder is the way all fork modules are going to be build in the future.
Since in the other way around, all logic is shattered in different folders Command/DBALType/Entity/... 

The directory /Domain/ContentBlock is there so other folders can be created, like f.e.: /Domain/Settings,  ...

The action class is indeed something that can be discussed... since the module is so small...
But, if we want to be consistent, I think that we should use the new Action class names all around. Just like we did in the new MediaGalleries and MediaLibrary modules: MediaGalleryAdd, MediaGalleryDelete, MediaFolderDelete, MediaItemEdit, MediaItemIndex, MediaItemUpload, ...
=> Which also offers easy integration of new Action classes if the module ever should be expanded... ContentBlockRevisionDelete, ContentBlockMassAction, ... That would save big time later (changes in locale, Action rights, ...)

jonasdekeukelaere:
I think is more logically and clearer when you split it up in folder instead of dropping it all in 1 folder, as then you'll have no idea what each class is suppose to be.

About the class name, I think the long name with the module name is not a good way. You'll need to repeat this in so many places, like url creation, templates, permission checking, ... I think the short way is so much easier.

carakas:
@jeroendesloovere I think in this case I would keep the action classes the same as they are in fork 4
With block you have post and category, it makes sense there, but here it just confuses. Also I'm not convinced by the examples you provide, there is no confusion possible here like it is with the blog

apart from that I agree with your points.

carakas:
@jonasdekeukelaere It makes more sense to group all things related to the entity in one parent namespace and separate that from the structure of the cms

If you have a category and a post for instance just adding everything in repository is the wrong approach imho (and also others) since you are grouping it wrong, The repository of the category has nothing to do with the repository of the post.

The reason I am for this is because this is a way simpler module so it is easier to point to then a more complex module like the media library.

The structure should be consistent

jonasdekeukelaere:
For me it's more clearer to have (for example) all repositories, (or all Form types) in a folder together than have all those classes in 1 folder under Domain. Then you know directly what each class is suppose to do. But that sounds too easy maybe.

carakas:
@jeroendesloovere I usually point to https://github.com/sumocoders/fork-partners/tree/master/src/Backend/Modules/Partners since it is a rather simple module. I also just used add edit delete there because it made sense

carakas:
@jonasdekeukelaere no need to get passive aggressive

jonasdekeukelaere:
For example the partners module, you have everything in this 1 folder https://github.com/sumocoders/fork-partners/tree/master/src/Backend/Modules/Partners/Domain/Partner - doesn't look right, really.

carakas:
the problem with your approach @jonasdekeukelaere is that the main goal of namespaces is to group and avoid naming collisions.

If you have a post and a category, and both of them have a status value object but it is not the same value object
Then in your case you would need to change the proper name of the class to keep your file stucture while you should do it with namespaces.

When there are a lot of things of the same type withing a domain namespace I always add an extra sublevel, for instance with commands

jonasdekeukelaere:
Then you can start making subfolders in your (for example) Repository or ValueObject folder.

carakas:
Then you are grouping them wrong.
The thing they have in common is the entity they belong to, so that should be the thing you group them by.

If you have red trousers, blue trousers, a red sweater, a yellow sweater, a blue tshirt, a black tshirt, a red laptop, a blue phone and a red tablet

you are going to devide them in for instance electronics and clothing, you aren't going to create a group of red thing, a group of blue things etc

jeroendesloovere:
@jonasdekeukelaere
When I first started using doctrine and symfony, Entity/Type/ValueObject/DBALTYpe folders were the best things I ever heard of...
Then @WouterSioen and @carakas told me about a Domain structure.
At first, I was against it, just like you.
But then I decided to try it and damn now I created a shitload of new modules/apps using the Domain structure. It's so much nicer.

Take a look at this "Realizations module" example:
I swear, "I'm never going back to Entity/Form/..." structure.
schermafbeelding 2017-06-14 om 14 47 45

jonasdekeukelaere:
Ok agree about that the entity is the common thing. But then I would make subfolders in the Domain/ContentBlock folder. Because now in @jeroendesloovere his example you just have a bunch of files that not really belong together. Some are Entity, some are DTO, some Form type, some Repository.

src/Backend/Modules/ContentBlocks/Actions/ContentBlockAdd.php
```diff
@@ -12,28 +12,23 @@
...
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\Command\CreateContentBlock;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\ContentBlockType;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\Event\ContentBlockCreated;
+use Symfony\Component\Form\Form;
 
 /**
  * This is the add-action, it will display a form to create a new item
  */
-class Add extends BackendBaseActionAdd
+class ContentBlockAdd extends BackendBaseActionAdd
```
carakas: I would keep this as Add if for some unforeseeable reason this would once in the future make the naming incorrect we can tackle it then, for now I would keep it at Add, Edit, Delete, and Index

src/Backend/Modules/ContentBlocks/Actions/ContentBlockAdd.php
```diff
@@ -43,27 +38,55 @@ public function execute(): void
         }
 
         /** @var CreateContentBlock $createContentBlock */
```
carakas: I don't think we need this anymore since the method has a return type and that gives us the autocompletion that this provided in the past

src/Backend/Modules/ContentBlocks/Actions/ContentBlockAdd.php
```diff
@@ -43,27 +38,55 @@ public function execute(): void
...
 
         $this->redirect(
-            BackendModel::createURLForAction(
-                'Index',
-                null,
-                null,
+            $this->getBackLink(
                 [
-                    'report' => 'added',
+                    'report' => 'content-block-added',
```
carakas: why did you change this?

src/Backend/Modules/ContentBlocks/Actions/ContentBlockAdd.php
```diff
@@ -43,27 +38,55 @@ public function execute(): void
...
+            $parameters
+        );
+    }
+
+    private function getForm(): Form
+    {
+        $form = $this->createForm(
+            ContentBlockType::class,
+            new CreateContentBlock(),
+            ['theme' => $this->get('fork.settings')->get('Core', 'theme', 'Core')]
```
carakas: see previous feedback, can you rebase on 5.0.0-dev because you are missing changes or fix this manually and check that you didn't occidentally remove something, I would personally prefer the rebase

src/Backend/Modules/ContentBlocks/Actions/ContentBlockDelete.php
```diff
@@ -0,0 +1,69 @@
...
+use Backend\Core\Language\Locale;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\Command\DeleteContentBlock;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\ContentBlock;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\Event\ContentBlockDeleted;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\Exception\ContentBlockNotFound;
+
+/**
+ * This is the delete-action, it will delete an item.
+ */
+class ContentBlockDelete extends BackendBaseActionDelete
```
carakas: same comment as above

src/Backend/Modules/ContentBlocks/Actions/ContentBlockDelete.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $this->get('event_dispatcher')->dispatch(
+            ContentBlockDeleted::EVENT_NAME,
+            new ContentBlockDeleted($contentBlock)
+        );
+
+        $this->redirect(
+            $this->getBackLink(
+                [
+                    'report' => 'content-block-deleted',
```
carakas: idem

jonasdekeukelaere:
Sure they belong together, but for example all the code in 1 class belongs together as well, you will split that up into methods, right? For example the refactoring of all the installers. Why is that different with files and ok to just drop them al in 1 folder?

src/Backend/Modules/ContentBlocks/Actions/ContentBlockEdit.php
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * This is the edit-action, it will display a form to edit an existing item
+ */
+class ContentBlockEdit extends BackendBaseActionEdit
+{
+    public function execute(): void
+    {
+        parent::execute();
+
+        /** @var ContentBlock $contentBlock */
```
carakas: idem

src/Backend/Modules/ContentBlocks/Actions/ContentBlockEdit.php
```diff
@@ -0,0 +1,130 @@
...
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\ContentBlockRevisionDataGrid;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\ContentBlockType;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\Event\ContentBlockUpdated;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\Exception\ContentBlockNotFound;
+use Symfony\Component\Form\Form;
+
+/**
+ * This is the edit-action, it will display a form to edit an existing item
+ */
+class ContentBlockEdit extends BackendBaseActionEdit
```
carakas: idem

src/Backend/Modules/ContentBlocks/Actions/ContentBlockEdit.php
```diff
@@ -0,0 +1,130 @@
...
+            $this->tpl->assign('contentBlock', $contentBlock);
+            $this->tpl->assign('revisions', ContentBlockRevisionDataGrid::getHtml($contentBlock, Locale::workingLocale()));
+
+            $this->parse();
+            $this->display();
+
+            return;
+        }
+
+        /** @var UpdateContentBlock $updateContentBlock */
```
carakas: idem

src/Backend/Modules/ContentBlocks/Actions/ContentBlockEdit.php
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    private function getForm(ContentBlock $contentBlock): Form
+    {
+        $form = $this->createForm(
+            ContentBlockType::class,
+            new UpdateContentBlock($contentBlock),
+            [
+                'theme' => $this->get('fork.settings')->get('Core', 'theme', 'Core'),
```
carakas: idem

src/Backend/Modules/ContentBlocks/Actions/ContentBlockIndex.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+use Backend\Core\Engine\Base\ActionIndex as BackendBaseActionIndex;
+use Backend\Core\Language\Locale;
+use Backend\Modules\ContentBlocks\Domain\ContentBlock\ContentBlockDataGrid;
+
+/**
+ * This is the index-action (default), it will display the overview
+ */
+class ContentBlockIndex extends BackendBaseActionIndex
```
carakas: idem

src/Backend/Modules/ContentBlocks/Installer/Data/locale.xml
```diff
@@ -31,20 +16,22 @@
...
-        <translation language="it"><![CDATA[Sei sicuro di voler eliminare il blocco di contenuto "%1$s"?]]></translation>
-        <translation language="ru"><![CDATA[Вы действительно хотите удалить блок "%1$s"?]]></translation>
-        <translation language="zh"><![CDATA[你确定要删除内容块“%1$s”吗？]]></translation>
-        <translation language="de"><![CDATA[Soll der Inhaltsblock "%1$s" wirklich gelöscht werden?]]></translation>
-        <translation language="fr"><![CDATA[Etes-vous sûr de vouloir effacer le bloc de contenu "%1$s"?]]></translation>
-        <translation language="hu"><![CDATA[Biztos törölni akarod a(z) "%1$s" tartalmi blokkot?]]></translation>
-        <translation language="es"><![CDATA[¿Estás seguro de que quieres eliminar el bloque de contenido "%1$s"?]]></translation>
-        <translation language="sv"><![CDATA[Är du säker på att du vill till ta bort innehållsblocket "%1$s"?]]></translation>
-        <translation language="el"><![CDATA[Σίγουρα θέλετε να διαγράψετε το μπλοκ περιεχομένου "%1$s";]]></translation>
+      <item type="label" name="ContentBlockAdd">
```
carakas: I would also revert this

src/Backend/Modules/ContentBlocks/Installer/Installer.php
```diff
@@ -44,10 +44,10 @@ private function configureBackendRights(): void
     {
         $this->setModuleRights(1, $this->getModule());
 
-        $this->setActionRights(1, $this->getModule(), 'Add');
-        $this->setActionRights(1, $this->getModule(), 'Delete');
-        $this->setActionRights(1, $this->getModule(), 'Edit');
-        $this->setActionRights(1, $this->getModule(), 'Index');
+        $this->setActionRights(1, $this->getModule(), 'ContentBlockAdd');
```
carakas: idem

src/Backend/Modules/ContentBlocks/Config.php
```diff
@@ -16,4 +16,10 @@
  */
 class Config extends BackendBaseConfig
 {
+    /**
+     * The default action
+     *
+     * @var string
+     */
+    protected $defaultAction = 'ContentBlockIndex';
```
carakas: idem

src/Backend/Modules/ContentBlocks/Layout/Templates/ContentBlockEdit.html.twig
```diff
@@ -51,7 +51,7 @@
     <div class="col-md-12">
       <div class="btn-toolbar">
         <div class="btn-group pull-left" role="group">
-          {% if showContentBlocksDelete %}
+          {% if showContentBlocksContentBlockDelete %}
```
carakas: idem, this actually points out nicely why in this case it is a bad idea, this is almost unreadable

src/Backend/Modules/ContentBlocks/Layout/Templates/ContentBlockEdit.html.twig
```diff
@@ -64,21 +64,21 @@
           </button>
         </div>
       </div>
-      {% if showContentBlocksDelete %}
+      {% if showContentBlocksContentBlockDelete %}
```
carakas: idem

src/Backend/Modules/ContentBlocks/Layout/Templates/ContentBlockEdit.html.twig
```diff
@@ -64,21 +64,21 @@
...
               <div class="modal-body">
-                <p>{{ 'msg.ConfirmDelete'|trans|format(title)|raw }}</p>
+                <p>{{ 'msg.ConfirmDelete'|trans|format(contentBlock.title)|raw }}</p>
               </div>
               <div class="modal-footer">
                 <button type="button" class="btn btn-default" data-dismiss="modal">
                   <span class="fa fa-times"></span>{{ 'lbl.Cancel'|trans|ucfirst }}
                 </button>
-                <a href="{{ geturl('delete') }}&amp;id={{ id }}" class="btn btn-danger">
+                <a href="{{ geturl('ContentBlockDelete') }}&amp;id={{ contentBlock.id }}" class="btn btn-danger">
```
carakas: idem but should be `Delete`

src/Backend/Modules/ContentBlocks/Layout/Templates/ContentBlockIndex.html.twig
```diff
@@ -1,10 +1,10 @@
 {% extends '/Layout/Templates/base.html.twig' %}
 
 {% block actionbar %}
-  {% if showContentBlocksAdd %}
+  {% if showContentBlocksContentBlockAdd %}
```
carakas: idem

src/Backend/Modules/ContentBlocks/Layout/Templates/ContentBlockIndex.html.twig
```diff
@@ -1,10 +1,10 @@
...
     <div class="btn-toolbar pull-right">
       <div class="btn-group" role="group">
-        {{ macro.buttonIcon( geturl('add'), 'plus-square', 'lbl.Add'|trans|ucfirst) }}
+        {{ macro.buttonIcon( geturl('ContentBlockAdd'), 'plus-square', 'lbl.Add'|trans|ucfirst) }}
```
carakas: idem but should be `Add`

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlockDataTransferObject.php
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public $id;
+
+    /**
+     * @var int
+     */
+    public $extraId;
+
+    /**
+     * @var ?int
```
carakas: not sure it this is valid phpdoc, I think it needs to be int|null

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlockDataTransferObject.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        return $this->contentBlockEntity;
+    }
+
+    public function hasExistingContentBlock(): bool
+    {
+        return $this->contentBlockEntity instanceof ContentBlock;
+    }
+
+    public function setContentBlockEntity(ContentBlock $contentBlockEntity): void
```
carakas: where do you use this one?

I'm not really a fan of seeing that here, if it is needed for a command maybe we should make the entity protected and add that method in the command.

Cause this can have serious side effects, you have one entity, this DTO is populated based on that and then you can overwrite the entity but the DTO won't get the values from it

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlock.php
```diff
@@ -1,16 +1,15 @@
...
 use Backend\Core\Engine\Model;
-use Backend\Modules\ContentBlocks\ValueObject\ContentBlockStatus;
 use Common\Locale;
 use DateTime;
 use Doctrine\ORM\Mapping as ORM;
 
 /**
  * @ORM\Table(name="content_blocks")
- * @ORM\Entity(repositoryClass="Backend\Modules\ContentBlocks\Repository\ContentBlockRepository")
+ * @ORM\Entity(repositoryClass="Backend\Modules\ContentBlocks\Domain\ContentBlock\ContentBlockRepository")
```
carakas: you should try it first, but I think because they are in the same namespace you can drop the FQCN and just use `ContentBlockRepository`

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlock.php
```diff
@@ -125,26 +124,29 @@ private function __construct(
...
-        string $text,
-        bool $isHidden,
-        string $template = self::DEFAULT_TEMPLATE
-    ) : self {
+    public static function fromDataTransferObject(ContentBlockDataTransferObject $dataTransferObject)
+    {
+        if ($dataTransferObject->hasExistingContentBlock()) {
+            $dataTransferObject->getContentBlockEntity()->status = Status::archived();
+
+            return self::create($dataTransferObject);
```
carakas: you repeat what is done below so you can just have the ifstatement and use one return here

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlockStatus.php
```diff
@@ -1,68 +0,0 @@
-<?php
```
carakas: can you rebase and actually remove these files in the same commit as you create the new class so that git sees it as a move?

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlockStatusDBALType.php
```diff
@@ -1,32 +0,0 @@
-<?php
-
-namespace Backend\Modules\ContentBlocks\DBALType;
-
-use Backend\Modules\ContentBlocks\ValueObject\ContentBlockStatus;
-use Doctrine\DBAL\Types\Type;
-use Doctrine\DBAL\Platforms\AbstractPlatform;
-
-class ContentBlockStatusDBALType extends Type
```
carakas: idem

src/Backend/Modules/ContentBlocks/Installer/Installer.php
```diff
@@ -35,8 +35,8 @@ private function configureBackendNavigation(): void
         $this->setNavigation(
             $navigationModulesId,
             $this->getModule(),
-            'content_blocks/index',
-            ['content_blocks/add', 'content_blocks/edit']
+            'content_blocks/content_block_index',
```
carakas: I would revert this

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlockStatus.php
carakas: @jeroendesloovere this shows as outdated but I need you to look at it :p

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlockDataTransferObject.php
jeroendesloovere: Nice catch

jeroendesloovere:
@carakas thanks for your feedback. I've fixed it.

jeroendesloovere:
Except merging forkcms 5.0.0 dev

jeroendesloovere:
@carakas, can I rename the frontend action "Detail" to "ContentBlock"?

carakas:
you can add a rename for the widget if you also provide the update query and info in the upgrade guide

PR #2100, opened 2017-06-15
Add missing translations for extensions module

jonasdekeukelaere:
## Type
- Enhancement

## Pull request description
Will add some missing translations for extensions module. 
Note I only translated the ones I know for sure. I didn't want to use Google Translate because that can be a bit wonky.


carakas:
@jonasdekeukelaere can you maybe check if this is an issue on master as well?
And if it is create the pr targetting master instead?

jonasdekeukelaere:
I'll add it to the to do list

jonasdekeukelaere:
@carakas Looks ok on master. I think it is a thing on 5.0.0-dev because of the new breadcrumb system. (but not sure about that)

PR #2102, opened 2017-06-15
Frontend js language is lowercase

jonasdekeukelaere:
## Type
- Critical bugfix

## Pull request description
The frontend js language is lowercase, as defined in https://github.com/forkcms/forkcms/blob/5.0.0-dev/src/Frontend/Core/Header/Header.php#L105 - this gives an JS-error in the frontend, which breaks the other JS.


carakas:
it should be LANGUAGE , can you create a PR to fix it in the jsData?

PR #2104, opened 2017-06-15
Fix blog image path

bart-webleads:
## Type
- Non critical bugfix

## Pull request description
Blog post items did not return the `image_AAxBB` fields.
As you can see in `src/Backend/Modules/Blog/Actions/Add.php` the image path is in lowercase, not CamelCase.

### DIFF
--- a/src/Frontend/Modules/Blog/Engine/Model.php
+++ b/src/Frontend/Modules/Blog/Engine/Model.php
@@ -57,7 +57,7 @@ class Model implements FrontendTagsInterface
 
         // image?
         if (isset($return['image'])) {
-            $folders = FrontendModel::getThumbnailFolders(FRONTEND_FILES_PATH . '/Blog/Images', true);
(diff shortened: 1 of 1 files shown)


jeroendesloovere:
General discussion for Fork CMS 5:
Which folder name rules are we going to use, CamelCase or lowercased?
schermafbeelding 2017-06-15 om 10 56 39

carakas:
@jeroendesloovere for fork 5 you should create an issue, this is for fork 4

bart-webleads:
Forkcms uses lowercase.

But modules from others seem to be using CamelCase.

carakas:
the lowercase is a left over from when the module names where lowercase
I personally think the folder name should match the casing of the module

jeroendesloovere:
Discussion issue for fork cms 5 created: https://github.com/forkcms/forkcms/issues/2105

PR #2110, opened 2017-06-15
Usertemplates title and description

StijnVrolijk:
## Type

- Enhancement

## Pull request description

Displays the title and description of a usertemplate in the page template. This allows users to see more easily which usertemplates they have in the page.


carakas:
If there is no serialised data the entry in the database should be null and not ''

StijnVrolijk:
Status update: I'm currently battling a SpoonForm bug. Spoon somehow applies htmlentities if htmlentities are already present.. It's a really weird bug.

carakas:
@StijnVrolijk any idea when you can update this one?

jeroendesloovere:
@StijnVrolijk: were you able to fix you problems with spoon? Does it has something to do with your code using `htmlentities(...)` on two places?
F.e.:
```
$block['formElements']['hidExtraType'] = $this->frm->addHidden(
    'block_extra_data_' . $block['index'],
    htmlentities($block['extra_data'])
);
```

StijnVrolijk:
@jeroendesloovere yes it does. If I don't encode them, they don't get parsed properly because Spoon parses it as `value="{"title":"test"}"` which obviously doesn't work. Yet if I do encode them like it currently is in the code, it parses it like this: `value="{&amp;quot;title&amp;quot;:&amp;quot;test&amp;quot;}"`

carakas:
if needed you can make a PR to spoon library to fix this
It can be BC breaking since there are already some things ready that break BC

StijnVrolijk:
@carakas I've had a look already but couldn't really find anything. I'll take another look though

carakas:
@StijnVrolijk can you rebase on the current 5.0.0-dev branch? might fix the tests

Issue #1948, opened 2017-01-19
Mailmotor: integrate a ping method to check apiKey and ListId


jeroendesloovere:
## Type

- Enhancement

## Expected behavior / Proposed solutions

> Mailmotor: integrate a ping method to check apiKey and ListId
https://github.com/mailmotor/mailmotor-bundle/issues/8

StijnVrolijk:
This will be fixed in the mailmotor itself https://github.com/mailmotor/mailmotor-bundle/issues/8

jeroendesloovere:
Closing this issue was actually wrong.

I'm reopening this issue again,
because there will need to be a new "ApiKeyAndPageId" Validator class in the Backend Mailmotor module which will indeed must use a new function in the mailmotor-bundle.

jeroendesloovere:
Closing since the PR #2113 to fix this is merged

PR #2113, opened 2017-06-20
Mailmotor ping integration

jeroendesloovere:
## Type

- Enhancement
- Feature

> Fixes https://github.com/forkcms/forkcms/issues/1948

## Pull request description

This will fix the following:
- No more false "API-key" and "List-ID", since the new ping method in MailMotorBundle will check if your credentials are valid when you are submitting them in the backend Settings Action. If not valid it will reset everything (to prevent the (un)subscribe forms from not working).

And improve the following in code:
- We are now using a method from MailMotorBundle that loads in the available "mail-engine" services (summary: moved logic to bundle).


src/Backend/Modules/Mailmotor/Actions/Settings.php
```diff
@@ -52,9 +52,11 @@ public function execute(): void
             new SettingsSavedEvent($settings)
         );
 
+        $redirectAction = ($settings->mailEngine === 'not_implemented') ? 'Settings' : 'Ping';
```
carakas: you don't need the ()

jeroendesloovere: Fixed by #7bd908d

jeroendesloovere:
@carakas can you merge?

PR #2115, opened 2017-06-24
Add debouncer to generatedUrl

jessedobbelaere:
## Type

- Enhancement
- Bugfix

When filling in a title or name in a module, per character you type an ajax call gets fired away. In general, such keyevents need a debouncer so it waits a few hundred ms before executing. If you type a word, it'll only fire one ajax event. It's a general performance boost. If you type a long word, it can easily fire off 20-30 ajax calls (=and database calls most of the time). Using a debounce it'll do it when you stop typing

## Pull request description

Added a debouncer function to utils, and using it in the backend meta javascript. 
We could also start using http://underscorejs.org/ which contains a _.debounce function... Then we don't need it in utils. 
http://underscorejs.org/#debounce

Also, I snuck in the input event to catch additional changes like copy/paste in addition to keystrokes (in modern browsers)

### DIFF
--- a/src/Backend/Core/Js/jquery/jquery.backend.js
+++ b/src/Backend/Core/Js/jquery/jquery.backend.js
@@ -49,7 +49,7 @@
             var $urlOverwrite = $(options.urlOverwriteSelector);
 
             // bind keypress
-            $element.bind('keyup', calculateMeta);
+            $element.bind('keyup input', utils.events.debounce(calculateMeta, 400));
 
             // bind change on the checkboxes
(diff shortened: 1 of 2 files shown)


jessedobbelaere:
Should I point this to v5 or do we merge it in both branches?

carakas:
bugfixes from master get merged into 5
not the other way around so this is good I think

carakas:
I think this can be added in a bugfix release since it will fix the issue you sometimes have when you type a lot that you get an older version of the slug because that request was slower

PR #2116, opened 2017-06-24
Add missing generated_url_selector

jessedobbelaere:
## Type

- Non critical bugfix

When you have a generated url link on your module, it should create the url while you are typing in a name. The default value is #generatedUrl but if you look in the page source code, you'll see that the value is not set, so the selector isn't used in javascript. 

## Pull request description
Adding the generated url selector to the view variables, so it's not an empty string anymore and is set to the default value or the value you passed to the metatype. 

This was already fixed in v5 I believe, but just adding it back into master now


carakas:
This has already been merged in 5.0.0. Thanks for backporting the fix @jessedobbelaere

PR #2118, opened 2017-06-26
Allow null database values for images

jessedobbelaere:
## Type
- Non critical bugfix

Using an ImageDbalType and the underlying AbstractImageType and AbstractFile, it's impossible to store NULL to the database when the user doesn't upload an image in his module. Should be fixed with this.

## Pull request description
* When a user doesnt upload an image in the module, my entity object contains null for my image field, and when it's persisted, it goes through `convertToDatabaseValue`. In there, the null is typecasted to a string ==> end result is `""` in the database (empty string). Not sure if that's intended? 

To fix it we have to do a bit of ugly null checking, but otherwise we're storing empty strings in the database instead of null. Harder to query to find users without image etc if you also need to check for empty strings in those fields instead of `WHERE image IS NULL`....

Let me know if it can be solved more elegantly

### DIFF
--- a/src/Common/Doctrine/Type/AbstractFileType.php
+++ b/src/Common/Doctrine/Type/AbstractFileType.php
@@ -25,12 +25,12 @@ abstract class AbstractFileType extends Type
     }
 
     /**
-     * @param AbstractFile $file
+     * @param AbstractFile|null $file
      * @param AbstractPlatform $platform
      *
      * @return string|null
      */
-    public function convertToDatabaseValue($file, AbstractPlatform $platform): ?string
+    public function convertToDatabaseValue(?$file, AbstractPlatform $platform): ?string
     {
         return $file !== null ? (string) $file : null;
     }
--- a/src/Common/Doctrine/Type/AbstractFileType.php
+++ b/src/Common/Doctrine/Type/AbstractFileType.php
@@ -30,7 +30,7 @@ abstract class AbstractFileType extends Type
      *
      * @return string|null
      */
-    public function convertToDatabaseValue(?$file, AbstractPlatform $platform): ?string
+    public function convertToDatabaseValue($file, AbstractPlatform $platform): ?string
     {
         return $file !== null ? (string) $file : null;
     }
--- a/src/Common/Doctrine/Type/AbstractFileType.php
+++ b/src/Common/Doctrine/Type/AbstractFileType.php
@@ -28,11 +28,11 @@ abstract class AbstractFileType extends Type
      * @param AbstractFile $file
      * @param AbstractPlatform $platform
      *
-     * @return string
+     * @return string|null
      */
-    public function convertToDatabaseValue($file, AbstractPlatform $platform): string
+    public function convertToDatabaseValue($file, AbstractPlatform $platform): ?string
     {
-        return (string) $file;
+        return $file ? (string) $file : null;
     }
 
     abstract protected function createFromString(string $fileName): AbstractFile;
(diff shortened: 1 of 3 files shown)


src/Common/Doctrine/ValueObject/AbstractFile.php
```diff
@@ -201,9 +201,9 @@ public function __toString(): string
         return (string) $this->fileName;
     }
 
-    public static function fromString(string $fileName): self
+    public static function fromString(string $fileName): ?self
```
carakas: doesn't this need to be `?string`  then?

src/Common/Doctrine/Type/AbstractImageType.php
```diff
@@ -28,11 +28,11 @@ public function convertToPHPValue($imageFileName, AbstractPlatform $platform): A
...
      * @param AbstractPlatform $platform
      *
-     * @return string
+     * @return string|null
      */
-    public function convertToDatabaseValue($image, AbstractPlatform $platform): string
+    public function convertToDatabaseValue($image, AbstractPlatform $platform): ?string
     {
-        return (string) $image;
+        return $image ? (string) $image : null;
```
carakas: maybe make the checks a bit more explicit

src/Common/Doctrine/Type/AbstractFileType.php
```diff
@@ -28,11 +28,11 @@ public function convertToPHPValue($fileName, AbstractPlatform $platform): Abstra
...
      * @param AbstractPlatform $platform
      *
-     * @return string
+     * @return string|null
      */
-    public function convertToDatabaseValue($file, AbstractPlatform $platform): string
+    public function convertToDatabaseValue($file, AbstractPlatform $platform): ?string
     {
-        return (string) $file;
+        return $file ? (string) $file : null;
```
carakas: maybe make the checks a bit more explicit

src/Common/Doctrine/ValueObject/AbstractFile.php
jessedobbelaere: Yeah thanks! I'm using master in my project and that doesn't have the php7 typehints yet so that one went unnoticed

src/Common/Doctrine/Type/AbstractFileType.php
```diff
@@ -25,14 +25,14 @@ public function convertToPHPValue($fileName, AbstractPlatform $platform): Abstra
...
     /**
-     * @param AbstractFile $file
+     * @param AbstractFile|null $file
      * @param AbstractPlatform $platform
      *
-     * @return string
+     * @return string|null
      */
-    public function convertToDatabaseValue($file, AbstractPlatform $platform): string
+    public function convertToDatabaseValue(?$file, AbstractPlatform $platform): ?string
```
carakas: this doesn't work :p ?$file

src/Common/Doctrine/Type/AbstractImageType.php
```diff
@@ -25,14 +25,14 @@ public function convertToPHPValue($imageFileName, AbstractPlatform $platform): A
...
     /**
-     * @param AbstractImage $image
+     * @param AbstractImage|null $image
      * @param AbstractPlatform $platform
      *
-     * @return string
+     * @return string|null
      */
-    public function convertToDatabaseValue($image, AbstractPlatform $platform): string
+    public function convertToDatabaseValue(?$image, AbstractPlatform $platform): ?string
```
carakas: this does'nt work :p no tyêhints here because the interface doesn't have them

jessedobbelaere: Lol yeah removed it again then :)

PR #2121, opened 2017-06-28
Fix ckfinder not finding our AppKernel and KernelLoader

carakas:
## Type

- Non critical bugfix

## Pull request description

Ck finder didn't work anymore because it couldn't load the Appkernel and KernelLoader classes


carakas:
I had separated the code styles and the actual fix just because of that whitespace :p

jessedobbelaere:
Too used to dirty commits, i should've checked commit per commit :P thanks for the fix!

PR #2123, opened 2017-07-03
Use symfony http code constants

carakas:
## Type

- Enhancement

## Pull request description

This removes our own constants for status codes in favour of those of the Symfony Response class


jeroendesloovere:
:+1::+1::+1: Thumbs up for @carakas

PR #2124, opened 2017-07-04
Profiles: fix avatar path in installer

bart-webleads:
## Type

- Non critical bugfix

## Pull request description

The profiles avatar path should be in CamelCase. The rest of the module seems to be using "Profiles/Avatars".


jeroendesloovere:
@carakas Can I ask why you merged this. I thought that fork cms master was to be snakecase and fork 5.0.0 CamelCase for all...?

carakas:
because he was correct, in the code we used the CamelCased and in the installer this one.
Since the camel cased version is expected in the code it makes sense to correct this.
in fork 4 there wasn't a real rule, in fork 5 it should me CamelCased

PR #2133, opened 2017-07-11
Fix mailmotor when not installed [5.0.0-dev branch]

### LINKED ISSUE #2131: When installing Fork CMS without installing mailmotor, mailmotor.mail_engine should be set
## Type

- Critical bug

## Problem description

> When installing "Fork CMS" and not installing "mailmotor module". `mailmotor.mail_engine` is missing. It should be set to `not_implemented` instead of `null`. This comes because the installer doesn't go into the `src/Backend/Modules/Mailmotor/DependencyInjection/Compiler/MailmotorCompilerPass.php`, where this is set...

## Steps to reproduce

1. Install fork cms
2. Don't install mailmotor
=> Error will be shown after installation and breaks everything.

> `mailmotor.mail_engine` should be set to `not_implemented`.

jeroendesloovere:
## Type

- Critical bugfix

> Fixes #2131 in 5.0.0 dev branch

## Pull request description

> Fixes the installation of Fork CMS when Mailmotor isn't installed


carakas:
this is already in master, no need for a separate PR, I'll merge master later again in the 5.0.0-dev branch

PR #2136, opened 2017-07-13
fix the double serialising of the user settings

carakas:
## Type

- Critical bugfix

## Pull request description
The usersettings are dubble serialised at the moment.
I removed the serialising in the installer class since it already happens in the model

### DIFF
--- a/src/Backend/Modules/Users/Installer/Installer.php
+++ b/src/Backend/Modules/Users/Installer/Installer.php
@@ -150,20 +150,18 @@ class Installer extends ModuleInstaller
 
             // build settings
             $settings = [];
-            $settings['nickname'] = serialize('Fork CMS');
-            $settings['name'] = serialize('Fork');
-            $settings['surname'] = serialize('CMS');
-            $settings['interface_language'] = serialize($this->getVariable('default_interface_language'));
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Users/Installer/Installer.php
```diff
@@ -150,20 +150,18 @@ private function loadGodUser(): void
...
-            $settings['password_strength'] = serialize($this->getPasswordStrength());
-            $settings['current_password_change'] = serialize(time());
-            $settings['avatar'] = serialize('god.jpg');
+            $settings['nickname'] = 'Fork CMS';
+            $settings['name'] = 'Fork';
+            $settings['surname'] = 'CMS';
+            $settings['interface_language'] = $this->getVariable('default_interface_language');
+            $settings['date_format'] = 'j F Y';
+            $settings['time_format'] = 'H:i';
+            $settings['datetime_format'] = $settings['date_format'] . ' ' . $settings['time_format'];
```
jeroendesloovere: Is this correct? You removed serialize, but also unserialize?

carakas: The serialisation happens in the model. It unserialized here because a couple of lines above this, it was created and serialised

PR #2139, opened 2017-07-20
2111 prepare database for fork 5

### LINKED ISSUE #2111: Check if there are database changes that need to be made
## Type

- Enhancement

## Problem description
A lot of the tables have the option to add serialised data to it, it is handy for in situations where we want to add something in a backwards compatible way. But maybe some of those things should be placed in a separate field instead. So we should go over the database and check that

carakas:
## Type

- Enhancement

closes #2111 

## Pull request description

bye bye enums and Y/N as booleans


jonasdekeukelaere:
@carakas The tests fail.

carakas:
@jonasdekeukelaere I know, I'm working on it

UPGRADE_5.0.md
```diff
@@ -284,3 +284,12 @@ Backend\Core\Engine\Base\AjaxAction and Frontend\Core\Engine\Base\AjaxAction no
...
 
 The response class has many more constants like this for all the other status codes
+
+## Bye bye enums
+
+Because doctrine doesn't support enums out of the box and adding, removing, or changing an enum value is a lot of work we changed them all to varchars
+
+We made the changes in de code and you can find the migration queries below
+
+- `ALTER TABLE blog_comments MODIFY type VARCHAR(255) NOT NULL default 'comment';`
```
jonasdekeukelaere: in theory this (the length of the field) should only be the length of the longest word possible

UPGRADE_5.0.md
```diff
@@ -293,3 +293,6 @@ We made the changes in de code and you can find the migration queries below
 
 - `ALTER TABLE blog_comments MODIFY type VARCHAR(255) NOT NULL default 'comment';`
 - `ALTER TABLE blog_comments MODIFY status VARCHAR(249) NOT NULL default 'moderation';` (we cant do 255 because that is too big for the index)
+- `ALTER TABLE blog_posts MODIFY status VARCHAR(244) NOT NULL;` (we cant do 255 because that is too big for the index)
+- `ALTER TABLE blog_posts MODIFY hidden VARCHAR(1) NOT NULL default 'N';`
```
jonasdekeukelaere: wouldn't it be better to change Y/N to doctrine bool type (tinyint 0/1)?

src/Backend/Modules/ContentBlocks/Domain/ContentBlock/StatusDBALType.php
```diff
@@ -2,19 +2,13 @@
 
 namespace Backend\Modules\ContentBlocks\Domain\ContentBlock;
 
-use Backend\Modules\ContentBlocks\Domain\ContentBlock\Status;
-use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Types\StringType;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 
-class StatusDBALType extends Type
+class StatusDBALType extends StringType
```
jonasdekeukelaere: So this can be BooleanType? (see other comment about Y/N enum)

UPGRADE_5.0.md
```diff
@@ -422,3 +422,29 @@ ALTER TABLE users MODIFY active VARCHAR(1) NOT NULL DEFAULT 'Y' COMMENT 'is this
 ALTER TABLE users MODIFY deleted VARCHAR(1) NOT NULL DEFAULT 'N' COMMENT 'is the user deleted?';
 ALTER TABLE users MODIFY is_god VARCHAR(1) NOT NULL DEFAULT 'N';
 ```
+
+## SEOIndex and SEOFollow are now fields in the meta table
```
jonasdekeukelaere: 👍

UPGRADE_5.0.md
```diff
@@ -450,3 +450,58 @@ WHERE data IS NOT NULL AND data LIKE '%seo_follow";s:4:"none"%';
 ```
 
 In the installer the method `insertMeta` has a new signature so we can add the seo_follow and seo_index
+
+## Move from Y and N to true booleans
```
jonasdekeukelaere: 👍

UPGRADE_5.0.md
jonasdekeukelaere: Nvm, has been done later

carakas: I did it in two steps to ease the migration

UPGRADE_5.0.md
carakas: true, but if we limit the length here then we need to give up the freedom of adding types that have a longer name without database changes

jonasdekeukelaere: When we'll be using Doctrine, I thought it did this automatically.

carakas: Nope, not automatically, because in that case it would need to run structural updates each time a longer entry is added

jonasdekeukelaere: Hmm my mistake

jeroendesloovere:
@carakas, you can enable "enum" in doctrine with the following code:
```
# Doctrine Configuration
doctrine:
    dbal:
        default_connection:       default
        connections:
            default:
                #...
                # required, because enums aren't understand by default
                mapping_types:
                    enum: string
```

Did you know this?

carakas:
I know, but that is a hack, so all enums must die

jeroendesloovere:
👍 👍 👍 OK, good job!

PR #2141, opened 2017-07-24
Add the test command to composer

carakas:
## Type

- Enhancement

## Pull request description

Add the test command as a composer script

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,12 @@ Because we support multiple php versions it gave some issues. Therefore we use t
 
 Running the tests:
 
-    ./bin/simple-phpunit
+    composer test
 
 Running only the unit tests or the functional tests
 
-     ./bin/simple-phpunit --testsuite=functional
-     ./bin/simple-phpunit --testsuite=unit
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ matrix:
 
 script:
   - bin/phpcs --standard=psr2 --extensions=php --warning-severity=0 --ignore=src/Backend/Core/Js/ckfinder,src/Backend/Cache,src/Frontend/Cache,src/Backend/Core/Js/ckeditor --report=full "src"
-  - travis_wait 20 bin/simple-phpunit --coverage-clover=coverage.clover
+  - travis_wait 20 bin/simple-phpunit test -- --coverage-clover=coverage.clover
```
jessedobbelaere: Can travis also use composer test or do we keep bin/simple-phpunit here?

carakas: no, I tried, that test is a thing I forgot to remove
The travis_wait doesn't go well together with the composer test

PR #2143, opened 2017-07-25
2105 camel cased files folder names

### LINKED ISSUE #2105: CamelCased files folder names
## Type

- Enhancement

## Problem description

> In Fork CMS 5 we have a mixture of CamelCased and lowercased files folder names. We should uniform them.

Which folder name rules are we going to use, CamelCased or lowercased?
=> I prefer going to CamelCased, so they match the CamelCased module names.
schermafbeelding 2017-06-15 om 10 56 39

carakas:
## Type

- Enhancement

fixes #2105 

## Pull request description

By matching the folder name inside the Files directory it will be more clear to what module it belongs
If it already had the same name but a different casing I also updated it

I left the userfiles directory as is for now, but I'm not sure about that. It is used by ckfinder and during the import in the blog from wordpress. Should we leave it like this or move it to something like `Core/CKFinder`


tijsverkoyen:
I would rename userfiles to Core/CKEditor

carakas:
@tijsverkoyen why CKEditor and not CKFinder?

tijsverkoyen:
Or /Core/CKFinder

PR #2144, opened 2017-07-26
587 switching themes

### LINKED ISSUE #587: Switching themes
It's not possible to change your theme without a mouse action. The themes should be able to be focussed by keyboard.

mathiashelin:
## Type

- Enhancement

fixes #587 

## Pull request description

Themes can now be switched by using the keyboard only.


StijnVrolijk:
Wouldn't this also be a good time to convert this page/js to use `data-role` instead of `class="js-foo-bar"`?

Other than that, this gets a thumbs up from me 👍

PR #2146, opened 2017-07-27
Remove ping_services from Blog module

ohvitorino:
## Type

- Non critical bugfix

`ping_services` form field is still being used in the Blog module Add action.

## Pull request description

This PR simply remove the usage of the `ping_services` field and its surrounding SEO panel.


ohvitorino:
@carakas there is also a usage of ping_services in `test_db.sql`, probably also not necessary. Other than that I don't find other.

ohvitorino:
`ping_services` is being added to the module_settings in `test_db.sql`. Should I remove this one as well?

carakas:
you can remove that one indeed @ohvitorino

ohvitorino:
@carakas f09d18172 remove the insert statement from `test_db.sql`.

PR #2148, opened 2017-07-31
We can't be sure the json field isn't null

StijnVrolijk:
@ohvitorino found out the hard way, pages have `NULL` by default in extra_data, the code wasn't prepared for this.

### DIFF
--- a/src/Backend/Modules/Pages/Actions/Edit.php
+++ b/src/Backend/Modules/Pages/Actions/Edit.php
@@ -863,7 +863,7 @@ class Edit extends BackendBaseActionEdit
         $this->template->assign('deleteForm', $deleteForm->createView());
     }
 
-    private function getHiddenJsonField(string $name, ?string $json = null): SpoonFormHidden
+    private function getHiddenJsonField(string $name, ?string $json): SpoonFormHidden
     {
         return new class($name, htmlspecialchars($json)) extends SpoonFormHidden {
             public function getValue($allowHTML = null)
--- a/src/Backend/Modules/Pages/Actions/Edit.php
+++ b/src/Backend/Modules/Pages/Actions/Edit.php
@@ -863,7 +863,7 @@ class Edit extends BackendBaseActionEdit
         $this->template->assign('deleteForm', $deleteForm->createView());
     }
 
-    private function getHiddenJsonField(string $name, string $json): SpoonFormHidden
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -863,7 +863,7 @@ private function loadDeleteForm(): void
         $this->template->assign('deleteForm', $deleteForm->createView());
     }
 
-    private function getHiddenJsonField(string $name, string $json): SpoonFormHidden
+    private function getHiddenJsonField(string $name, ?string $json = null): SpoonFormHidden
```
carakas: doesn't this method also exist in the add action, also, you don't need both ? and = null
= null will automagically make it accept null as it is the default and with ?string you do need to pass something

StijnVrolijk: I feel the `= null` provides some extra clarity even though it's not 100% necessary.

The add message also has this method but the json is never null, since you add the blocks yourself in which case the json is filled

carakas: Meh, really not a fan of it, looks to me like the dev doesn't know how it works

ohvitorino: I think the question is: should the second parameter be optional or not?

carakas: If it is the first ? doesn't need to be there, if it isn't the = null shouldn't be there, that is what I meant with the comment above

StijnVrolijk: The method always gets called with 2 param, the 2nd parameter is just `null` sometimes

StijnVrolijk: But why? The `= null` is just there to provide some extra information. It doesn't matter it's the default and the developer should 'know how it works'. If that's the way you think about it, I'll make a PR removing all `public` keywords from functions since it's the default and developers should just know that. Sounds like some visual debt bs to me

carakas: you said that you always need to pass 2 parameters, and that doesn't match your current implementation

StijnVrolijk: i never said that... It gets called with 2 params, it doesn't need to. Whatever I'll change it

PR #2158, opened 2017-08-03
Https

jonasdekeukelaere:
## Type
- Enhancement

## Pull request description
Add https entry in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,11 @@ To install the dependencies, you can run the command below in the document-root:
 
 If you discover any security related issues, please email core@fork-cms.com instead of using the issue tracker.
 
+
+## SSL
+
+If you want to activate https redirection, go into the .htaccess file and uncomment the lines about https.
+
 ## Bugs
 
 If you encounter any bugs, please create an issue on [Github](https://github.com/forkcms/forkcms/issues).


.htaccess
```diff
@@ -69,6 +69,8 @@ deny from env=stayout
 
 	# redirect all trafic to https
 	# Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
+	# Header always edit Set-Cookie (.*) "$1; HTTPOnly; Secure"
+	# Header set Strict-Transport-Security: "max-age=63072000"
```
carakas: Doesn't this conflict with `Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS`at line 71?

.htaccess
```diff
@@ -69,6 +69,8 @@ deny from env=stayout
 
 	# redirect all trafic to https
 	# Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
+	# Header always edit Set-Cookie (.*) "$1; HTTPOnly; Secure"
```
carakas: This is the default when you set cookies with fork, but you can choose other options there, won't this prevent people from setting the cookies the way they want?

.htaccess
jonasdekeukelaere: My mistake, did a check for the first addition, never checked for the second one.

jonasdekeukelaere: Fixed 2726cdd43

.htaccess
jonasdekeukelaere: Removed it.

PR #2164, opened 2017-08-04
It is now saved as a boolean

freshface:
## Type

- Critical bugfix

The check is wrong because it is not stored as a Y/N value


carakas:
There are 2 PR's for the same issue but the implementation of #2179  is slightly better so we will merge that one. Thank you for your PR.

PR #2173, opened 2017-08-09
Assign the fork settings in the frontend twig template

### LINKED ISSUE #2163: THEME_URL is not set
## Type

- Critical bug

## Problem description

THEME_URL is not set because `$this->forkSettings` is `null` https://github.com/forkcms/forkcms/blob/master/src/Common/Core/Twig/BaseTwigTemplate.php#L142

## Steps to reproduce

Do this in your theme template: `{{ dump(THEME_URL) }}`

StijnVrolijk:
## Type

- Critical bugfix

## Issue

Fixes #2163

## Pull request description

During a cleanup of the twig templates this assignment was accidentally removed. This is needed to access stuff like the theme url, site title, ... in the frontend templates


src/Frontend/Core/Engine/TwigTemplate.php
```diff
@@ -39,6 +39,7 @@ public function __construct(
         parent::__construct($environment, $parser, $locator);
 
         $container = Model::getContainer();
+        $this->forkSettings = $container->get('fork.settings');
```
carakas: Can you then also use $this->forkSettings at the other places where we get it from the container in this class?

carakas:
@StijnVrolijk can you check my fixes?

StijnVrolijk:
@carakas it's a bit dirty but if it works, it works! It's for tests anyway so yeah.. :shipit:

PR #2174, opened 2017-08-09
Switch the upload and download icons to match the import button

StijnVrolijk:
## Type

- Non critical bugfix

## Pull request description

The import button in the locale has a download icon (http://demo.fork-cms.com/private/en/locale/import?token=ri1dar0dm&type%5B0%5D=lbl&language%5B0%5D=en) but the one on the index has an upload icon. I think the download icon (while not in name) makes more sense than the upload icon so I changed it to match.

### DIFF
--- a/src/Backend/Modules/Locale/Layout/Templates/Import.html.twig
+++ b/src/Backend/Modules/Locale/Layout/Templates/Import.html.twig
@@ -30,7 +30,7 @@
       <div class="btn-toolbar">
         <div class="btn-group pull-right" role="group">
           <button id="importButton" type="submit" name="add" class="btn btn-primary">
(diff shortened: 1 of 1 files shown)


mathiashelin:
I'd change the one on http://demo.fork-cms.com/private/en/locale/import?token=vywzszkj0a&type%5B0%5D=lbl&language%5B0%5D=en

Import = upload a file
Export = download a file

StijnVrolijk:
@mathiashelin I know but the icons themselves don't make sense to me. It shows an arrow leading _away_ from the server-looking thingy to indicate you're putting something _in_ the server? The only reason it makes sense is the upload arrow points up and the download array points down

StijnVrolijk:
@mathiashelin I've changed the import button on the import action itself

mathiashelin:
:shipit:

PR #2175, opened 2017-08-10
Cache mailmotor interests

jonasdekeukelaere:
## Type
- Critical bugfix

## Pull request description
Widgets shouldn't make API calls on every single request. So cache the interests call for mailmotor.

### DIFF
--- a/src/Frontend/Modules/Mailmotor/Domain/Subscription/SubscribeType.php
+++ b/src/Frontend/Modules/Mailmotor/Domain/Subscription/SubscribeType.php
@@ -10,6 +10,7 @@ namespace Frontend\Modules\Mailmotor\Domain\Subscription;
  */
 
 use Common\ModulesSettings;
+use DateTime;
 use Frontend\Core\Engine\Navigation;
 use Frontend\Modules\Mailmotor\Domain\Subscription\Command\Subscription;
 use MailMotor\Bundle\MailMotorBundle\Exception\NotImplementedException;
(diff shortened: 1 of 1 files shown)


src/Frontend/Modules/Mailmotor/Domain/Subscription/SubscribeType.php
```diff
@@ -91,7 +92,24 @@ public function getInterests(): array
     {
         $interests = [];
 
+        // should we be checking interests (CampaignMonitor for example has no interests)
+        $mailMotorInterestsNoCheck = $this->modulesSettings->get('Mailmotor', 'interests_no_check', false);
```
tijsverkoyen: Isn't it dangerous to set it to false by default?

tijsverkoyen: I also don't like negative boolean, I would name it "check_interests"

jonasdekeukelaere: the no check is for when we get a noimplemented exception from a subscriber gateway, so setting it to false when it is not filled in will always trigger the code to get the interests from the gateway

jonasdekeukelaere: Fixed f7cec962b

jonasdekeukelaere: Maybe f7cec962b will make that clearer

tijsverkoyen: 👍

jeroendesloovere:
Hi @jonasdekeukelaere 
Thanks for this PR.
Shouldn't there be a backend possibility to enable/disable this "check_interests" module setting?

jonasdekeukelaere:
I thought we established that settings was not the way to go? (re #general)

jeroendesloovere:
@jonasdekeukelaere Don't know if were are talking about the same thing here.

What I mean is the following: sometimes you want (or are required) to show interests on the subscription form. But that can be optionally.
So my suggestion is to add a backend setting "show interests on subscription form" (or not).

jonasdekeukelaere:
They are only showed if there are any present. As far as I see.

PR #2176, opened 2017-08-10
Specifically state the widget action should be 'Location'

StijnVrolijk:
## Type

- Critical bugfix

## Pull request description

A previous commit ( https://github.com/forkcms/forkcms/commit/8567d02acaa0acf975e5a5f114755420deba0403 ) caused the action of an extra type to be set to `null` if it wasn't specified. Well in the locations module it was still assumed it would take over the module's value so it received `null` as an action, causing it to be defaulted to `Index` which doesn't exist.

### DIFF
--- a/src/Backend/Modules/Location/Engine/Model.php
+++ b/src/Backend/Modules/Location/Engine/Model.php
@@ -217,6 +217,7 @@ class Model
         // insert extra
         $item['extra_id'] = BackendModel::insertExtra(
             ModuleExtraType::widget(),
+            'Location',
             'Location'
         );
 


src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -217,6 +217,7 @@ public static function insert(array $item): int
         // insert extra
         $item['extra_id'] = BackendModel::insertExtra(
             ModuleExtraType::widget(),
+            'Location',
```
carakas: This won't fix the already installed modules when upgrading, can you also update the config of the module so the default isn't index but has the correct value of Location?

StijnVrolijk: The Location module has an action called Index so I'm guessing this is the correct default. This means I can't change the default, so I had to change it here instead

PR #2185, opened 2017-08-16
getGroupsForDropdown expects an integer, we've sent null

StijnVrolijk:
## Type

- Critical bugfix

## Pull request description

There was a private variable `$id` which wasn't used in any way so it stayed null. We then used that variable to request the profile groups for a dropdown which expects an integer to be sent, PHP 7 does not like this. Therefor i've just sent `0` to the method so it works as it used to but the ideal solution would be to rewrite the method so it doesn't required a profile id.

### DIFF
--- a/src/Backend/Modules/Profiles/Actions/Import.php
+++ b/src/Backend/Modules/Profiles/Actions/Import.php
@@ -21,11 +21,6 @@ use Backend\Core\Engine\Csv;
  */
 class Import extends BackendBaseActionAdd
 {
-    /**
-     * @var int
-     */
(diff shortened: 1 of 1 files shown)


carakas:
You can also instead of rewriting that method just add another one

StijnVrolijk:
@carakas true but that would require an illogical method name such as `getGroupsForDropdownWithoutProfile` or something. `getGroupsForDropdown` is the most logical name

carakas:
@StijnVrolijk `getGroupsForImport`?

PR #2193, opened 2017-08-17
Update to CKFinder 3.4.1

StijnVrolijk:
## Type

- Critical bugfix

## Pull request description

The current CKFinder version is incompatible with PHP 7 which causes issues since Fork 5 you know... requires PHP 7. I've upgraded CKFinder and gave it a nice skin! Also included is a fix for making the Authentication compatible with Symfony 3.3 thanks to @cirykpopeye 

As you might notice, there are a lot of vendor files which may also be included in our own versions (albeit in different versions). This isn't very nice and a possible solution might be using CKFinder's own Symfony bundle https://github.com/ckfinder/ckfinder-symfony3-bundle although this forces us to place CKFinder configuration in our own configuration and writing some extra classes to make it work. The real kicker is the Symfony bundle basically just downloads CKFinder and plonks it somewhere, just like this pull request does. So the double dependencies issue still exists.

TL;DR: this version of CKFinder works with Fork 5 and PHP 7 but it needs a cleanup in the vendor department.

Oh yeah, almost forgot! Pretty pictures:
screen shot 2017-08-17 at 15 54 25
pasted image at 2017_08_17 03_54 pm


tijsverkoyen:
In my personal opinion: 
I would quick fix the current implementation, so it can be used again.
The proper solution is to ditch CKFinder and write a plugin for CKEditor that allows us to use the MediaLibrary.

carakas:
I'm working on cleaning this one up, I'll push asap

carakas:
has been cleaned up

PR #2195, opened 2017-08-18
Add missing Russian translations to ckfinder

linbis:
## Type
- Enhancement

## Pull request description

CKFinder Russion translate


StijnVrolijk:
Nice PR (aside from the 3 commits that should be 1 commit)

BUT!

We're currently in the process of updating CKFinder to 3.4.1 ( #2193 ) which includes a full russian translation

carakas:
Since this is your first PR ever I'm going to merge this one. Older versions of fork can always use this one as a reference to update the translations there.

@linbis  it would be amazing if you could also update/add some of the Russian translations for our modules

linbis:
@carakas no problem

PR #2199, opened 2017-08-22
Theme screen.css is loaded instead of a general css

### LINKED ISSUE #2188: screen.css Fork theme not loaded
## Type
Critical bug

## Problem description
In ealier versions you were able to add a `screen.css` to your theme and this would display theme specific css. In the new version it doesn't load the theme `screen.css` but a global.

## Steps to reproduce

Check version 4.5.4 which loads `screen.css` in the css array, 5.0.1 loads `/src/Frontend/Core/Layout/Css/screen.css`

I would expect `/src/Frontend/Themes/Fork/Core/Layout/Css/screen.css` instead of `/src/Frontend/Core/Layout/Css/screen.css`

jacob-v-dam:
## Type

- Non critical bugfix

fixes #2188 

## Pull request description

Theme screen.css is loaded instead of a general one.

### DIFF
--- a/src/Frontend/Core/Header/Header.php
+++ b/src/Frontend/Core/Header/Header.php
@@ -126,7 +126,12 @@ class Header extends KernelLoader
         $this->meta = new MetaCollection();
 
         // add some default CSS files
-        $this->addCSS('/src/Frontend/Core/Layout/Css/screen.css');
+        $themeCss = $container->getParameter('site.path_www') . '/src/Frontend/Themes/' . Theme::getTheme() . '/Core/Layout/Css/screen.css';
+        if(file_exists($themeCss)) {
+            $this->addCSS($themeCss);
+        } else {
+            $this->addCSS('/src/Frontend/Core/Layout/Css/screen.css');
+        }
 
         // debug stylesheet
         if ($container->getParameter('kernel.debug')) {


src/Frontend/Core/Header/Header.php
```diff
@@ -126,7 +126,7 @@ public function __construct(KernelInterface $kernel)
         $this->meta = new MetaCollection();
 
         // add some default CSS files
-        $this->addCSS('/src/Frontend/Core/Layout/Css/screen.css');
+        $this->addCSS('/src/Frontend/Themes/' . Theme::getTheme() . '/Core/Layout/Css/screen.css');
```
carakas: this will fail if the theme doesn't have a screen.css, with templates there is some code that checks if a file is available in a theme and that causes it to overwrite, maybe do something like that or check how it worked in 4, must have missed something

jacob-v-dam:
Aah noticed a big difference between the `/src/Frontend/Core/Layout/Css/screen.css` in 4 and 5. Screen.css in 4 does not contain any specific code:
```
/*
    Stylesheet for <>

    Media: screen, projection
    Copyright: Company <http://www.company.com>
    author: Name <name@domain.com>
*/

@media print {
    /*
        Stylesheet for <>

        Media: print
        Copyright: Company <http://www.company.com>
        author: Name <name@domain.com>
    */
}
```

While in version 5 it contains really specific code. Which is not wanted when you are developing your own theme.

src/Frontend/Core/Header/Header.php
jacob-v-dam: Aah noticed a big difference between the `/src/Frontend/Core/Layout/Css/screen.css` in 4 and 5. Screen.css in 4 does not contain any specific code:
```
/*
    Stylesheet for <>

    Media: screen, projection
    Copyright: Company <http://www.company.com>
    author: Name <name@domain.com>
*/

@media print {
    /*
        Stylesheet for <>

        Media: print
        Copyright: Company <http://www.company.com>
        author: Name <name@domain.com>
    */
}
```

While in version 5 it contains really specific code. Which is not wanted when you are developing your own theme.

carakas: with this you only scratch your itch, what would be better is a more generic solution, does this file exist in the current theme, ok, use that one, else fall back to the core

carakas:
@jacob-v-dam any update on this

carakas:
@jacob-v-dam did you already find some time to improve this one?

carakas:
You can overwrite this file by placing a file with the same name in the correct location in your theme.

Since there is already an implementation for this I'm going to close this pr

PR #2200, opened 2017-08-22
PHP version check added

jeroendesloovere:
## Type

- Enhancement

## Pull request description

> When installing Fork CMS using `composer create-project forkcms/forkcms .` on the command line. The PHP version there could be different then in your environment. This could happen unaware. Currently a white page is being shown, so you have totally no clue what's going on. That's why this PR gives you at least a decent warning.

### DIFF
--- a/index.php
+++ b/index.php
@@ -12,7 +12,7 @@ chdir(__DIR__);
 
 // PHP version check
 if (version_compare(phpversion(), '7.1.0', '<')) {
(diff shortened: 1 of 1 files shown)


WouterSioen:
if you add it here, you should also add it in the `bin/console` file, because you could face the same issues there.

But I'm not really in favour of this. If you add this check, you could/should add checks for all other dependencies too, which would be a mess.

carakas:
Going to close this one for the reasons @WouterSioen mentioned

PR #2205, opened 2017-08-24
FormLayout: added "help text" to a form text field

jeroendesloovere:
## Type

- Feature

## Pull request description

> Allows you to add "help text" to a form text field.

Adding the attribute help, will now create a ".help-block" class span just below the text field.
```
$builder->add(
    'name',
    TextType::class,
    [
        'label' => 'lbl.Naming',
        'attr' => [
            'help' => 'msg.NamingHelp',
        ],
    ]
)
```
Before:
schermafbeelding 2017-08-24 om 11 10 27

After:
schermafbeelding 2017-08-24 om 11 07 48

Updated: You can also use the following in your template:
```
{{ form_widget(form.name, {'attr.help': 'msg.NamingHelp'|trans}) }}
```
> Which will do the same thing.

### DIFF
--- a/src/Backend/Core/Layout/Templates/FormLayout.html.twig
+++ b/src/Backend/Core/Layout/Templates/FormLayout.html.twig
@@ -350,11 +350,7 @@
     </div>
     <div class="panel-body">
       {{- form_widget(form) -}}
-
-      {% for attributeName, attributeValue in attr %}
-        {% if attributeName == 'help' %}
-          <span class="help-block">{{ attributeValue|trans }}</span>{% endif %}
-      {% endfor %}
+      {{- block('help_text') -}}
     </div>
     {% if (not compound or force_error|default(false)) and not valid %}
       <div class="panel-footer">
@@ -378,4 +374,11 @@
       </div>
     {% endif %}
   </div>
+{% endblock %}
+
+{% block help_text %}
+  {% for attributeName, attributeValue in attr %}
+    {% if attributeName == 'help' %}
+      <span class="help-block">{{ attributeValue|trans }}</span>{% endif %}
+  {% endfor %}
 {% endblock %}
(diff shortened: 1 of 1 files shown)


jeroendesloovere:
Note: when using bootstrap 4, the code should be `<span class="form-text text-muted">...</span>`

jeroendesloovere:
Change added: you can now also use `{{ form_widget(form.name, {'help': 'msg.NamingHelp'|trans}) }}`

src/Backend/Core/Layout/Templates/FormLayout.html.twig
```diff
@@ -374,3 +380,13 @@
     {% endif %}
   </div>
 {% endblock %}
+
+{% block help %}
+  {% for attributeName, attributeValue in attr %}
+    {% if attributeName == 'help' %}<span class="help-block">{{ attributeValue|trans }}</span>{% endif %}
```
carakas: you can now have 2 help blocks by doing it like this if you set one from the template and one from the php can you fix this please, also if possible, avoid the duplicate html

jeroendesloovere: Yes indeed, that is possible.
Is that a problem?

carakas: why would you even allow that?
I don't mind that you can overwrite it from the template, but this is a very strange implementation

jeroendesloovere: So you prefer the "attr.help"? And not the variable "help"?

carakas: indeed

WouterSioen:
nice feature, I like

Issue #2092, opened 2017-06-13
Feature to easily copy to other locale.


jeroendesloovere:
## Type

- Feature

## Problem description

Fork CMS is a "multi-language" CMS.
Sometimes you need to add another language to the CMS and want to have everything **duplicated in the new language** by using `bin/console forkcms:locale:enable`.

Currently the following modules have "CopyToOtherLocale" methods/commands available:
* Pages
* ContentBlocks
> These are automatically executed when you run `bin/console forkcms:locale:enable`.

But in other (custom) modules you have to create such feature yourself and you have to execute them yourself.

## Expected behavior / Proposed solutions

What if this could be automated very easily *by adding the following method in the module installer*?!?
```php
$this->addModule('ContentBlocks');
$this->addCopyCommand(CopyContentBlocksToOtherLocale::class);
```
[Edited]: the above can be replaced by a `command bus handler` using a specific "symfony tag".

In Fork CMS we are working towards a `Domain`-base structure with `Commands`.
I guess if we have a Core `CopyToOtherLocale` command that we can extend, things will be awesome.
```php
<?php

namespace Backend\Core\Command\Base;

use Backend\Core\Language\Locale;

final class CopyToOtherLocale
{
    /** @var Locale */
    public $toLocale;

    /** @var Locale */
    public $fromLocale;

    public function __construct(Locale $toLocale, Locale $fromLocale = null)
    {
        if ($fromLocale === null) {
            $fromLocale = Locale::workingLocale();
        }

        $this->toLocale = $toLocale;
        $this->fromLocale = $fromLocale;
    }
}
```

```php
<?php

namespace Backend\Modules\ContentBlocks\Domain\ContentBlock\Command;

use Backend\Core\Command\Base\CopyToOtherLocale;

final class CopyContentBlocksToOtherLocale extends CopyToOtherLocale
{
}
```

Seems like something that can be integrated easily.

Some thoughts?

jeroendesloovere:
## Another solution

Using `service tags` to get all available "copy commands".

PR #2207, opened 2017-08-24
Meta cloning

jeroendesloovere:
## Type

- Feature

## Pull request description

> Allows us to create "clone/copy" methods on other entities and duplicate the meta.
I've used this in a "Categories module" "CopyCategoriesToOtherLocaleHandler".

Behavior:
* $id is set to null, so when the command bus executes everything a new meta id will be given
* $url creates an unique url


src/Common/Doctrine/Entity/Meta.php
```diff
@@ -167,6 +168,15 @@ public function update(
         $this->seoIndex = $seoIndex;
     }
 
+    public function duplicate()
```
carakas: https://secure.php.net/manual/en/language.oop5.cloning.php

jeroendesloovere:
Can come in handy for this issue https://github.com/forkcms/forkcms/issues/2092

src/Common/Doctrine/Entity/Meta.php
```diff
@@ -170,11 +170,8 @@ public function update(
 
     public function __clone()
```
carakas: can you add the :void typehint
did you also test it?

jeroendesloovere: - : void return type declaration is not allowed
- nope, didn't tested this

carakas: can you maybe test this by writing a test for it :)

carakas:
@jeroendesloovere features do not belong in bugfix releases :) http://semver.org/

jeroendesloovere:
@carakas test added. Everything is working fine.

src/Common/Doctrine/Entity/Meta.php
```diff
@@ -167,6 +168,12 @@ public function update(
         $this->seoIndex = $seoIndex;
     }
 
+    public function __clone()
+    {
+        $this->id = null;
+        $this->url = Model::addNumber($this->url);
```
carakas: hmm we can't be sure that this url will be unique
if you have my-url and my-url-2 and you clone my-url that will also turn in to my-url-2

carakas:
This is probably not something you can solve like this so I'm going to close this one

PR #2209, opened 2017-08-25
2167 fix cropper flash

### LINKED ISSUE #2167: medialibrary: cropper not enabled, but when uploading mass images, you see it
## Type

- Enhancement

## Problem description

> When uploading some images (f.e. 5) and you didn't check "show cropper", then you see a flash of the "cropper" dialog... A flash, the dialog closes itself. But it would be better if this dialog-popup-flash would be removed...

carakas:
## Type

- Non critical bugfix

fixes #2167 

## Pull request description

Fixes the flashing of the cropper when not enabled and uploading multiple images

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
+++ b/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
@@ -905,7 +905,6 @@ jsBackend.mediaLibraryHelper.cropper =
 
             jsBackend.mediaLibraryHelper.cropper.switchToCropperModal($dialog);
 
-
             return;
         }
 
--- a/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
+++ b/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
@@ -821,11 +821,7 @@ jsBackend.mediaLibraryHelper.cropper =
 
         jsBackend.mediaLibraryHelper.cropper.isCropping = true;
         var $dialog = jsBackend.mediaLibraryHelper.cropper.getDialog();
-
-        if ($('[data-role="enable-cropper-checkbox"]').is(':checked')) {
-            jsBackend.mediaLibraryHelper.cropper.switchToCropperModal($dialog);
-        }
-
+        jsBackend.mediaLibraryHelper.cropper.switchToCropperModal($dialog);
         jsBackend.mediaLibraryHelper.cropper.processNextImageInQueue($dialog);
     },
 
@@ -909,6 +905,7 @@ jsBackend.mediaLibraryHelper.cropper =
 
             jsBackend.mediaLibraryHelper.cropper.switchToCropperModal($dialog);
 
+
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/Backend/Modules/MediaLibrary/Resources/views/BackendMediaLibraryUpload.html.twig
```diff
@@ -2,9 +2,6 @@
     ====================================================================== -->
 <script type="text/template" id="qq-template-gallery">
   <div class="qq-uploader-selector qq-uploader qq-gallery" qq-drop-area-text="{{ 'lbl.DropFilesHere'|trans|ucfirst }}">
-    <div class="qq-total-progress-bar-container-selector qq-total-progress-bar-container">
```
jeroendesloovere: Why did you remove the progress bar?

carakas: It didn't show half of the time and was flashing, that made it look like something was wrong as stated in the commit message

jeroendesloovere: Hmmm it probably goes to fast if you upload smaller image.
But when uploading larger MB files that could come in handy?!

My suggestion is too not include this in your PR.
We can always remove it later.

carakas: It needs to go or needs styling, but the current state is not acceptable, it was also not clear that it was a progressbar. I thought it was some sort of panel glitch

carakas: @forkcms/frontenders @forkcms/backenders @forkcms/core-contributors  what do you think?

mathiashelin: I think we should remove the progress bar: If I upload a large image (7MB), I first see the placeholder image for some time en than the progress bar flashes really quick and disappears. So the progress bar seems kind of useless to me.

Also: selecting multiple images at once flashes the cropper popup for some reason, maybe that should be fixed too.

carakas: This pr is fixing the flashing :)

src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
```diff
@@ -909,6 +905,7 @@ jsBackend.mediaLibraryHelper.cropper =
 
             jsBackend.mediaLibraryHelper.cropper.switchToCropperModal($dialog);
 
+
```
mathiashelin: Why?

carakas: probably debug mistake

carakas: fixed in 687de40

PR #2210, opened 2017-08-28
1978 fix locale analyse

### LINKED ISSUE #1978: Locale analyse doesn't work correctly
## Type

- Non critical bug

## Problem description

/locale/analyse -> i get the message : 'No missing translations were found.' but there are missing translation , so i guess something went wrong.

## Steps to reproduce

Go to /locale/analyse

carakas:
## Type

- Non critical bugfix

fixes #1978 

## Pull request description

This will fix the long broken analysis of the translations in fork


src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
```diff
@@ -22,7 +24,47 @@ public function __construct(string $application, string $modulesPath, array $ins
...
+     */
+    private function getInBetweenStrings(string $start, string $end, string $str)
+    {
+        $matches = [];
+        preg_match_all("@$start([a-zA-Z0-9_]*)$end@", $str, $matches);
+
+        return isset($matches[1]) ? current($matches[1]) : '';
+    }
+
+    private function findModuleFiles(): array
```
StijnVrolijk: Wouldn't it be more logical to loop all modules and fetch each's files? I know that ignores modules that aren't installed but it seems to make more sense

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
```diff
@@ -47,6 +49,76 @@ private function getInBetweenStrings(string $start, string $end, string $str)
...
+            '"\.locale\.(act|err|lbl|msg)\(\'(\w+?)\'(?:, ?\'(\w+?)\')?\)"',
+            '"\.locale\.get\(\'(\w+?)\', ?\'(\w+?)\'(?:, ?\'(\w+?)\')?\)"',
+        ];
+        foreach ($patterns as $pattern) {
+            $matches = [];
+            preg_match_all($pattern, $fileContent, $matches);
+            $locale[] = $this->getLocaleArrayFromRegexMatches($this->getDefaultModule($moduleName), $matches);
+        }
+
+        if (count($locale) === 0) {
```
StijnVrolijk: I prefer using `empty` as it increases readability and is actually better performance wise 

https://stackoverflow.com/questions/2216110/checking-for-empty-arrays-count-vs-empty

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
```diff
@@ -47,6 +49,76 @@ private function getInBetweenStrings(string $start, string $end, string $str)
...
+        }
+    }
+
+    private function findLocaleInModuleFiles(string $moduleName, array $module): array
+    {
+        $locale = [];
+        foreach ($module as $filename => $file) {
+            $fileLocale = $this->findLocaleInFile($moduleName, $filename, $file);
+
+
```
StijnVrolijk: Obsolete whitespace

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
```diff
@@ -70,7 +70,6 @@ private function findLocaleInModuleFiles(string $moduleName, array $module): arr
         foreach ($module as $filename => $file) {
             $fileLocale = $this->findLocaleInFile($moduleName, $filename, $file);
 
-
```
StijnVrolijk: 👍

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
```diff
@@ -25,11 +25,47 @@ public function __construct(string $application, string $modulesPath, array $ins
...
+
+                            if (array_key_exists($nonExistingKey, $nonExisting)) {
+                                $usedInFiles = unserialize($nonExisting[$nonExistingKey]['used_in']);
+                                $usedInFiles[] = $file['file'];
+                                $nonExisting[$nonExistingKey]['used_in'] = serialize($usedInFiles);
+
+                                continue;
+                            }
+
+                            $nonExisting[$nonExistingKey] = [
```
StijnVrolijk: Would it be an idea to create a value object (of some sort) to hold this data? Seems a bit more robust than a simple array. Not saying you have to do it, I just think it may be a good idea

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
```diff
@@ -162,4 +198,51 @@ private function getDefaultModule(string $currentModuleName): string
...
+        foreach ($array1 as $key => $value) {
+            if (!array_key_exists($key, $array2) || !is_array($value)) {
+                if (!in_array($value, $array2, true)) {
+                    $outputDiff[$key] = $value;
+                }
+
+                continue;
+            }
+
+            if (is_array($value)) {
```
StijnVrolijk:
```
if (!is_array($value)) {
  continue;
}
```

for improved readability :)

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
carakas: As discussed in person: I think it is overkill and not very useful since the array is also the source of a datagrid

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
carakas: I think it would make the code more complex

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
carakas: fixed in 3bed9c3b4

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
carakas: fixed in 13b1f8df6

carakas: fixed in 13b1f8df6

src/Backend/Modules/Locale/Engine/LocaleAnalyser.php
carakas: fixed in 3bed9c3b4

PR #2218, opened 2017-08-30
update bootstrap-tagsinput to the version of mathiashelin/bootstrap-tagsinput

mathiashelin:
## Type

- Enhancement

## Pull request description

This replaces the span with the x-icon by a button with hidden text for screenreaders. This way, the button can be focused by keyboard controls and will read out "Remove tagname".

The original repo is no longer maintained, so we use a customized forked version now.


carakas:
@mathiashelin do you promise to never remove that repo?

mathiashelin:
No, because I don't like to break promises. But I can tell you I have no intention to remove this fork or my account in the foreseeable future.

carakas:
good enough :)

PR #2219, opened 2017-08-30
Integrated getsiblingsnavigation twig method

jeroendesloovere:
## Type

- Feature

You can now finally get the "siblings" without any hassle.

## Pull request description

`{{ getsiblingsnavigation(page.id) }}` will show the siblings from the same level.

If you have multiple levels, f.e.: 4 levels and you are currently on the 4th level. If you then use `{{ getsiblingsnavigation(page.id, true) }}` you will see a navigation from 2th, 3th and 4th level.


Katrienvh:
when you use `getnavigation('page', page.parent_id, null, null, <path-to-custom-navigation-template>)` and make an separate page template for the pages you want to show a subnavigation you do not need all this above. An example: https://www.muldernaturalfoods.com/nl/over-ons/over-mulder-natural-foods

jeroendesloovere:
@Katrienvh yes and no.
Yes, that is a solution.
No, it is not complete. Because you have to hardcode "should I show the navigation on this page or not". Because if you use that navigation on the "about us" page with no children ( home > about us) you will see a navigation containing 'about us', ...
=> `getsiblingsnavigation` is the way to go.

jeroendesloovere:
I'm using this feature for years now and it's time it got integrated in the core.

Katrienvh:
@jeroendesloovere if I understead you correctly than i still think my solution would work perfectly here. You do not need to check if you want or do not want to show the subnavigation on that page because you can choose to add the page template (that shows the subnavigation) on the pages you want to show the page navigation on. So for your example above you just use the default page template (that not shows subnavigation) for the about page (home > about)

jeroendesloovere:
@Katrienvh, this is what you suggest:
`make an separate page template for the pages you want to show a subnavigation`

That seems like a lot of work... when you just can use siblingsnavigation in your default theme...
I love simplicity! Saves me time.

jeroendesloovere:
Why did you close a perfectly valid solution for an existing problem?

carakas:
@Katrienvh already answered that

jeroendesloovere:
@Katrienvh here solution is only correct for a website with one level deep.
This solutions fixes the problem with highly nested levels.

jeroendesloovere:
Example:
- 1. Home
- 2. About
    - Our family
        - Me
        - My mother
        - My father
    - Our animals
        - Dogs
            - Jack
            - Russell
- 3. ...
     - ....
         - ....
         -

@Katrienvh How would you "get" the entire navigation for "2. About, ..."? Without using that "2. about" page id (because you are on level 4, you don't know the variable page.id of "2. about").

Katrienvh:
@jeroendesloovere I did something like your example here: https://www.codefever.be/nl/over-ons/wake-up-call/ict-impulsplan-van-vlaamse-overheid, nothing hardcoded. I needed some if's and else's, so it's a little more complicated but i works fine. And this/your example is a case you will not see a lot on website. So it's not impossible without your feature, but i guess your feature is less complicated. I am just wondering if we are going to use it often... Can you show an visual example? Like a screenshot?

carakas:
@jeroendesloovere can you answer the question of @Katrienvh

jeroendesloovere:
Closing this one.

PR #2221, opened 2017-08-31
Hide navigation and header in modal mode

ohvitorino:
## Type

- Feature

## Pull request description

The idea behind this pull request is to be able to open Fork backend pages from within Fork as a popup/modal, and when using this feature, the navigation and header should be hidden. This is a useful feature if we want to open a popup with the media library in it, for example.

Example:

screen shot 2017-08-31 at 15 38 56


carakas:
@ohvitorino can you  post a screenshot?

ohvitorino:
I realised I was using the wrong css classes. My initial solution will only work for the dashboard. I need to improve this.

ohvitorino:
@carakas I'll also post a screenshot.

jeroendesloovere:
I see why you want to integrate this functionality.
Could give us some nice new features.

ohvitorino:
@carakas I just notice, a problem with this solution. After the initial Fork installation, Fork gives the user the possibility to go to the backend using the same technique as the one used to open modal popups. This means the sidebar and header will also be hidden in this case, and therefore this shouldn't be merged.

ohvitorino:
@carakas Indeed. Close this one. If it's necessary to have such a functionality again, we'll have to think of something else.

PR #2224, opened 2017-08-31
Require pimple not only as a dev dependency but also production since ckfinder needs it

carakas:
## Type

- Critical bugfix

## Pull request description

Pimple was added as dev dependency while we also need it in production for ckfinder

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
         "google/recaptcha": "~1.1",
         "spoon/library": "^3.0",
         "league/flysystem-aws-s3-v3": "^1.0.13",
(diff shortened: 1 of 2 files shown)


StijnVrolijk:
@WouterSioen either one of the packages required requires pimple as a dev requirement incorrectly or @carakas made a booboo. I'm guessing the first since the pimple package was in fact in the lock file.

WouterSioen:
still a booboo :D

carakas:
@WouterSioen nobody is perfect :) it is indeed something I didn't notice, but I'm also the one fixing it :)

PR #2230, opened 2017-09-05
Pages order wcag 2.0

carakas:
## Type

- Enhancement

## Pull request description

This PR makes it possible to change the position of a page in the navigation tree using only the keyboard

screen shot 2017-09-05 at 16 52 09
screen shot 2017-09-05 at 16 52 14
screen shot 2017-09-05 at 16 52 24
screen shot 2017-09-05 at 16 52 36


src/Backend/Modules/Pages/Engine/Model.php
```diff
@@ -447,6 +445,21 @@ public static function get(int $pageId, int $revisionId = null, string $language
         return $return;
     }
 
+    public static function isForbiddenToDelete(int $pageId): bool
+    {
+        return in_array($pageId, [1, 404], true);
```
tijsverkoyen: Wouldn't it be useful to explain that 1 is the homepage and 404 is the error-page?
And maybe store the "protected" pages in a static array in the class?
Or use constants: `HOME_PAGE_ID = 1, ERROR_PAGE_ID = 404`

carakas: 2 methods use the same values but only the 404 is never allowed to have children
fixed in 12022a46d

tijsverkoyen:
I'm going to close this one, as this shouldn't be happening with Capistrano3
Even more importantly: I don't think composer should have the responsibility to run scripts on deploy.

WouterSioen:
@tijsverkoyen I guess this comment doesn't belong on this PR?

src/Backend/Modules/Pages/Engine/Model.php
```diff
@@ -1383,4 +1267,117 @@ public static function getEncodedRedirectUrl(string $redirectUrl): string
...
+            $newSequence = (int) $database->getVar(
+                'SELECT MAX(i.sequence)
+                 FROM pages AS i
+                 WHERE i.id = ? AND i.language = ? AND i.status = ?',
+                [$newParent, $language, 'active']
+            ) + 1;
+
+            $database->update(
+                'pages',
+                ['parent_id' => $newParent, 'sequence' => $newSequence, 'type' => $newType],
```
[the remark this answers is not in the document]
carakas: fixed in db88fe8ee

src/Backend/Modules/Pages/Engine/Model.php
```diff
@@ -1383,4 +1267,117 @@ public static function getEncodedRedirectUrl(string $redirectUrl): string
...
         return $redirectUrl;
     }
+
+    private static function getNewParent(int $droppedOnPageId, string $typeOfDrop, array $droppedOnPage): int
+    {
+        if ($droppedOnPageId === 0) {
+            return 0;
+        }
+
+        if ($typeOfDrop === 'inside') {
```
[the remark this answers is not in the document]
carakas: fixed in d7d6c9a9f

src/Backend/Modules/MediaLibrary/Ajax/MediaFolderMove.php
```diff
@@ -83,7 +84,7 @@ private function getTypeOfDrop(): string
             throw new AjaxExitException('no type provided');
         }
 
-        if (!in_array($typeOfDrop, ['before', 'after', 'inside'], true)) {
+        if (!in_array($typeOfDrop, Model::POSSIBLE_TYPES_OF_DROP, true)) {
```
tijsverkoyen: I don't find this readable, in my opinion, it would make more sense to list the possible drop-types here in the method.

WouterSioen: It might be less readable, but the good thing about this is that if you change it in your model, it's automatically changed everywhere. Imo, this is better (less duplication, less possible future bugs) than listing them up here.

tijsverkoyen: I do not agree, I prefer readability over developer-laziness.
If you don't want to have code-duplication, the proper solution would be to extract the check into a separate method which checks if the type of drop is allowed.

carakas:
check to update the 404 id after #2278  has been merged

carakas:
still need to some js code styles

PR #2231, opened 2017-09-06
Editable hreflang tags for pages

wouterverstuyf:
## Type

- Enhancement

The hreflang tags are displayed based on the page ID, so if these are not equal for different languages the tags are not correct.

## Pull request description

When editing or creating a new page, you can select the matching page for each (other) language from a dropdown.

## Improvements

1. Updating the installer so the hreflang tags after install are correct
2. Maybe move this functionality to the SEO workflow, instead of printing them
3. Find a way to use this also for specific modules like Blog


src/Backend/Modules/Pages/Actions/Add.php
```diff
@@ -420,6 +441,16 @@ private function validateForm(): void
                     $data['image'] = $this->getImage($this->templates[$templateId]['data']['image']);
                 }
 
+                // just execute if the site is multi-language
+                if ($this->getContainer()->getParameter('site.multilanguage')) {
+                    // loop active languages
+                    foreach (BL::getActiveLanguages() as $language) {
+                        if($language != BL::getWorkingLanguage()) {
+                            $data['hreflang_' . $language] = $this->form->getfield('hreflang_' . $language)->getValue();
```
carakas: maybe only add it when the field is actually filled, and the same in the edit action

src/Frontend/Core/Engine/Page.php
```diff
@@ -382,12 +382,15 @@ private function addAlternateLinkForLanguage(string $language): void
...
-        $url = '';
-        if (!empty($pageInfo['data']['hreflang_' . $language])) {
-            $url = Navigation::getUrl($pageInfo['data']['hreflang_' . $language], $language);
+
+        // Check if hreflang is set for language
+        if (!isset($pageInfo['data']['hreflang_' . $language])) {
+            return;
         }
 
+        // Set url
```
carakas: this is actually a bit of a useless comment, you can remove this one

src/Frontend/Core/Engine/Page.php
```diff
@@ -379,7 +379,17 @@ private function addAlternateLinkForLanguage(string $language): void
...
+        // Get page data
+        $pageInfo = Model::getPage($this->pageId);
+
+
+        // Check if hreflang is set for language
+        if (!isset($pageInfo['data']['hreflang_' . $language])) {
+            return;
+        }
+
+        // Set url
```
carakas: This is a bit of a useless comment

src/Backend/Modules/Pages/Actions/Add.php
```diff
@@ -420,6 +441,16 @@ private function validateForm(): void
                     $data['image'] = $this->getImage($this->templates[$templateId]['data']['image']);
                 }
 
+                // just execute if the site is multi-language
+                if ($this->getContainer()->getParameter('site.multilanguage')) {
+                    // loop active languages
+                    foreach (BL::getActiveLanguages() as $language) {
+                        if ($language != BL::getWorkingLanguage()) {
+                            $data['hreflang_' . $language] = $this->form->getfield('hreflang_' . $language)->getValue();
```
carakas: maybe only add it when the field is actually filled, and the same in the edit action

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -690,6 +711,16 @@ private function validateForm(): void
                     $data['remove_from_search_index'] = true;
                 }
 
+                // just execute if the site is multi-language
+                if ($this->getContainer()->getParameter('site.multilanguage')) {
+                    // loop active languages
+                    foreach (BL::getActiveLanguages() as $language) {
+                        if ($language != BL::getWorkingLanguage()) {
+                            $data['hreflang_' . $language] = $this->form->getfield('hreflang_' . $language)->getValue();
```
carakas: maybe only add it when the field is actually filled, and the same in the edit action

src/Backend/Modules/Pages/Actions/Add.php
```diff
@@ -446,7 +446,9 @@ private function validateForm(): void
                     // loop active languages
                     foreach (BL::getActiveLanguages() as $language) {
                         if ($language != BL::getWorkingLanguage()) {
-                            $data['hreflang_' . $language] = $this->form->getfield('hreflang_' . $language)->getValue();
+                            if ($this->form->getfield('hreflang_' . $language)->isFilled()) {
```
carakas: you can combine these 2 ifs

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -716,7 +716,9 @@ private function validateForm(): void
                     // loop active languages
                     foreach (BL::getActiveLanguages() as $language) {
                         if ($language != BL::getWorkingLanguage()) {
-                            $data['hreflang_' . $language] = $this->form->getfield('hreflang_' . $language)->getValue();
+                            if ($this->form->getfield('hreflang_' . $language)->isFilled()) {
```
carakas: idem

wouterverstuyf:
The installeer needs an update, but not 100% sure how to do this the best way...
I think the best location is in:
`\src\Backend\Core\Installer\ModuleInstaller.php - function completePageRevisionRecord`
So the changes should be something like:
- loop through all the languages (except the language of the revision entry)
- add data hreflang_<language> tags with the same value as the $revision['id']
Am I correct? If so, I will try this

carakas:
@wouterverstuyf not sure if that would work properly in all cases because if you only add a page to 1 language in the installer and in another installer to another language you have an id mismatch again

wouterverstuyf:
@carakas Correct, and if you add the first page there are no other records in the other languages, so you can't add the hreflang tags... gonna think about it.
If someone has an idea, help is appreciated, would be great to have this PR included :)

wouterverstuyf:
@carakas 
I've tried with a simple change to the function I mentioned and it seems to work. I have reinstalled the core, added some modules afterwards and all of the created pages get linked well. Hmz, seems to easy... what do you think?

carakas:
If it works it works :)

src/Backend/Core/Installer/ModuleInstaller.php
```diff
@@ -710,7 +710,11 @@ private function completePageRevisionRecord(array $revision, array $meta = []):
             $revision['type']
         );
         $revision['meta_id'] = $revision['meta_id'] ?? $this->getNewMetaId($meta, $revision['title']);
-
+        foreach ($this->getLanguages() as $language) {
+            if ($language != $revision['language']) {
```
carakas: should be` !==`

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
```diff
@@ -196,6 +196,23 @@
...
+            <div class="row">
+              <div class="col-md-12">
+                <div class="panel panel-default">
+                  <div class="panel-heading">
+                    <p class="tab-pane-title">{{ 'lbl.Languages'|trans|ucfirst }}</p>
+                  </div>
+                  <div class="panel-body">
+                    {% for language, field in hreflangFields %}
+                      <div class="form-group{% if loop.last %} last{% endif %}">
+                        <label for="">{{ language|upper|raw }}</label>
```
carakas: you need to add the id to the for attribute

src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
```diff
@@ -310,6 +310,23 @@
...
+            <div class="row">
+              <div class="col-md-12">
+                <div class="panel panel-default">
+                  <div class="panel-heading">
+                    <p class="tab-pane-title">{{ 'lbl.Languages'|trans|ucfirst }}</p>
+                  </div>
+                  <div class="panel-body">
+                    {% for language, field in hreflangFields %}
+                      <div class="form-group{% if loop.last %} last{% endif %}">
+                        <label for="">{{ language|upper|raw }}</label>
```
carakas: empty for attribute

src/Frontend/Core/Engine/Page.php
```diff
@@ -379,7 +379,15 @@ private function addAlternateLinkForLanguage(string $language): void
             return;
         }
 
-        $url = Navigation::getUrl($this->pageId, $language);
+        // Get page data
+        $pageInfo = Model::getPage($this->pageId);
+
```
carakas: enter too much

index.php
```diff
@@ -31,6 +31,8 @@
 $env = getenv('FORK_ENV') ?: 'prod';
 $debug = getenv('FORK_DEBUG') === '1';
 
+$debug = 1;
```
carakas: this should be here

carakas:
@wouterverstuyf  I think you accidentally commit to much here, there is an entire custom module icluded in this pr

carakas:
the problem is in this pr https://github.com/forkcms/forkcms/pull/2231/commits/1a4eed5e74f5148f67ca40d4f705c912f91f8905

wouterverstuyf:
Messy commits, don't know how I dit it... :) 
Delete this PR please, I will create a new PR with the implementation of the Hreflang tags.

PR #2232, opened 2017-09-06
Fix tagsinput

mathiashelin:
## Type

- Critical bugfix

## Pull request description

Tagsinput 0.8.0 no longer supported items as objects and the possibility to add new tags, so now the tagsinput uses an array of strings, which does allow adding new tags as well as using pre-existings tags.


src/Backend/Modules/Tags/Engine/Model.php
```diff
@@ -81,6 +81,16 @@ public static function getAll(string $language = null): array
         );
     }
 
+    public static function getTagNames(string $language = null): array
+    {
+        return (array) BackendModel::getContainer()->get('database')->getColumn(
+            'SELECT i.tag AS name
```
jonasdekeukelaere: The "AS name" can be removed. Also no need to use "i" as alias.

src/Backend/Modules/Tags/Engine/Model.php
```diff
@@ -81,6 +81,16 @@ public static function getAll(string $language = null): array
         );
     }
 
+    public static function getTagNames(string $language = null): array
+    {
+        return (array) BackendModel::getContainer()->get('database')->getColumn(
+            'SELECT tag
+             FROM tags
+             WHERE i.language = ?',
```
carakas: you forgot to remove the `i.` here

PR #2235, opened 2017-09-07
Https error on header

carakas:
## Type

- Non critical bugfix

## Pull request description
When enabling https for production you sometimes can get an error locally if you locally don't have the header module enabled

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -6,8 +6,8 @@ DirectorySlash Off
 
 # Make sure the htaccessfile is protected
 <Files .htaccess>
-	order allow,deny
-	deny from all
+    order allow,deny
+    deny from all
 </Files>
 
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -75,6 +75,9 @@ deny from env=stayout
 	RewriteRule docker-compose\.override\.yml - [F]
 
 	# redirect all trafic to https
+    # <IfModule headers_module>
+    # 	Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
+    # </IfModule>
 	# Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
```
jonasdekeukelaere: This line can go then?

carakas: oops, thanks 98279ae74

PR #2237, opened 2017-09-08
2120 mailmotor different list for all languages

### LINKED ISSUE #2120: Mailmotor different list for all languages
## Type

- Enhancement

## Problem description

There is no possibility to use other lists for the different languages.

carakas:
## Type

- Feature

fixes #2120 

## Pull request description

In the old mailmotor module it was possible to connect a specific list to a specific language, this PR brings back the good old days.

The current list id will be the default but it can be overruled when a language specific list id has been entered
screen shot 2017-09-08 at 16 29 53


src/Backend/Modules/Mailmotor/Actions/Ping.php
```diff
@@ -24,14 +25,38 @@ public function execute(): void
...
 
         $this->resetMailEngine();
         $this->redirect($this->getBackLink(['error' => 'wrong-mail-engine-credentials']));
     }
 
+    private function ping(): bool
+    {
+        $gateway = $this->getContainer()->get('mailmotor.factory')->getSubscriberGateway();
+
+        if (!$gateway->ping($this->getContainer()->getParameter('mailmotor.list_id'))) {
```
tijsverkoyen: Could you add some comments, because I have no clue what this method does. I find the name `ping` not really descriptive. I guess it checks if some list exists?

src/Backend/Modules/Mailmotor/Actions/Ping.php
```diff
@@ -24,14 +25,38 @@ public function execute(): void
...
+            return false;
+        }
+
+        $settings = $this->getContainer()->get('fork.settings');
+        foreach (Language::getActiveLanguages() as $language) {
+            $languageListId = $settings->get('Mailmotor', 'list_id_' . $language);
+
+            if ($languageListId === null) {
```
tijsverkoyen: Same here, why continue if the id is null? Shouldn't the list be created?

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
```diff
@@ -26,24 +27,56 @@ public function __construct(
...
         }
 
-        $this->modulesSettings->set($module, 'api_key', $settings->apiKey);
-        $this->modulesSettings->set($module, 'list_id', $settings->listId);
+        $this->saveSetting('api_key', $settings->apiKey);
+        $this->saveSetting('list_id', $settings->listId);
+        $this->saveLanguageListIds($settings->languageListIds);
+    }
+
+    private function saveSetting(string $key, $value): void
```
tijsverkoyen: Isn't this bogus code? In my opinion this is not improving/simplifying the code, it is just "aliasing" a method.
I think it is wrong to alter the terminology we use. We `set` a setting, not `save`.

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
```diff
@@ -26,24 +27,56 @@ public function __construct(
...
+    {
+        $this->modulesSettings->set('Mailmotor', $key, $value);
```
tijsverkoyen: If you don't agree, with the comment above, at least use a constant for the Module-name?
Related: wouldn't it be a good idea to make that a requirement in upcoming release? So we can use the module-name in a well-know constant?

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
```diff
@@ -26,24 +27,56 @@ public function __construct(
...
+    }
+
+    private function deleteSetting(string $key): void
```
tijsverkoyen: See above

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
```diff
@@ -26,24 +27,56 @@ public function __construct(
...
+    {
+        $this->modulesSettings->delete('Mailmotor', $key);
+    }
+
+    public function isActiveLanguage(string $language): bool
```
tijsverkoyen: Wouldn't it be better to make this method part of the language-class?

src/Frontend/Modules/Mailmotor/EventListener/SubscribeFromFormBuilderSubmittedForm.php
```diff
@@ -22,60 +23,58 @@
...
-                        [],
-                        false // will ignore double-optin and so subscribes the user immediately
-                    );
-                }
-            } catch (NotImplementedException $e) {
-                // We do nothing as fallback when no mail-engine is chosen in the Backend
+        // We subscribe the replyTo email
+        try {
+            if (!$this->subscriber->exists($email)) {
+                // @TODO check if we are allowed to do this like this according to GDPR
```
tijsverkoyen: You can't do that. The only way you could implement this, if you would add a checkbox in the form that explains to the user on which list he/she will be subscribed, the interval, ... 
So doing this automatically should not be in the default implementation of Fork, so I advice to remove all related code.

src/Backend/Modules/Mailmotor/Actions/Ping.php
carakas: fixed in a7352acf1

src/Frontend/Modules/Mailmotor/EventListener/SubscribeFromFormBuilderSubmittedForm.php
carakas: solved in a7352acf1

src/Backend/Modules/Mailmotor/Actions/Ping.php
carakas: clarified in 4cf05dc

src/Backend/Modules/Mailmotor/Actions/Ping.php
carakas: fixed in 35de0b801

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
carakas: changed in 81acd5752

How do you see that working? The constants?

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
carakas: changed in 81acd5752

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
carakas: changed in 81acd5752

src/Backend/Modules/Mailmotor/Domain/Settings/Command/SaveSettingsHandler.php
carakas: true, done in 214416344

carakas:
@tijsverkoyen can you check this pr again

PR #2238, opened 2017-09-11
Fix formbuilder data action

carakas:
## Type

- Non critical bugfix

## Pull request description

The default value for a date could be an empty string, this resulted in errors

Apart from that updating the filter resulted in a redirect to the index page because the id was lost


ohvitorino:
This looks like a good fix for the problem.

PR #2245, opened 2017-09-14
Capistrano htaccess

jeroendesloovere:
## Type

- Enhancement

## Pull request description

When deploying your Fork CMS website using **Capistrano** and **"file"** to clear the opcache. It is utmost necessary to add the following line to your htaccess, otherwise the entire deployment will always FAIL.


tijsverkoyen:
Not relevant, as not everybody uses Capistrano. I updated the docs for [capistrano-forkcms](https://github.com/tijsverkoyen/capistrano-forkcms#opcache_reset_strategy)

carakas:
@tijsverkoyen @jeroendesloovere the htaccess is also wrong here https://github.com/tijsverkoyen/capistrano-forkcms#opcache_reset_strategy

It should be `^RewriteRule php-opcache-reset\.php`$ - [L]`

tijsverkoyen:
Fixed in the docs

PR #2246, opened 2017-09-14
Throw error when template syntax is faulty

carakas:
## Type

- Critical bugfix

## Pull request description

When editing the template structure for the pages module you could enter faulty data without getting an error.
This was the result of fixing things for fork 5 and was updated in the add but not in the edit action


iamannewatson:
Interesting article! Thank you for sharing them! I hope you will continue to have similar posts to share with everyone! I believe a lot of people will be surprised to read this article! [cheap web hosting and domain](http://www.gistnetwork.org/node/47473)

PR #2257, opened 2017-09-18
Location widget marker was missing the address

jeroendesloovere:
## Type

- Non critical bugfix

## Pull request description

In the location widgets, in the marker, the address was missing.
Because of a renamed data-attribute.

Before the fix
schermafbeelding 2017-09-18 om 13 38 53
After the fix
schermafbeelding 2017-09-18 om 13 39 21


carakas:
1 question, the big green text, is that your theme or fork? :O

PR #2259, opened 2017-09-20
Add default value as backup

freshface:
When the user has not saved his profile before the $lineEnding value is null

## Type

- Non critical bugfix

### DIFF
--- a/src/Backend/Core/Engine/Csv.php
+++ b/src/Backend/Core/Engine/Csv.php
@@ -63,7 +63,7 @@ class Csv extends \SpoonFileCSV
 
     private static function getLineEnding(): string
     {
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Engine/Csv.php
```diff
@@ -63,7 +63,7 @@ public static function outputCSV(
 
     private static function getLineEnding(): string
     {
-        $lineEnding = Authentication::getUser()->getSetting('csv_line_ending');
```
carakas: Can you fix this in the installer instead? That would be the correct place for this.

carakas:
duplicate of #2262 but that pr's fix is better

PR #2262, opened 2017-09-21
Set the default CSV settings for our god user

StijnVrolijk:
## Type

- Bugfix

## Pull request description

When doing a fresh install of Fork, the god user's default CSV settings would be `NULL`. As there is no fallback when fetching these settings in the `Csv` class, an error would be thrown. Instead of setting a fallback, the correct default settings should've been set. I did this by fetching the possible settings to choose from in the Add/Edit forms and simply selecting the first options.

### DIFF
--- a/src/Backend/Modules/Users/Installer/Installer.php
+++ b/src/Backend/Modules/Users/Installer/Installer.php
@@ -161,6 +161,10 @@ class Installer extends ModuleInstaller
             $settings['password_strength'] = $this->getPasswordStrength();
             $settings['current_password_change'] = time();
             $settings['avatar'] = 'god.jpg';
+            $possibleCSVSplitCharacters = BackendUsersModel::getCSVSplitCharacters();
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Users/Installer/Installer.php
```diff
@@ -161,6 +162,10 @@ private function loadGodUser(): void
             $settings['password_strength'] = $this->getPasswordStrength();
             $settings['current_password_change'] = time();
             $settings['avatar'] = 'god.jpg';
+            $possibleCSVSplitCharacters = Model::getCSVSplitCharacters();
```
carakas: There is already a use statement for it: `BackendUsersModel`

carakas:
duplicate of #2259  but this fix is better

StijnVrolijk:
@carakas I've used the pre-existing model in my amend. I hate aliases :p

PR #2263, opened 2017-09-21
Send a proper response when the ajax call is unauthenticated

carakas:
## Type

- Non critical bugfix

## Pull request description

Ajax calls should give a proper response mentioning that the call was unauthenticated instead of throwing exceptions

### DIFF
--- a/src/Backend/Core/Engine/Ajax.php
+++ b/src/Backend/Core/Engine/Ajax.php
@@ -14,6 +14,7 @@ use ForkCMS\App\KernelLoader;
 use Symfony\Component\HttpFoundation\Response;
 use Backend\Core\Engine\Base\AjaxAction as BackendBaseAJAXAction;
 use Backend\Core\Language\Language as BackendLanguage;
+use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Backend/Core/Engine/Ajax.php
```diff
@@ -36,8 +37,15 @@ public function display(): Response
...
     public function initialize(): void
     {
-        // check if the user is logged in
-        $this->validateLogin();
+        try {
+            // check if the user is logged in
+            $this->validateLogin();
+        } catch (UnauthorizedHttpException $e) {
+            $this->ajaxAction = new BackendBaseAJAXAction($this->getKernel());
+            $this->ajaxAction->output(Response::HTTP_UNAUTHORIZED, null, $e->getMessage());
```
tijsverkoyen: Als ik https://github.com/symfony/http-foundation/blob/master/Response.php#L44 dan is dat 401. Maar https://github.com/justcarakas/forkcms/blob/185544c21fde5e26615dd9515de2214e88f5ed86/src/Backend/Core/Js/backend.js#L91 verwacht een 403.

carakas: fixed in 7e8cf405c

PR #2264, opened 2017-09-21
Dont try to minify external urls

carakas:
## Type

- Non critical bugfix

## Pull request description

Developers needed to manually disable `minify` and enable `overwritePath` even when it was clear it was an external asset. This resulted in unexpected behaviour when the developer didn't pay attention to it since the assets don't minify in debug mode

After this fix you will be able to do the following without extra effort or parameters and it will work as expected
```
        $this->header->addJS('//text.com/test.js');
        $this->header->addJS('http://text.com/test.js');
        $this->header->addJS('https://text.com/test.js');
```

### DIFF
--- a/src/Backend/Core/Engine/Header.php
+++ b/src/Backend/Core/Engine/Header.php
@@ -162,7 +162,7 @@ final class Header extends KernelLoader
                 $addTimestamp,
                 $priority ?? ($overwritePath ? Priority::standard() : Priority::forModule($module))
             ),
-            $minify && !$this->getContainer()->getParameter('kernel.debug')
+            $minify
         );
     }
 
--- a/src/Backend/Core/Engine/Header.php
+++ b/src/Backend/Core/Engine/Header.php
@@ -139,7 +139,7 @@ final class Header extends KernelLoader
      * @param string $module The module wherein the file is located.
      * @param bool $overwritePath Should we overwrite the full path?
      *                            An external url will always be handled with $overwritePath as true.
-     * @param bool $minify Should the CSS be minified? An external url will never be minified.
+     * @param bool $minify Should the CSS be minified?
      * @param bool $addTimestamp May we add a timestamp for caching purposes?
      * @param Priority $priority the files are added based on the priority
      *                           defaults to standard for full links or core or module for core or module css
@@ -174,7 +174,7 @@ final class Header extends KernelLoader
      *
      * @param string $file The file to load.
      * @param string $module The module wherein the file is located.
-     * @param bool $minify Should the module be minified? An external url will never be minified.
+     * @param bool $minify Should the javascript be minified?
      * @param bool $overwritePath Should we overwrite the full path?
      *                            An external url will always be handled with $overwritePath as true.
      * @param bool $addTimestamp May we add a timestamp for caching purposes?
--- a/src/Backend/Core/Engine/Header.php
+++ b/src/Backend/Core/Engine/Header.php
@@ -184,7 +184,7 @@ final class Header extends KernelLoader
         Priority $priority = null
     ): void {
         $module = $module ?? $this->url->getModule();
-
+        $overwritePath = $overwritePath || strpos($file, 'http') === 0 || strpos($file, '//') === 0;
         $this->jsFiles->add(
             new Asset(
                 $overwritePath ? $file : $this->buildPathForModule($file, $module ?? $this->url->getModule(), 'Js'),
--- a/src/Backend/Core/Engine/Header.php
+++ b/src/Backend/Core/Engine/Header.php
@@ -189,6 +189,7 @@ final class Header extends KernelLoader
         bool $addTimestamp = false,
         Priority $priority = null
     ): void {
+        $module = $module ?? $this->url->getModule();
         $isExternalUrl = $this->get('fork.validator.url')->isExternalUrl($file);
         $overwritePath = $overwritePath || $isExternalUrl;
         $minify = $minify && !$isExternalUrl;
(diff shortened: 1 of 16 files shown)


src/Backend/Core/Engine/Header.php
```diff
@@ -184,7 +184,7 @@ public function addJS(
         Priority $priority = null
     ): void {
         $module = $module ?? $this->url->getModule();
-
+        $overwritePath = $overwritePath || strpos($file, 'http') === 0 || strpos($file, '//') === 0;
```
tijsverkoyen: Wouldn't it be more failsafe to check if the given path is an url? Because now if you have a valid but non-external path (e.g.: /foo//bar-http.js) it won't be minified.

src/Common/Core/Header/AssetCollection.php
```diff
@@ -19,7 +19,7 @@ public function __construct(Minifier $minifier)
 
     public function add(Asset $asset, bool $minify = true): void
     {
-        if ($minify) {
+        if ($minify && strpos($asset->getFile(), 'http') !== 0 && strpos($asset->getFile(), '//') !== 0) {
```
tijsverkoyen: Wouldn't it be more logical/readable to reset `$minify` to `false` if the url is external.
Also same remark as above.

src/Backend/Core/Engine/Header.php
carakas: this has been fixed

src/Common/Core/Header/AssetCollection.php
carakas: this has been fixed

src/ForkCMS/Bundle/CoreBundle/Tests/Validator/UrlValidatorTest.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace ForkCMS\Bundle\CoreBundle\Tests\Validator;
+
+use ForkCMS\Bundle\CoreBundle\Validator\UrlValidator;
+use PHPUnit\Framework\TestCase;
+
+class UrlValidatorTest extends TestCase
+{
+    public function testExternalUrlValidation()
```
tijsverkoyen: Isn't this a bit generic. I would separate this into two test methods:
* testExternalUrlValidationWithValidUrls
* testExternalUrlValidationWithInvalidUrls

src/Frontend/Core/Engine/Base/Block.php
```diff
@@ -137,6 +137,8 @@ public function addCSS(
         bool $minify = true,
         bool $addTimestamp = false
     ): void {
+        $overwritePath = $overwritePath || $this->get('fork.validator.url')->isExternalUrl($file);
```
tijsverkoyen: I don't find this very readable, I would at least add a comment which explains that external urls always override the path.

src/Backend/Core/Engine/Header.php
```diff
@@ -183,7 +189,9 @@ public function addJS(
         bool $addTimestamp = false,
         Priority $priority = null
     ): void {
-        $module = $module ?? $this->url->getModule();
+        $isExternalUrl = $this->get('fork.validator.url')->isExternalUrl($file);
+        $overwritePath = $overwritePath || $isExternalUrl;
```
tijsverkoyen: I would add a comment that explains that if the given path is an external url it always overwrites the path.

src/Frontend/Core/Header/Header.php
```diff
@@ -144,7 +144,7 @@ public function __construct(KernelInterface $kernel)
      * Add a CSS file into the array
      *
      * @param string $file The path for the CSS-file that should be loaded.
-     * @param bool $minify Should the CSS be minified?
+     * @param bool $minify Should the CSS be minified? An external url will never be minified.
```
tijsverkoyen: see above

src/Backend/Core/Engine/Header.php
```diff
@@ -169,8 +174,9 @@ public function addCSS(
      *
      * @param string $file The file to load.
      * @param string $module The module wherein the file is located.
-     * @param bool $minify Should the module be minified?
+     * @param bool $minify Should the module be minified? An external url will never be minified.
```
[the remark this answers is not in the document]
carakas: fixed in 91883c2c3

src/Frontend/Core/Header/Header.php
carakas: fixed in 91883c2c3

src/Frontend/Core/Engine/Base/Block.php
carakas: fixed in 0ca9bdeb7

src/ForkCMS/Bundle/CoreBundle/Tests/Validator/UrlValidatorTest.php
carakas: fixed in a706fd97c

src/Backend/Core/Engine/Header.php
```diff
@@ -152,13 +153,17 @@ public function addCSS(
         Priority $priority = null
     ): void {
         $module = $module ?? $this->url->getModule();
+        $isExternalUrl = $this->get('fork.validator.url')->isExternalUrl($file);
+        $overwritePath = $overwritePath || $isExternalUrl; // external urls always overwrite the path
+        $minify = $minify && !$isExternalUrl;
```
tijsverkoyen: Am I correct if I say the files are always minified, even in debug-mode?
In my personal opinion this is a good thing, but I can think of some people (who aren't using gulp an write plain css instead of using sass) that can have some issues.

src/Backend/Core/Engine/Header.php
```diff
@@ -184,6 +190,9 @@ public function addJS(
...
```
tijsverkoyen: Same here

src/Backend/Core/Engine/Header.php
carakas: No, the debug check is inside the minifier.

tijsverkoyen: 👍

PR #2269, opened 2017-09-25
OpenGraph error when you add a OneImage MediaLibrary widget without images

jeroendesloovere:
## Type

- Non critical bugfix

When you add a "OneImage" MediaLibrary widget to a page. But you haven't uploaded images to it yet, an error is thrown because there is no image to addToOpenGraph.

## Solution

Check added to verify that it is empty.


carakas:
@jeroendesloovere can you create a proper commit message?

carakas:
nm I'll fix it with a squash

PR #2278, opened 2017-09-28
Improve 404 and forbidden pages

carakas:
## Type

- Enhancement

## Pull request description

In the frontend we redirect to the 404 page. That implementation has been there for a while but it actually isn't the correct implementation. We should show the 404 on the url where it is actually happening

This includes also a clean up of some of the actions and widgets


src/Frontend/Core/Engine/Page.php
```diff
@@ -117,8 +118,17 @@ public function load(): void
         // create header instance
         $this->header = new Header($this->getKernel());
 
+        try {
+            $this->handlePage(Navigation::getPageId(implode('/', $this->url->getPages())));
+        } catch (NotFoundHttpException $notFoundHttpException) {
+            $this->handlePage(Response::HTTP_NOT_FOUND);
```
tijsverkoyen: I assume the constant `Response::HTTP_NOT_FOUND` is 404? What if they change the value of the constant? I think the purpose of using class-constants is to be able to change the internal value without breaking the public API

src/Frontend/Core/Engine/Page.php
```diff
@@ -181,34 +191,43 @@ private function checkAuthentication(): void
...
-        $this->header->parse();
-        $this->breadcrumb->parse();
-        $this->parseLanguages();
-        $this->footer->parse();
+            return new Response(
+                $this->template->getContent($this->templatePath),
+                $this->statusCode
+            );
+        } catch (NotFoundHttpException $notFoundHttpException) {
+            $this->handlePage(Response::HTTP_NOT_FOUND);
```
tijsverkoyen: This could potentially cause a loop, if the 404-page is deleted. But #yolo

src/Frontend/Modules/Blog/Actions/Category.php
```diff
@@ -92,7 +93,7 @@ private function getData(): void
 
         // redirect if the request page doesn't exists
```
tijsverkoyen: `request` => `requested`

src/Frontend/Modules/Profiles/Actions/Activate.php
```diff
@@ -47,10 +48,10 @@ public function execute(): void
                 $this->template->assign('activationSuccess', true);
             } else {
                 // failure
-                $this->redirect(FrontendNavigation::getUrl(404));
+                throw new NotFoundHttpException();
```
tijsverkoyen: Seems like this code could be simplified with some early returns/throws

src/Frontend/Modules/Profiles/Actions/ResendActivation.php
```diff
@@ -38,7 +39,7 @@ public function execute(): void
             $this->parse();
         } else {
             // profile logged in
```
tijsverkoyen: Faulty comment?

src/Frontend/Modules/Profiles/Actions/ResetPassword.php
```diff
@@ -51,14 +52,14 @@ public function execute(): void
 
                 // validate
                 $this->validateForm();
-            } elseif ($this->url->getParameter('sent') != 'true') {
-                $this->redirect(FrontendNavigation::getUrl(404));
+            } elseif ($this->url->getParameter('sent') !== 'true') {
+                throw new NotFoundHttpException();
             }
 
             // parse
```
tijsverkoyen: Useless comment

src/Frontend/Core/Engine/Page.php
carakas: the 404 cannot be deleted, it is protected

src/Frontend/Modules/Profiles/Actions/Activate.php
carakas: I had been thinking about cleaning it up but in the end I didn't do it since a lot of the actions need cleaning up, but I'll do the those that have changes in this pr

src/Frontend/Core/Engine/Page.php
carakas: fixed in cf7eaae29

src/Frontend/Core/Engine/Page.php
tijsverkoyen: I know, hence the #yolo

carakas:
note to self => we will need a different cache pool for the search as clearing the cache bool completely results in errors

src/Frontend/Modules/Profiles/Actions/ChangeEmail.php
```diff
@@ -38,94 +37,79 @@ class ChangeEmail extends FrontendBaseBlock
...
-            // profile not logged in
-            $this->redirect(
-                FrontendNavigation::getUrlForBlock(
-                    'Profiles',
-                    'Login'
-                ) . '?queryString=' . FrontendNavigation::getUrlForBlock('Profiles', 'ChangeEmail'),
-                307
-            );
+        if (!FrontendProfilesAuthentication::isLoggedIn()) {
+            throw new InsufficientAuthenticationException('You need to log in to change your email');
```
tijsverkoyen: If you are not logged in, you can't know if the user has insufficient rights, so a Forbidden would be more logical

src/Frontend/Modules/Search/Actions/Index.php
```diff
@@ -147,16 +113,14 @@ private function getRealData(): void
             throw new NotFoundHttpException();
         }
 
-        if ($this->getContainer()->getParameter('kernel.debug')) {
+        if (!$this->getContainer()->getParameter('kernel.debug')) {
```
tijsverkoyen: I don't get what this line is doing here. Could you add a comment?

.htaccess
```diff
@@ -66,7 +66,6 @@ deny from env=stayout
     RewriteRule var/logs - [F]
     RewriteRule app/config - [F]
     RewriteRule src/Frontend/Cache/CompiledTemplates - [F]
-    RewriteRule src/Frontend/Cache/Search - [F]
```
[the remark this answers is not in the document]
carakas: The only thing that needs blocking here at the moment are the compiled templates, all the other files need to be accessable. PHP files are already blocked globally so I think we can leave it like this at the moment

src/Frontend/Core/Engine/Page.php
```diff
@@ -150,9 +163,7 @@ private function checkAuthentication(): void
         }
 
         if (!FrontendAuthenticationModel::isLoggedIn()) {
-            $this->redirect(
-                Navigation::getUrlForBlock('Profiles', 'Login') . '?queryString=' . $this->url->getQueryString()
-            );
+            throw new InsufficientAuthenticationException('You must log in to see this page');
```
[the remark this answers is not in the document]
carakas: This is the way symfony works
> InsufficientAuthenticationException is thrown if the user credentials are not sufficiently trusted.
> This is the case when a user is anonymous and the resource to be displayed has an access role.

http://api.symfony.com/2.8/Symfony/Component/Security/Core/Exception/InsufficientAuthenticationException.html

src/Frontend/Modules/Blog/Actions/Archive.php
```diff
@@ -78,128 +44,143 @@ public function execute(): void
...
-        $this->pagination['num_items'] = FrontendBlogModel::getAllForDateRangeCount($this->startDate, $this->endDate);
-        $this->pagination['num_pages'] = (int) ceil($this->pagination['num_items'] / $this->pagination['limit']);
+    private function setDateRange(): void
+    {
+        $baseUrl = Navigation::getUrlForBlock($this->getModule(), $this->getAction());
+        $slug = trim(str_replace($baseUrl, '', $this->getRequest()->getPathInfo()), '/');
 
-        // redirect if the request page doesn't exists
-        if ($requestedPage > $this->pagination['num_pages'] || $requestedPage < 1) {
+        $this->hasMonth = (bool) mb_strpos($slug, '/');
```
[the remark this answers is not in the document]
carakas: that won't work because you can't create a valid date (line 99) from that so you'll get a 404

src/Frontend/Core/Engine/Block/ExtraInterface.php
```diff
@@ -150,9 +150,9 @@ public function execute(): void
      * REMARK: You should not use this method from your code, but it has to be
      * public so we can access it later on in the core-code
      *
-     * @return string
+     * @return string|null
      */
-    public function getAction(): string
+    public function getAction(): ?string
```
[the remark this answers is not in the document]
carakas: I'll add a comment to explain it. This is used to trigger the default action

src/Frontend/Modules/Blog/Actions/Archive.php
```diff
@@ -78,128 +44,143 @@ public function execute(): void
...
+    /**
+     * Complete the slug to prevent wrong months when the current day is higher than possible in the requested month
+     *
+     * @param string $slug
+     *
+     * @return string
+     */
+    private function getStartDateSlug(string $slug):string
+    {
+        if (!$this->hasMonth) {
```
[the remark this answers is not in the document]
carakas: this method is used to make sure we have the correct format to be able to create a date time.
The slug can be only a year or year/month and what you need to pass the correct date string

src/Frontend/Modules/Blog/Actions/Archive.php
```diff
@@ -78,128 +44,143 @@ public function execute(): void
...
+        return $slug . '/01';
+    }
 
-        // populate count fields in pagination
-        $this->pagination['num_items'] = FrontendBlogModel::getAllForDateRangeCount($this->startDate, $this->endDate);
-        $this->pagination['num_pages'] = (int) ceil($this->pagination['num_items'] / $this->pagination['limit']);
+    private function setDateRange(): void
+    {
+        $baseUrl = Navigation::getUrlForBlock($this->getModule(), $this->getAction());
+        $slug = trim(str_replace($baseUrl, '', $this->getRequest()->getPathInfo()), '/');
```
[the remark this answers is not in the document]
carakas: fixed in ad1cf7286

src/Frontend/Modules/Blog/Actions/Category.php
```diff
@@ -56,94 +31,87 @@ public function execute(): void
...
+    private function buildUrl(): string
+    {
+        return FrontendNavigation::getUrlForBlock($this->getModule(), $this->getAction())
+               . '/' . $this->category['url'];
+    }
 
-        // populate count fields in pagination
-        $this->pagination['num_items'] = FrontendBlogModel::getAllForCategoryCount($requestedCategory);
-        $this->pagination['num_pages'] = (int) ceil($this->pagination['num_items'] / $this->pagination['limit']);
+    private function buildPaginationConfig(): array
```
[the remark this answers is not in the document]
carakas: yea, sounds like a good idea

src/Frontend/Modules/Blog/Actions/Category.php
```diff
@@ -56,94 +31,87 @@ public function execute(): void
...
+
+        $this->articles = FrontendBlogModel::getAllForCategory(
+            $this->category['url'],
             $this->pagination['limit'],
             $this->pagination['offset']
         );
     }
 
-    private function parse(): void
+    private function addLinkToRssFeed(): void
```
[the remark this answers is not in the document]
carakas: I think it is overkill in this case

src/Frontend/Modules/Blog/Actions/Index.php
```diff
@@ -56,17 +57,24 @@ private function getData(): void
         $this->articles = FrontendBlogModel::getAll($this->pagination['limit'], $this->pagination['offset']);
     }
 
-    private function addLinkToRssFeed(): void
+    private function addLinkToRssFeeds(): void
```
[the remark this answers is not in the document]
carakas: fixed in 57b096a0d

src/Frontend/Modules/Blog/Actions/Index.php
```diff
@@ -56,17 +57,24 @@ private function getData(): void
...
     {
+        // General rss feed
         $this->header->addRssLink(
             $this->get('fork.settings')->get($this->getModule(), 'rss_title_' . LANGUAGE),
             FrontendNavigation::getUrlForBlock($this->getModule(), 'Rss')
         );
+
+        // Rss feed for the comments of this blog
```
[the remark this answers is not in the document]
carakas: we aren't talking about a specific post/article here.
This is the recent comments across all posts/articles, so on the blog as a whole

src/Frontend/Modules/Faq/Actions/Category.php
```diff
@@ -39,32 +36,36 @@ public function execute(): void
...
 
-        // anything found?
-        if (empty($this->record)) {
+        if (empty($category)) {
             throw new NotFoundHttpException();
         }
 
-        $this->record['full_url'] = FrontendNavigation::getUrlForBlock('Faq', 'Category') . '/' . $this->record['url'];
-        $this->questions = FrontendFaqModel::getAllForCategory($this->record['id']);
+        $baseUrl = FrontendNavigation::getUrlForBlock($this->getModule(), $this->getAction());
```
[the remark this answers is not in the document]
carakas: fixed in 0d493cf14

src/Frontend/Modules/Faq/Actions/Detail.php
```diff
@@ -25,274 +25,299 @@
...
-        if ($this->status == FL::getAction('Success')) {
-            $this->status = 'success';
-        }
-        if ($this->status == FL::getAction('Spam')) {
-            $this->status = 'spam';
+    private function getStatus(): ?string
+    {
+        switch ($this->url->getParameter(2)) {
+            case FL::getAction('Success'):
+                return 'success';
```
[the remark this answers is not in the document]
carakas: it is the only place where it is used, so I think that in this case a constant would be a bit overkill, it is only used to set a template parameter with that name to true

src/Frontend/Modules/Profiles/Actions/ChangeEmail.php
carakas: see previous comment, this is how symfony handles it

src/Frontend/Modules/Search/Actions/Index.php
carakas: woops, that was debug code :)
fixed in 80fc6ae80

carakas:
feedback has been processed

PR #2286, opened 2017-10-03
Fix check if the email or displayname exists without excluding a profile

carakas:
## Type

- Non critical bugfix

## Pull request description

not all mysql servers support  this caused the check to see if the email address or display name was valid to fail sometimes


jonasdekeukelaere:
Looks like the wrong fix to me, you should build the query accordingly (with or without the id check).

tijsverkoyen:
I don't get what the bug is, and why we should pass 0 instead of null? 
I agree with Jonas, that the query should be fixed, not the parameters.

carakas:
@jonasdekeukelaere @tijsverkoyen true, it was a bit of a quickfix, improved it in 
7b52960

PR #2289, opened 2017-10-04
Remove licence info blocks in php files

carakas:
## Type

- Enhancement

## Pull request description

Since not all files have this notice and we use MIT I don't think it is relevant to keep them hanging around

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -2,13 +2,6 @@
 
 namespace ForkCMS\App;
 
-/*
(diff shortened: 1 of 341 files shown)


tijsverkoyen:
#hacktoberfest zekers?

carakas:
@tijsverkoyen don't really need this one for hacktoberfest, was working on cleaning up some code and since this is something that impacts the entire codebase I put it in a separate PR

PR #2295, opened 2017-10-06
fix showing "data" in the marker

jeroendesloovere:
## Type

- Non critical bugfix

Data wasn't being shown in the marker popup in the "Location Index Block"


jeroendesloovere:
Related to the merged https://github.com/forkcms/forkcms/pull/2257

Issue #2273, opened 2017-09-27
Bug: Redirects in Search module


Diewy:
## Type

- Non critical bug

## Problem description

After setting a redirect on a page the user would assume that the content is not accessible anymore. But a truncated version is still visible via a search on that content, which reveals all your (dirty) secrets!

## Steps to reproduce

1. Create content on a page in the backend
![screenshot 2017-09-27 10 23 09](https://user-images.githubusercontent.com/1014380/30903382-e1c1a34a-a36e-11e7-8b97-40b445a5b7ed.png)

2. Set a redirect on the same page
![screenshot 2017-09-27 10 23 18](https://user-images.githubusercontent.com/1014380/30903381-e1bfc93a-a36e-11e7-8f1d-3575f8ddc491.png)

3. Search for the created content in the frontend
![screenshot 2017-09-27 10 22 59](https://user-images.githubusercontent.com/1014380/30903380-e1be80ca-a36e-11e7-84ed-8a159ccfb523.png)

## Expected behavior / Proposed solutions

Exclude redirect pages from search

carakas:
@Diewy thank you for this bugreport, is this something you can fix by creating a PR for it?

lex111:
I would like to try to fix this. What are the recommendations on how to do this?

carakas:
@lex111 Awesome, we have a docker container with php 7.1 and the env variables you need to enable debug. So if you have docker that will be the easy road to get fork running on your machine

The search works with indexes that are created in the backend when adding or editing a page. I think it would be probably the best way to target this bug. By checking if the page is a redirect page and in that case delete any existing indexes and don't add a new one.

If you have more questions feel free to ping me or ask it on the [Fork CMS slack](https://fork-cms.herokuapp.com/)

PR #2296, opened 2017-10-06
Bug redirects in search module

ohvitorino:
## Type

- Non critical bugfix

Fixes: https://github.com/forkcms/forkcms/issues/2273

## Pull request description

This PR modifies the Add and Edit actions of the Pages modules so that it only adds pages to the search index if they aren't pages containing redirections.

### DIFF
--- a/src/Backend/Modules/Pages/Actions/Add.php
+++ b/src/Backend/Modules/Pages/Actions/Add.php
@@ -514,7 +514,7 @@ class Add extends BackendBaseActionAdd
                     }
 
 
-                    if ($redirectValue == 'none') {
+                    if ($redirectValue === 'none') {
                         // add to search index
                         BackendSearchModel::saveIndex(
                             $this->getModule(),
--- a/src/Backend/Modules/Pages/Actions/Add.php
+++ b/src/Backend/Modules/Pages/Actions/Add.php
@@ -513,12 +513,20 @@ class Add extends BackendBaseActionAdd
                         $text .= ' ' . $block['html'];
                     }
 
-                    // add to search index
-                    BackendSearchModel::saveIndex(
-                        $this->getModule(),
-                        $page['id'],
-                        ['title' => $page['title'], 'text' => $text]
-                    );
+
+                    if ($redirectValue == 'none') {
+                        // add to search index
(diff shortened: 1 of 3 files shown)


src/Backend/Modules/Pages/Actions/Add.php
```diff
@@ -513,12 +513,20 @@ private function validateForm(): void
...
                     }
 
-                    // add to search index
-                    BackendSearchModel::saveIndex(
-                        $this->getModule(),
-                        $page['id'],
-                        ['title' => $page['title'], 'text' => $text]
-                    );
+
+                    if ($redirectValue == 'none') {
```
carakas: ===

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -778,17 +778,16 @@ private function validateForm(): void
                         $text .= ' ' . $block['html'];
                     }
 
-                    // add to search index, only if authentication is false
-                    if ($data['remove_from_search_index'] == false) {
-                        BackendSearchModel::saveIndex(
+                    if ($data['remove_from_search_index'] == true || $redirectValue != 'none') {
```
carakas: !==

carakas: and if possible === true

src/Backend/Modules/Search/Engine/Model.php
```diff
@@ -137,7 +137,7 @@ public static function invalidateCache(): void
      */
     public static function removeIndex(string $module, int $otherId, string $language = null): void
     {
-        if (BackendModel::isModuleInstalled('Search')) {
+        if (!BackendModel::isModuleInstalled('Search')) {
```
carakas: nice catch

src/Backend/Modules/Pages/Actions/Add.php
ohvitorino: Fixed 67efeb4a4238e02dd44d670716629a237d3df989

src/Backend/Modules/Search/Engine/Model.php
ohvitorino: 👍

PR #2301, opened 2017-10-10
Fix sessions in the console

carakas:
## Type

- Non critical bugfix

## Pull request description

Implement the mock session for when no request is available instead of throwing an exception


src/Common/Core/Model.php
```diff
@@ -336,4 +337,13 @@ public static function getSession(): SessionInterface
 
         return $session;
     }
+
+    private static function getMockSession(): Session
+    {
+        if (!self::getContainer()->has('fork.mock.session')) {
```
tijsverkoyen: Could you at least explain what this does in a comment or the PR?

carakas: added in 52bee3e5a

PR #2302, opened 2017-10-10
Add additional blocks for Mailmotor

jeroendesloovere:
## Type

- Enhancement

In some templates the "heading/footer" blocks were missing.

## Pull request description

Added missing blocks, so heading/footer are always overwritable


carakas:

PR #2303, opened 2017-10-10
Add additional blocks for Blog module

jeroendesloovere:
## Type

- Enhancement

In some layouts, there were missing "heading/footer" blocks.


carakas:

PR #2304, opened 2017-10-10
Add additional blocks for Faq module

jeroendesloovere:
## Type

- Enhancement

Some "heading/footer" blocks were missing in the FAQ modules.


carakas:

PR #2305, opened 2017-10-10
Add additional blocks for Formbuilder module

jeroendesloovere:
## Type

- Enhancement

Sometimes you need extra "heading/footer" blocks to overwrite


carakas:

PR #2310, opened 2017-10-10
Add additional blocks for Pages module

jeroendesloovere:
## Type

- Enhancement

## Pull request description

Add additional blocks for Pages module


carakas:

PR #2313, opened 2017-10-10
Page data no longer used

cirykpopeye:
## Type

- Critical bugfix

## Pull request description
This was no longer fetched from the database, this is required for the navigation to hide the page in the nav if auth required.


carakas:
@cirykpopeye can you please update the description?

carakas:
we should test if it doesn't have side effects

PR #2261, opened 2017-09-21
CKEditor plugin for media library

ohvitorino:
## Type

- Feature

## Pull request description

At this moment we use CKFinder to enable users to include images or files in the WYSIWYG-editor.
CKFinder isn't free, each year Fork has to search for contributors to sponsor the OEM-license.

Since the MediaLibrary is part of the core of Fork it would be nice to have a plugin that replaces CKFinder and uses the MediaLibrary.

This initial implementation adds ~~a button~~ two buttons to CKEditor (insert link and insert dialog) ~~that show~~ each showing a dialog. ~~This~~ The first dialog (insert link) enables the user to insert a link either by filling it in manually or by using the button 'Browse server'. By browsing the server the user will be in fact browsing the MediaLibrary. The user can then select one of the previously uploaded files or add a new one and select it. By selecting a file, its URL will automatically filled in in the dialog.
The second dialog (insert image) works in similar way but in inserts an image in de editor. The selection is limited to images only.

~~This PR also update current version of CKEditor to version 4.7.2.~~

~~Depends on #2221~~

PR #2314, opened 2017-10-10
Load ckeditor with yarn

carakas:
## Type

- Enhancement

## Pull request description

Ckeditor is now build with gulp and installed with yarn, this will make it easy to update the editor

I also removed unused plugins


src/Backend/Core/Js/backend.js
```diff
@@ -364,10 +364,10 @@ jsBackend.ckeditor = {
     uploadUrl: '/src/Backend/Core/Js/ckfinder/core/connector/php/connector.php?command=QuickUpload&type=Files&responseType=json',
 
     // load some extra plugins
-    extraPlugins: 'stylesheetparser,mediaembed',
```
jeroendesloovere: Is this changed by intention or accidentally?

carakas: intentionally, we didn't use the mediaembed and the others needed to be activated because the build is slightly different

ohvitorino:
I'll need to integrate this in my branch https://github.com/forkcms/forkcms/pull/2261

carakas:
@ohvitorino indeed, it is made in such a way that your files won't need to move, but I think you created a custom link plugin so you'll need to stop copying that one over

src/Backend/Core/Js/backend.js
carakas: @jeroendesloovere do you need any other changes to this pr?

PR #2288, opened 2017-10-04
[WIP] Fixed Analytics in Backend: return type was wrong

jeroendesloovere:
## Type

- Critical bugfix

## Resolves the following issues

The return type is not `array`, but `Google_Service_Analytics_GaData`.

## Pull request description

GoogleClient Connector return type for getAnalyticsData was falsely set to `array`,
this must be `Google_Service_Analytics_GaData` instead.

PR #2317, opened 2017-10-13
Fix google analytics

jonasdekeukelaere:
## Type
- Critical bugfix

## Pull request description
The Connector::getAnalyticsData should return an array, but the call make in this method, returns
a Google_Service_Analytics_GaData object.

### DIFF
--- a/src/Backend/Modules/Analytics/GoogleClient/Connector.php
+++ b/src/Backend/Modules/Analytics/GoogleClient/Connector.php
@@ -4,6 +4,7 @@ namespace Backend\Modules\Analytics\GoogleClient;
 
 use Common\ModulesSettings;
 use Google_Service_Analytics;
+use Google_Service_Analytics_GaData;
 use Psr\Cache\CacheItemPoolInterface;
 
 /**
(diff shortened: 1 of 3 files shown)


jeroendesloovere:
Hi there,
This is related to https://github.com/forkcms/forkcms/pull/2288.
Return data is not an array, but is indeed Google_Service_Analytics_GaData.
I've already fixed this in a PR, but I need to update the test cases...

jeroendesloovere:
Or I can close my PR and we only accept array like you are doing.
@carakas you decide...

jonasdekeukelaere:
Fixed in 7355a7250

carakas:
@jeroendesloovere can you check this one again?

carakas:
@jonasdekeukelaere do you get errors when you run `yarn test`

jeroendesloovere:
Can anybody confirm that the stats are working?
I just deployed a fork cms v5.0.6 and the charts are not working...

schermafbeelding 2017-12-07 om 11 08 14

schermafbeelding 2017-12-07 om 11 08 20

PR #2318, opened 2017-10-13
[WIP] Sitemap generator

StijnVrolijk:
## Type

- Feature

## Pull request description

This feature will create a `sitemap.xml` file for your Fork CMS website. It works by dispatching events when certain items (items that have their own URL, specifically) change, thus altering the  sitemap. 

**This is currently a work in progress.**

The basics are there but I want might to investigate some further options such as automatic dispatching of events based on the Meta entity.

## How it works

Easy, in your controllers/command handlers just do something like this:
```
        /** @var EventDispatcher $dispatcher */
        $dispatcher = $this->get('event_dispatcher');
        $dispatcher->dispatch(SitemapItemChanged::EVENT_NAME, new SitemapItemChanged(
            'Pages', //Module
            'Page', //Entity
            1, //Id
            'en', //Language
            'news', //Slug
            '/en/news' //Url
        ));
```

The generator will handle the rest.

## TODO

- [ ] Add the sitemap to the robots.txt file
- [ ] Make sure the sitemap doesn't end up in git
- [ ] Take a look at Hreflang connections between pages
- [ ] Try to implement lastmod
- [ ] Make sure the sitemap doesn't get overwritten when deploying
- [ ] How are we going to handle the sitemap on a fresh install? 
- [ ] Add event dispatches to the core modules' CRUD
- [ ] ... ?

### DIFF
--- a/src/Backend/Modules/Settings/Js/Settings.js
+++ b/src/Backend/Modules/Settings/Js/Settings.js
@@ -98,7 +98,7 @@ jsBackend.settings =
         $statusAlert.toggleClass('hidden');
 
         // start the dot animation
-        var dotAnimation = jsBackend.settings.startDotAnimation();
+        var dotAnimation = jsBackend.settings.startDotAnimation($statusAlert);
 
         // start the action clearing
         $.ajax(
@@ -129,7 +129,7 @@ jsBackend.settings =
                 complete: function()
                 {
                     // stop the dot animation
-                    jsBackend.settings.stopDotAnimation(dotAnimation);
+                    jsBackend.settings.stopDotAnimation(dotAnimation, $statusAlert);
                     // hide the status
                     $statusAlert.toggleClass('hidden');
                     // reset the button
@@ -140,7 +140,7 @@ jsBackend.settings =
         );
     },
 
-    startDotAnimation: function (speed, dotAmount)
+    startDotAnimation: function ($element, speed, dotAmount)
     {
         // set the default speed
         if (!speed) {
@@ -152,7 +152,7 @@ jsBackend.settings =
             dotAmount = 3;
         }
 
-        var $dotsAnimation = $('[data-role="fork-dots-animation"]');
+        var $dotsAnimation = $element.find('[data-role="fork-dots-animation"]');
 
         // clear the initial content
         $dotsAnimation.text('');
@@ -170,10 +170,10 @@ jsBackend.settings =
         )
     },
 
-    stopDotAnimation: function (animation)
+    stopDotAnimation: function (animation, $element)
     {
         // clear the text
-        $('[data-role="fork-dots-animation"]').text('');
+        $element.find('[data-role="fork-dots-animation"]').text('');
 
         // clear the interval
         clearInterval(animation);
--- a/src/Backend/Modules/Settings/Js/Settings.js
+++ b/src/Backend/Modules/Settings/Js/Settings.js
@@ -15,6 +15,7 @@ jsBackend.settings =
 
         $('#testEmailConnection').on('click', jsBackend.settings.testEmailConnection);
         $('[data-role="fork-clear-cache"]').on('click', jsBackend.settings.clearCache);
+        $('[data-role="fork-generate-sitemap"]').on('click', jsBackend.settings.generateSitemap);
 
         $('#activeLanguages input:checkbox').on('change', jsBackend.settings.changeActiveLanguage).change();
     },
@@ -135,6 +136,57 @@ jsBackend.settings =
                     // reset the button
                     $clearCacheButton.on('click', jsBackend.settings.clearCache);
                     $clearCacheButton.attr('disabled', false);
+                }
+            }
+        );
+    },
+
+    generateSitemap: function (e)
+    {
+        // prevent default
+        e.preventDefault();
+
+        // save the button for later use
+        var $generateSitemapButton = $('[data-role="fork-generate-sitemap"]');
+
+        // disable the handler to prevent sending too many requests
+        $generateSitemapButton.off('click', jsBackend.settings.generateSitemap);
+        $generateSitemapButton.attr('disabled', 'disabled');
+
+        // display the status alert
+        var $statusAlert = $('[data-role="fork-generate-sitemap-status"]');
+        $statusAlert.toggleClass('hidden');
+
+        // start the dot animation
+        var dotAnimation = jsBackend.settings.startDotAnimation($statusAlert);
+
+        $.ajax(
+            {
+                timeout: 60000, // generating the sitemap might take a while
+                data: {
+                    fork: {
+                        module: 'SitemapGenerator',
+                        action: 'GenerateSitemap'
+                    }
+                },
+                success: function(data)
+                {
+                    jsBackend.messages.add('success', jsBackend.locale.msg('SitemapGenerated'));
+                },
+                error: function()
+                {
+                    // show error in case something goes wrong with the call itself
+                    jsBackend.messages.add('danger', jsBackend.locale.err('SomethingWentWrong'));
+                },
+                complete: function()
+                {
+                    // stop the dot animation
+                    jsBackend.settings.stopDotAnimation(dotAnimation, $statusAlert);
+                    // hide the status
+                    $statusAlert.toggleClass('hidden');
+                    // reset the button
+                    $generateSitemapButton.on('click', jsBackend.settings.generateSitemap);
+                    $generateSitemapButton.attr('disabled', false);
                 }
             }
         );
--- a/src/Backend/Modules/SitemapGenerator/Domain/SitemapItemChanged.php
+++ b/src/Backend/Modules/SitemapGenerator/Domain/SitemapItemChanged.php
@@ -6,6 +6,8 @@ use Symfony\Component\EventDispatcher\Event;
 
 class SitemapItemChanged extends Event
 {
+    const EVENT_NAME = 'sitemap.item_changed';
+
     /** @var string */
     private $module;
 
--- a/src/Backend/Modules/SitemapGenerator/Domain/SitemapItemChanged.php
+++ b/src/Backend/Modules/SitemapGenerator/Domain/SitemapItemChanged.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Backend\Modules\SitemapGenerator\Domain;
+
+use Symfony\Component\EventDispatcher\Event;
+
+class SitemapItemChanged extends Event
+{
+    /** @var string */
+    private $module;
+
+    /** @var string */
+    private $entity;
+
+    /** @var int */
+    private $id;
+
+    /** @var string */
+    private $slug;
+
+    /** @var string */
+    private $url;
+
+    public function __construct(string $module, string $entity, int $id, string $slug, string$url)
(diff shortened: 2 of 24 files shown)


src/Backend/Modules/Settings/Js/Settings.js
```diff
@@ -140,7 +141,58 @@ jsBackend.settings =
         );
     },
 
-    startDotAnimation: function (speed, dotAmount)
+    generateSitemap: function (e)
+    {
+        // prevent default
```
tijsverkoyen: Some comments are not really adding any value in my opinion.

src/Backend/Modules/Settings/Js/Settings.js
```diff
@@ -140,7 +141,58 @@ jsBackend.settings =
...
+                    $statusAlert.toggleClass('hidden');
+                    // reset the button
+                    $generateSitemapButton.on('click', jsBackend.settings.generateSitemap);
+                    $generateSitemapButton.attr('disabled', false);
+                }
+            }
+        );
+    },
+
+    startDotAnimation: function ($element, speed, dotAmount)
```
tijsverkoyen: Isn't there a `loading` class for buttons?

StijnVrolijk: There is but this animation has nothing to do with the button

carakas:
@StijnVrolijk can you fix the tests?

src/Backend/Modules/SitemapGenerator/info.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module>
+  <name>SitemapGenerator</name>
+  <version>1.0.0</version>
+  <requirements>
+    <minimum_version>5.0.5</minimum_version>
+  </requirements>
+  <description>
+    <![CDATA[
+      This module houses an event to throw every time something about your website changes.
```
carakas: wouldn't it be better to use contains instead of houses?

src/Backend/Modules/SitemapGenerator/EventListener/SitemapItemChangedListener.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->update($currentItem, $event);
+    }
+
+    private function update(SitemapEntry $currentItem, SitemapItemChanged $event): void
+    {
+        /** @var SitemapEntry[] $children */
+        $children = $this->sitemapEntryRepository->getChildren($currentItem);
+
+        foreach ($children as $child) {
+            $newUrl = str_replace($currentItem->getUrl(), $event->getUrl(), $child->getUrl());
```
carakas: why do you need the currentUrl here? Can you add a little explanation in comments?

jeroendesloovere:
Don't know if this PR is making me happy or not...
I'm not liking the PHP integration with events and saving into a database... :-(.
Seems like another cache to maintain and something that can go wrong.

Take a look at this Sylius Sitemap Bundle - https://github.com/stefandoorn/sitemap-plugin.
It uses "providers" to generate the different pages.

So integrating this in your custom module would just require to:
- Create a custom Provider, f.e.: "BlogArticleSitemapProvider.php"
- And adding that to your resources.

jeroendesloovere:
Using custom events will cause a pain in the ass when we start using doctrine events everywhere.
Especially when you have Entities with nested levels like f.e.: Category -> CategoryItem, CategoryProduct, ...

tijsverkoyen:
@jeroendesloovere could you elaborate on why "events" are a bad idea? In my opinion this is the best way to handle this.

jeroendesloovere:
Well, using custom events that need to be send manually is something that can go wrong very fast.
Especially when you have modules that contain nested levels and multiple variables to rely on (status, category, parent, type, publishOn, visible), if an item should be visible or not.

I've experienced this when developing the MediaLibrary module. Eventually I was even stuck and the only solution was to rebuild the module to use "doctrine events" instead. Creating "subscribers" or "listeners" to hook to the proper event.
=> This way could be used.

Or the alternative I suggest, is creating providers like this:
```
<?php

class BlogArticleSitemapProvider implements SitemapProviderInterface
{
    /** @var BlogArticleRepository */
    private $blogArticleRepository;

    public function __construct(BlogArticleRepository $blogArticleRepository)
    {
        $this->blogArticleRepository = $blogArticleRepository;
    }

    public function getRows(): SitemapRowCollection
    {
        $collection = new SitemapRowCollection();

        // Add all articles from Blog to sitemap
        foreach ($this->blogArticleRepository->findAllActive() as $blogArticle) {
            $collection->add(new SitemapRow(
                $blogArticle->getLocale(),
                $blogArticle->getTitle(),
                $blogArticle->getMeta()->getUrl(),
                ChangeFrequently::never()
            ));
        }

        return $collection;
    }
}

```

tijsverkoyen:
I don't see the point. Implementing a provider can be forgotten also.

carakas:
@jeroendesloovere this isn't running on doctrine events but symfony events, so the nested entity thing isn't an issue for that

src/Backend/Modules/SitemapGenerator/Domain/SitemapItemChanged.php
```diff
@@ -0,0 +1,68 @@
...
+    /** @var string */
+    private $language;
+
+    /** @var string */
+    private $slug;
+
+    /** @var string */
+    private $url;
+
+    public function __construct(string $module, string $entity, int $id, string $language, string $slug, string$url)
```
jeroendesloovere: Space is missing between string and $url

jeroendesloovere:
This PR can work. But it's not advanced enough. People have to integrate those custom events all around their module and they already have to do this for Search, so please don't go this way.

jeroendesloovere:
## Proposed solution:
Don't use SitemapItemChanged events.
But use `providers` instead. Have to be defined only once. A `SitemapSubscriber` fetches all events and does the rest.

## More information
- Providers, f.e.: `PageSitemapProvider`, `BlogArticleSitemapProvider`, `BlogCategorySitemapProvider`, `FaqCategorySitemapProvider`, `FaqArticleSitemapProvider`, `TagSitemapProvider`, ... They work automatically.
- We have a `sitemap subscriber` that invalidates collections of articles.

```php
<?php

namespace ForkCMS\Bundle\SitemapBundle\EventListener;

use Doctrine\Common\Persistence\Event\LifecycleEventArgs;

class SitemapSubscriber implements EventSubscriber
{
    public function onAnyEntityChangeFromAnyModule(LifecycleEventArgs $eventArgs)
    {
        // Will can then invalidate the cache for `BlogArticle::class`
        $this->get('fork.sitemap')->invalidate($eventArgs->getObject());
    }
}
```

jeroendesloovere:
I also prefer having multiple sitemaps.
That is a practice that more and more developers are using in the last years.
https://moz.com/blog/multiple-xml-sitemaps-increased-indexation-and-traffic
https://www.sitemaps.org/protocol.html#index

Old way of doing:
`sitemap.xml` - indexed 9.000 of 10.000

New way of doing things will also have the advantage that you know what is going on:
`sitemap.xml` (contains sitelinks to all other sitemaps)
`sitemap_for_pages.xml` - indexed 599 of 600
`sitemap_for_blog.xml` - indexed 5.000 of 6.000
`sitemap_for_faq.xml` - indexed 3.900 of 3.400

carakas:
@jeroendesloovere what do you do in case you have a module that isn't working with doctrine? since you are leaning heavily on doctrine?

jeroendesloovere:
Hi Stijn, I really want a sitemap feature in fork.
Just wondered the following:
- Is there any other reason why you save item in the database? Except then using it to generate the sitemap.
- What happens when you edit a "page", where a module relies on. For example: imagine that the page is called "/blog/artikel", what happens when somebody renames the "page" "blog" to "news"...?

StijnVrolijk:
@jeroendesloovere your 2nd question answers the first :)

When we edit a page (or any item for that matter) there is code in place which will edit all child URL's based on their current URL. You can see this here: https://github.com/forkcms/forkcms/blob/f89da397bd36af1302257c9259086c6d6a5126eb/src/Backend/Modules/SitemapGenerator/EventListener/SitemapItemChangedListener.php#L43

Here's an example:  

**Base table**  

module | entity | other_id | language | slug | url | created_on | edited_on
-- | -- | -- | -- | -- | -- | -- | --
Pages | Page | 231 | nl | faq | /nl/faq | 2017-10-09 12:03:23 | 2017-10-09 12:03:23
Faq | Question | 12 | nl | mijn-vraag | /nl/faq/detail/mijn-vraag | 2017-10-09 12:07:31 | 2017-10-09 12:07:31
Blog | BlogPost | 8 | nl | fork-cms | /nl/blog/detail/fork-cms | 2017-10-09 12:09:19 | 2017-10-09 12:09:19

We rename the `FAQ` page to `Vragen`

```
$children = $this->sitemapEntryRepository->getChildren($currentItem);
```

This will find all children with `/nl/faq/` in their URL. It's not the most optimal way of doing this but it is by far the simplest.

We will now loop over each child

```
$newUrl = str_replace($currentItem->getUrl(), $event->getUrl(), $child->getUrl());
```

Here we create the child's new URL by replacing the `/nl/faq/` with the `/nl/vragen/` URL we got from the event.
Afterwards we save the child.
After every child is saved, we save the parent.

**End result**

module | entity | other_id | language | slug | url | created_on | edited_on
-- | -- | -- | -- | -- | -- | -- | --
Pages | Page | 231 | nl | vragen | /nl/vragen | 2017-10-09 12:03:23 | 2017-10-21 12:03:23
Faq | Question | 12 | nl | mijn-vraag | /nl/vragen/detail/mijn-vraag | 2017-10-09 12:07:31 | 2017-10-21 17:32:02
Blog | BlogPost | 8 | nl | fork-cms | /nl/blog/detail/fork-cms | 2017-10-09 12:09:19 | 2017-10-09 12:09:19

If we don't save this in a table I would have to request _each and every_ URL from every module so saving it in a table sounds a lot more efficient. It also helps with debugging (you can instantly compare what is in the sitemap table and what is in the other tables (pages for example)) and bugfixing: If a customer complains something is wrong you can just alter the table, regenerate the sitemap and then focus on fixing the actual problem

StijnVrolijk:
There's still the issue of what happens in the next scenario:

**Start**
module | entity | other_id | language | slug | url | created_on | edited_on
-- | -- | -- | -- | -- | -- | -- | --
Pages | Page | 231 | nl | vragen | /nl/vragen | 2017-10-09 12:03:23 | 2017-10-21 12:03:23
Faq | Question | 12 | nl | mijn-vraag | /nl/vragen/detail/mijn-vraag | 2017-10-09 12:07:31 | 2017-10-21 17:32:02
Blog | BlogPost | 8 | nl | fork-cms | /nl/blog/detail/fork-cms | 2017-10-09 12:09:19 | 2017-10-09 12:09:19

We add a _new_ page called `Foo` and add an `FAQ` module on it. How are we going to know which questions we need to add URL's for? 

First thing I can come up with is that every frontend action (or even widget) should implement a static method that returns the partial url (such as `detail/my-question`) of each of the items it's going to display. That of course means even more implementing of code...

It's quite a complicated matter, this sitemap generator

PR #2319, opened 2017-10-16
Fix bootstrap tabs html5 form validation

carakas:
## Type

- Non critical bugfix

## Pull request description

If the first invalid field is in an inactive tab the user doesn't see the error
This pr fixes that

### DIFF
--- a/src/Backend/Core/Js/backend.js
+++ b/src/Backend/Core/Js/backend.js
@@ -1211,6 +1211,19 @@ jsBackend.forms =
         jsBackend.forms.tagsInput();
         jsBackend.forms.meta();
         jsBackend.forms.datePicker();
+        jsBackend.forms.bootstrapTabFormValidation();
+    },
+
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Js/backend.js
```diff
@@ -1211,6 +1211,19 @@ jsBackend.forms =
         jsBackend.forms.tagsInput();
         jsBackend.forms.meta();
         jsBackend.forms.datePicker();
+        jsBackend.forms.bootstrapTabFormValidation();
+    },
+
+    bootstrapTabFormValidation: function () {
```
tijsverkoyen: What if there are multiple invalid fields?

carakas: it will always go to the first invalid field it finds, if there are multiple you can still correct them or you go to that tab the next attempt to submit.

PR #2320, opened 2017-10-16
Implement the bootstrap accessibility plugin

carakas:
## Type

- Enhancement
- Feature

Solves a number of accessibility issues with bootstrap

## Pull request description

Implement the [accessibilty plugin for bootstrap](https://github.com/paypal/bootstrap-accessibility-plugin)

### Components
- Alert
- Tooltip
- Popover
- Modal dialog
- Dropdown menu
- Tab panel
- Collapse
- Carousel


src/Backend/Core/Layout/Templates/FormLayout.html.twig
```diff
@@ -63,44 +53,18 @@
...
-
-
 {%- block label_asterisk -%}
-  <abbr title="{{ 'lbl.RequiredField'|trans({}, translation_domain)|ucfirst }}">*</abbr>
+  {% import 'Layout/Templates/macros.html.twig' as macro %}
+  {{ macro.required }}
 {%- endblock label_asterisk -%}
 
 {% block form_errors -%}
+  {% import "Layout/Templates/macros.html.twig" as macro %}
```
Katrienvh: you only have to import it once per file, i see multiple import in this file. I should move the import to the top of the file, out of the block

src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/form_theme.html.twig
```diff
@@ -202,7 +202,7 @@
     {% if form.parent %}<span class="help-block errorMessage">{% else %}<div class="alert alert-danger errorMessage">{% endif %}
     <ul class="list-unstyled">
 {%- for error in errors -%}
-  <li><span class="fa fa-exclamation-circle"></span> {{ error.message }}</li>
+  <li><span class="fa fa-exclamation-circle" aria-hidden="true"></span> {{ error.message }}</li>
```
Katrienvh: can't you use the macro icon here?

src/Backend/Core/Layout/Templates/FormLayout.html.twig
carakas: Does that also work for FormLayout files? Because I think I moved it in here because I was getting errors otherwise

src/ForkCMS/Bundle/InstallerBundle/Resources/views/Installer/form_theme.html.twig
carakas: I didn't do it because the macro files are bound with the backend and the frontend, and this is the installer bundle, it is something that stands apart so I didn't use it here. If you think I should use a macro here: do I use the back or frontend one?

Katrienvh: Mmh ok i understand why you didn't you it, i am not sure anymore if you should change it, maybe not. But if we change it, I should use the backend one

src/Backend/Core/Layout/Templates/FormLayout.html.twig
Katrienvh: I not 100% sure, but i never needed to do it the way you did above

StijnVrolijk: @carakas could you test this again? I remember having to do something like this for macro's as well, but it was inside of a new macro.

Example:
```
{% macro jobs(jobs) %}
  {% import _self as jobMacros %}
  {% for job in jobs %}
    {% if job.score %}
      {{ jobMacros.job(job.entity, job.score) }}
    {% else %}
      {{ jobMacros.job(job) }}
    {% endif %}
  {% endfor %}
{% endmacro %}
```

carakas: I tried, if you place it in the file instead of the block you get
```
An exception has been thrown during the rendering of a template ("Notice: Undefined index: macro").
```

StijnVrolijk: Very well

PR #2321, opened 2017-10-16
Mailmotor SubscribeType/UnsubscribeType: ucfirst was missing in label

jeroendesloovere:
## Type

- Enhancement

## Pull request description

In the Frontend Mailmotor templates, the button label did not had a capital letter.
This PR fixes that.

### DIFF
--- a/src/Frontend/Modules/Mailmotor/Domain/Subscription/SubscribeType.php
+++ b/src/Frontend/Modules/Mailmotor/Domain/Subscription/SubscribeType.php
@@ -76,7 +76,7 @@ class SubscribeType extends AbstractType
             'subscribe',
             SubmitType::class,
             [
-                'label' => 'lbl.Subscribe',
+                'label' => ucfirst(Language::lbl('Subscribe')),
             ]
         );
     }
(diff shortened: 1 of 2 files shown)


src/Frontend/Modules/Mailmotor/Domain/Subscription/SubscribeType.php
```diff
@@ -76,7 +76,7 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             'subscribe',
             SubmitType::class,
             [
-                'label' => 'lbl.Subscribe',
+                'label' => ucfirst(Language::lbl('Subscribe')),
```
carakas: wrong place to fix this, this should be, if needed fixed in the form templates not here

carakas:
Fixed by https://github.com/forkcms/forkcms/pull/2323

jeroendesloovere:
Did you refer to a wrong PR?
I just wanted to have "Send" instead of "send" in the button.
The other text labels do have "ucfirst", so why can't is use ucfirst on that button label?

carakas:
Most things use SpoonFilter:ucfirst because that is multi byte safe

carakas:
but apart from that, this is symfony form, so shouldn't this be fixed in the form template so it is fixed for all buttons

PR #2327, opened 2017-10-18
Short term search implementation

wouterverstuyf:
## Type

- Non critical bugfix
- Enhancement

Make it possible to search on short terms (less than 3 characters)

### DIFF
--- a/src/Frontend/Modules/Search/Engine/Model.php
+++ b/src/Frontend/Modules/Search/Engine/Model.php
@@ -125,6 +125,12 @@ class Model
             // get all terms to search for (including synonyms)
             $terms = self::getSynonyms((string) $term);
 
+            // search on short terms
+            $shortSearchTerm = implode('%', $terms);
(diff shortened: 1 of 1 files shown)


src/Frontend/Modules/Search/Ajax/Livesuggest.php
```diff
@@ -221,7 +221,7 @@ private function parsePagination(): void
             if ($useQuestionMark) {
                 $url = $this->pagination['url'] . '?page=' . ($this->pagination['requested_page'] - 1);
             } else {
-                $url = $this->pagination['url'] . '&amp;page=' . ($this->pagination['requested_page'] - 1);
```
carakas: what bug does this solve?

wouterverstuyf: Go to http://demo.fork-cms.com/nl/zoeken and search fe on lor, click on page 2, you will see that the paging (generated by ajax) will not work because of the &amp; in the url

PR #2329, opened 2017-10-18
[WIP] Sitemap module

jeroendesloovere:
## Type

- Enhancement

## Pull request description

This is amazing. Enabling the sitemap module will automatically update your sitemaps.
It's so easy to integrate in every module with providers.

### Integrated in the following core modules (as a proof of working concept):
* Pages

### Integrate in custom modules

For example a Realizations module - which uses Doctrine entities:
1. Create a `Backend\Modules\Realizations\Provider\RealizationSitemapProvider.php` class.

```php
<?php

namespace Backend\Modules\Realizations\Provider;

use Backend\Core\Engine\Model;
use Backend\Modules\Realizations\Domain\Realization\Realization;
use Backend\Modules\Realizations\Domain\Realization\Status;
use Backend\Modules\Sitemap\Domain\SitemapRow\ChangeFrequency;
use Backend\Modules\Sitemap\Domain\SitemapRow\SitemapRow;
use Backend\Modules\Sitemap\Domain\SitemapRowCollection\SitemapRowCollection;
use Backend\Modules\Sitemap\Provider\SitemapProviderInterface;
use Frontend\Core\Engine\Navigation;

class RealizationSitemapProvider implements SitemapProviderInterface
{
    public function getEntityClass(): string
    {
        return Realization::class;
    }

    public function getRows(string $locale): SitemapRowCollection
    {
        $realizationRepository = Model::getContainer()->get('realizations.repository.realization');

        $collection = new SitemapRowCollection();
        $parentLink = Navigation::getUrlForBlock('Realizations', 'Realization', $locale);
        $realizations = $realizationRepository->findBy(
            ['status' => Status::active()],
            ['sequence' => 'ASC']
        );

        /** @var Realization $realization */
        foreach ($realizations as $realization) {
            $collection->add(new SitemapRow(
                $parentLink . '/' . $realization->getMeta()->getUrl(),
                $realization->getEditedOn(),
                ChangeFrequency::monthly()
            ));
        }

        return $collection;
    }
}
```

2. Register it in `Resources\config\services.yml` somewhere like this:
```yaml
services:
    realizations.sitemap_provider.realization:
        class: Backend\Modules\Realizations\Provider\RealizationSitemapProvider
        tags:
            - { name: sitemap.provider }
```

Done.
On every entity change (postPersist, postUpdate and postRemove), the sitemap-index will be updated and the sitemap for this module will be generated again.

## The result

`sitemap.xml` - I've integrated "Pages" module and custom "Realization" module, "News" module (Article and Category).
schermafbeelding 2017-10-18 om 09 38 39

`sitemap_nl_Pages.xml`
schermafbeelding 2017-10-18 om 09 39 18


jeroendesloovere:
Another approach then the PR https://github.com/forkcms/forkcms/pull/2318

carakas:
how does this work for database changes without doctrine?

StijnVrolijk:
Implementing this in the current state of Fork will cause incomplete sitemaps because there is no integration with the 'oldschool' modules. Incomplete sitemaps are worse than no sitemaps at all.

imo

jeroendesloovere:
There is zero problem, as I integrated this in the Pages module for you guys to see how easy it is.
It works right out of the box.

For modules like "Pages", on every Add/Edit/Delete/Copy, the sitemap build is rebuild.
This is how I integrated it:
- https://github.com/forkcms/forkcms/pull/2329/commits/423948a27e223a06e6297e055886d33910b00a78
- https://github.com/forkcms/forkcms/pull/2329/commits/429a938757809f832a3808dd8ff9e3f538484bb4

But in other modules like blog that could be like:
```php
if (BackendModel::has('sitemap.builder')) {
    // We rebuild the entire cache, because other modules can depend on pages
    BackendModel::get('sitemap.builder')->buildCacheForEntityClass('BlogArticle');
}
```
+ Then add a `SitemapProvider`, like I described in my original PR description.

jeroendesloovere:
Integrating in modules that use Doctrine is even easier. Just add the `Provider`, that's all.

carakas:
interesting

carakas:
Apart from the code I would like to add one thing: I would assume you were up to no good if you would speak to me like you wrote your description, it sounds very American car salesman like

jeroendesloovere:
@StijnVrolijk Did you already figured out how to "generate the sitemap" after a "Capistrano deployment"?

StijnVrolijk:
@jeroendesloovere why would you want to generate the sitemap after deploying? Deploying doesn't change anything to your sitemap.

But in case you really want to do it, you can just call your console command in a Capistrano task

jeroendesloovere:
@StijnVrolijk, because the "sitemap" is a variable file that should not be included into your "git repository". So deploying a new release will also have "no sitemap" at start.

Yes, I'm looking to create that console command in Capistrano deployment.

ohvitorino:
@jeroendesloovere isn't it possible to set that file as a shared file in your Capistrano config? So it doesn't get deleted on deployment and instead symlinked. Or does it only work for directories?

StijnVrolijk:
I agree with @ohvitorino, that's what I was planning on doing. Potentially doing it as 'hacky' as placing the sitemap file(s) in the shared folder, deploying and returning the file(s) to their original location

jeroendesloovere:
@ohvitorino, @StijnVrolijk Moving to `shared` and then back to `current` is indeed a possible solution.
But probably better for your PR then for mine.

Because in my PR, I'm depending on PHP to generate URL's. So if you decide to restructure the URL for your module. Keeping the shared sitemaps isn't a solution then. Because they must be re-generated.

jeroendesloovere:
@StijnVrolijk 
That seems like a big disadvantage of using Events.
When you decide to restructure your URL structure for any reason/addition, you will need to trigger SitemapChangeEvent for every Entity item, right?

StijnVrolijk:
@jeroendesloovere how would you 'restructure' your URL structure? What do you even mean by that? Do you mean just editing a page? Because that wont be a problem

jeroendesloovere:
Add this Capistrano task to your deploy.rb to build the sitemap after deployment.
```
## Generate sitemap
namespace :sitemap do
  desc "Build the sitemap"
  task :build do
    on roles(:db) do
      execute "cd #{current_path} && bin/console sitemap:build"
    end
  end
end

before "deploy:cleanup", "sitemap:build"
after "deploy:rollback", "sitemap:build"
```

tijsverkoyen:
You can symlink files with Capistrano. It is done for parameters.yml.

jeroendesloovere:
Note to self, remove the custom "Sitemap" classes by some existing library/bundle...

PR #2332, opened 2017-10-18
Make it possible to reply to a blog comment

### LINKED ISSUE #2004: There is no way to reply as site owner to a blog comment
## Type

- Feature

## Problem description

People can leave comments on your blog but there is no way to reply to them as owner.

It would be nice if it was added and send a notification to the person that posted the comment

carakas:
## Type

- Feature

fixes #2004 

## Pull request description
This PR adds the option for authors to reply to a blog post comment, the person that left the comment will also be notified
Frontend:
screen shot 2017-10-18 at 14 33 10

Backend:
screen shot 2017-10-18 at 14 33 27

### DIFF
--- a/src/Backend/Modules/Blog/Actions/EditComment.php
+++ b/src/Backend/Modules/Blog/Actions/EditComment.php
@@ -55,7 +55,12 @@ class EditComment extends BackendBaseActionEdit
             'website' => $this->form->getField('website')->isFilled()
                 ? $this->form->getField('website')->getValue() : null,
             'text' => $this->form->getField('text')->getValue(),
+            'data' => null,
         ];
+
+        if ($this->form->getField('enableReplyAndSendNotificationToAuthor')->isFilled()) {
+            $comment['data']['reply'] = $this->form->getField('reply')->getValue();
+        }
 
         BackendBlogModel::updateComment($comment);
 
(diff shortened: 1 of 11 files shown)


StijnVrolijk (approved):

tijsverkoyen:
If I understand it correctly, you can only reply from the backend? 
In my opinion this doesn't add real value, it would be more useful to implement a threaded-comment-system. So everyone can reply on specific comments instead of only the users with access to the backend.

carakas:
@tijsverkoyen  the added value here is that it allows site owners to reply to a comment and actually notify the person that left the comment automatically.
A full threaded comment system is not something we can implement in an easy way because that would most likely result in database changes, we can always create an issue for that to address it in fork 6 but I'm not sure how relevant this is.

This is something I've actually heard people say they missed it or actually assumed it was possible.

tijsverkoyen:
In my personal opinion: The blog-module is around since the start of Fork CMS, in that time nobody has asked met to be able to reply on a specific comment as an author. A threaded commenting-system is something that was requested.

So in my opinion, this PR does not add a lot of value, and I would be more enthusiastic for a threaded system, which resolved the problem you are trying to solve here also.

carakas:
@forkcms/core-contributors @forkcms/backenders @forkcms/frontenders any other opinions?

mathiashelin:
I'd go for threaded comments or nothing add all.

PR #2338, opened 2017-10-26
Use fieldsets to improve accessibility

Katrienvh:
## Type

- Enhancement

## Pull request description

Use fieldsets to improve accessibility


src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
```diff
@@ -225,7 +225,7 @@
             </div>
             <div class="row">
               <div class="col-md-12">
-                <div class="panel panel-default jsCheckboxDropdownCombo" data-multiple="true">
+                <div class="panel panel-default jsCheckboxDropdownCombo last" data-multiple="true">
```
carakas: no fieldset here?

Katrienvh: I have not done it yet because visibly it was better in a panel.

PR #2342, opened 2017-10-31
Copy location widgets

### LINKED ISSUE #2324: Copy Location widgets when copy pages to other languages
## Type

- Enhancement

## Problem description

if you run /pages/copy?from=en&to=nl for example, it will copy alle the pages with content, content blocks, ... 
It should be very nice if the location widgets would also be copied. I think most of the time u use the location widget the location won't change when the language of the location changes.

StijnVrolijk:
## Type

- Enhancement

fixes #2324

## Pull request description

When copying pages from 1 locale to another the location widgets on these pages should also be copied.


src/Backend/Modules/Location/Command/CopyLocationWidgetsToOtherLocaleHandler.php
```diff
@@ -0,0 +1,52 @@
...
+use Backend\Core\Engine\Model;
+use Common\ModuleExtraType;
+use SpoonDatabase;
+
+final class CopyLocationWidgetsToOtherLocaleHandler
+{
+    public function handle(CopyLocationWidgetsToOtherLocale $copyLocationWidgetsToOtherLocale): void
+    {
+        /** @var SpoonDatabase $database */
+        $database = Model::get('database');
```
carakas: why don't you use the constructor with dependency injection?

jeroendesloovere:
@StijnVrolijk 

Did you now that you do not copy the location entities with this PR?
You just copy the module extra... Or is that your intention?

PR #2349, opened 2017-11-06
Location(Setting) entity

StijnVrolijk:
## Type

- Enhancement

## Pull request description

Convert the Location module to use a Doctrine entity instead of an array. This could be useful in the future when we want to implement stuff based on Events (such as a sitemap, varnish cache, ...).

The public signature of the model stays exactly the same for now so we don't have to change anything to the actions.


src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,295 @@
...
+namespace Backend\Modules\Location\Domain;
+
+use Common\Locale;
+use Backend\Core\Language\Locale as BackendLocale;
+use Frontend\Core\Language\Locale as FrontendLocale;
+use DateTime;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Table(name="location")
```
carakas: you are going to switch from myissam to innodb here because that is the default in doctrine

tijsverkoyen: Moet die table-name gespecificeerd worden, of doe je het voor backwords-compatible te zijn?

src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * @var Locale
+     *
+     * @ORM\Column(type="locale", name="language")
+     */
+    private $locale;
+
+    /**
+     * @TODO: map this to some sort of Extra entity
```
tijsverkoyen: Dit snap ik niet.

src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,295 @@
...
+    /**
+     * @ORM\PrePersist
+     */
+    public function prePersist(): void
+    {
+        $this->createdOn = $this->editedOn = new DateTime();
+    }
+
+    /**
+     * @ORM\PreUpdate()
```
carakas: you don't need the ()

src/Backend/Modules/Location/Domain/Location.php
tijsverkoyen: Misschien uitleggen in comments wat deze properties is/doet

src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,295 @@
...
+     *
+     * @ORM\Column(type="locale", name="language")
+     */
+    private $locale;
+
+    /**
+     * @TODO: map this to some sort of Extra entity
+     * @var integer
+     *
+     * @ORM\Column(type="integer", name="extra_id")
```
tijsverkoyen: Reden dat je column-name specifiek meegeeft?

src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,295 @@
...
+        Locale $locale,
+        string $title,
+        string $street,
+        string $number,
+        string $zip,
+        string $city,
+        string $country,
+        float $latitude,
+        float $longitude,
+        bool $showInOverview = true,
```
tijsverkoyen: Niet logischer om die niet in constructor te steken? Maar met zo static-methods te werken:
`public function createHiddenInOverview()`?

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -24,17 +27,21 @@ class Model
...
+        $repository = BackendModel::get('doctrine.orm.default_entity_manager')->getRepository(Location::class);
 
-        // get item
-        $item = self::get($id);
+        $location = $repository->find($id);
 
-        BackendModel::deleteExtraById($item['extra_id']);
+        if ($location instanceof Location) {
+            BackendModel::deleteExtraById($location->getExtraId());
+            BackendModel::getContainer()->get('database')->delete('location_settings', 'map_id = ?', [$location->getId()]);
```
tijsverkoyen: Dit ook omvormen naar een Entity?
Ook niet logischer om bovenstaande in Entity zelf af te handelen?

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -79,11 +87,16 @@ public static function get(int $id): array
...
     {
-        return (array) BackendModel::getContainer()->get('database')->getRecords(
-            'SELECT i.*
-             FROM location AS i
-             WHERE i.language = ? AND i.show_overview = ?',
-            [BL::getWorkingLanguage(), true]
+        /** @var EntityRepository $repository */
+        $repository = BackendModel::get('doctrine.orm.default_entity_manager')->getRepository(Location::class);
+
+        $locations = $repository->findAll();
```
tijsverkoyen: Gaat dit geen issue geven als er geen locations zijn?

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -24,17 +27,21 @@ class Model
      */
     public static function delete(int $id): void
     {
-        // get database
-        $database = BackendModel::getContainer()->get('database');
+        /** @var EntityRepository $repository */
+        $repository = BackendModel::get('doctrine.orm.default_entity_manager')->getRepository(Location::class);
```
carakas: maybe an idea to create a private static method to get this repository? Since it is a recurring piece of code

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -196,41 +209,50 @@ public static function getMapSettings(int $mapId): array
...
+        $extraId = BackendModel::insertExtra(
             ModuleExtraType::widget(),
             'Location',
             'Location'
         );
 
-        // insert new location
-        $item['created_on'] = $item['edited_on'] = BackendModel::getUTCDate();
-        $item['id'] = $database->insert('location', $item);
+        $location->update(
```
tijsverkoyen: Niet logischer om hier de volledige entity aan mee te geven?

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: It's to be backwards compatible but it's not completely necessary (there are database changes anyway) so I might remove these

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: It's to be backwards compatible but it's not completely necessary (there are database changes anyway) so I might remove these

src/Backend/Modules/Location/Tests/Engine/ModelTest.php
```diff
@@ -0,0 +1,132 @@
...
+namespace Backend\Modules\Location\Tests\Engine;
+
+use Backend\Modules\Location\Engine\Model;
+use Common\WebTestCase;
+
+class ModelTest extends WebTestCase
+{
+    public function setUp()
+    {
+        parent::setUp(); // TODO: Change the autogenerated stub
```
carakas: todo?

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: Well the 'modules_extras' table contains some information that could be transformed to an entity. If that is transformed we can easily map the entities together instead of using an id.

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: I don't completely follow with what you're saying here

src/Backend/Modules/Location/Engine/Model.php
StijnVrolijk: If we convert the settings to an entity we can easily delete this as well with some doctrine mapping. I chose not to convert the settings to an entity (yet) because it's outside of the current scope.

src/Backend/Modules/Location/Domain/Location.php
tijsverkoyen: k, so this could be changed if we switch to all Entities throughout-fork. In that case 👍

tests/data/test_db.sql
```diff
@@ -2218,21 +2218,21 @@ DROP TABLE IF EXISTS `location`;
...
-  `created_on` datetime NOT NULL,
-  `edited_on` datetime NOT NULL,
+  `lat` double NOT NULL,
+  `lng` double NOT NULL,
+  `show_overview` tinyint(1) NOT NULL DEFAULT 1,
+  `created_on` datetime NOT NULL COMMENT '(DC2Type:datetime)',
+  `edited_on` datetime NOT NULL COMMENT '(DC2Type:datetime)',
   PRIMARY KEY (`id`)
-) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```
carakas: If you change this one to MyISAM I think it will be backwards compatible, double is just a slightly bigger float.
And then we can when everything is moved over remove all those things at once

src/Backend/Modules/Location/Engine/Model.php
carakas: No, it will just return an empty array

src/Backend/Modules/Location/Engine/Model.php
carakas: fair point

src/Backend/Modules/Location/Domain/Location.php
tijsverkoyen: As discussed, not that relevant.

src/Backend/Modules/Location/Domain/Location.php
carakas: @StijnVrolijk I disagree, you can do this perfectly in a BC way

src/Backend/Modules/Location/Engine/Model.php
StijnVrolijk: Nope, it always returns an array so the array map will just be skipped if it's empty

src/Backend/Modules/Location/Domain/Location.php
carakas: see other comment

tijsverkoyen: Shouldn't be an issue

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: @carakas you're going to have the doctrine comments either way so unless that doesn't break BC it can't be avoided

carakas: that doesn't break BC. If they are not there, doctrine will just keep saying that you need to update your schema to add the comments. It can work perfectly without them

src/Backend/Modules/Location/Installer/Installer.php
```diff
@@ -19,6 +21,8 @@ public function install(): void
         $this->configureBackendNavigation();
         $this->configureBackendRights();
         $this->configureFrontendExtras();
+
+        Model::get('fork.entity.create_schema')->forEntityClass(Location::class);
```
jeroendesloovere: Can you please create a method `$this->configureEntities();` and move your code into that new method.
You best add that method call just below the `importLocale(...)`. Then it is in line with all other modules in fork cms core.

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -261,30 +283,47 @@ public static function setMapSetting(int $mapId, string $key, $value): void
...
-                'data',
-                [
-                    'id' => $item['id'],
-                    'extra_label' => \SpoonFilter::ucfirst(BL::lbl('Location', 'Core')) . ': ' . $item['title'],
-                    'language' => $item['language'],
-                    'edit_url' => BackendModel::createUrlForAction('Edit') . '&id=' . $item['id'],
-                ]
-            );
+        /** @var EntityRepository $repository */
+        $repository = BackendModel::get('doctrine.orm.default_entity_manager')->getRepository(Location::class);
```
jeroendesloovere: Why don't you just create an alias for this?
`$repository = BackendModel::get('locations.repository.location');`

And in `Backend/Modules/Locations/Resources/config/repositories.yml` add:
```yaml
services:
    locations.repository.location:
        class: Backend\Modules\Locations\Domain\Location\ Location Repository
        factory: ["@doctrine.orm.entity_manager", getRepository]
        arguments:
            - Backend\Modules\Locations\Domain\Location\Location
```

jeroendesloovere:
I still got issues with the "backwards compatibility" maintaining.

Why?
It is easy to provide a `update script` between two major Fork CMS Versions...

Half baked solutions
I know that the need for Doctrine is big.
Building modules from the ground up, would go faster than trying to build workarounds, just because we currently have some fields that we want to maintain...

For these easy modules "ContentBlocks" or "Location" it can work... but for those big modules it will be a pain in the ass and a waste of time to make them backwards compatible.

carakas:
@jeroendesloovere if the BC is too much of an issue we'll create a 6 branch and release it when everything works properly and remove spoon database completely

carakas:
but for now we can do it step by step when the BC is not that hard and just flip the switch all at once

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: We're going to have to switch to InnoDB any way because we need relations (not now but in the future)

src/Backend/Modules/Location/Engine/Model.php
StijnVrolijk: 3f7a987

StijnVrolijk:
Changed it to WIP because I still have to change the Frontend model

StijnVrolijk:
WIP again: Im converting the settings to an entity aswell

StijnVrolijk:
@jeroendesloovere @carakas @tijsverkoyen I think I've finished

src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,330 @@
...
+    {
+        return $this->latitude;
+    }
+
+    public function getLongitude(): float
+    {
+        return $this->longitude;
+    }
+
+    public function isShowInOverview(): ?bool
```
carakas: both in the constructor and the update the showInOverview can only be a boolean, so I don't think we need a ? here

src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,330 @@
...
+            $locale,
+            $item['title'],
+            $item['street'],
+            $item['number'],
+            $item['zip'],
+            $item['city'],
+            $item['country'],
+            $item['lat'],
+            $item['lng'],
+            isset($item['show_overview']) ? (bool) $item['show_overview'] : true,
```
carakas: `(bool) ($item['show_overview'] ?? true)`

src/Backend/Modules/Location/Domain/Location.php
```diff
@@ -0,0 +1,330 @@
...
+            isset($item['extra_id']) ? $item['extra_id'] : null
```
carakas: `$item['extra_id'] ?? null`

src/Backend/Modules/Location/Domain/LocationSetting.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        return $this->name;
+    }
+
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    public function update($value): void
```
carakas: wouldn't setValue or set not be a more logical name?
or changeValue

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -2,88 +2,63 @@
...
+            BackendModel::deleteExtraById($location->getExtraId());
 
-        BackendModel::deleteExtraById($item['extra_id']);
+            $entityManager = self::getEntityManager();
 
-        // delete location and its settings
-        $database->delete('location', 'id = ? AND language = ?', [$id, BL::getWorkingLanguage()]);
-        $database->delete('location_settings', 'map_id = ?', [$id]);
+            $entityManager->remove($location);
+            $entityManager->flush();
```
carakas: is there a reason you are not doing `$entityManager->flush($location);` here?

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -149,142 +124,159 @@ public static function getCoordinates(
...
-                    'id' => $item['id'],
-                    'extra_label' => \SpoonFilter::ucfirst(BL::lbl('Location', 'Core')) . ': ' . $item['title'],
-                    'language' => $item['language'],
-                    'edit_url' => BackendModel::createUrlForAction('Edit') . '&id=' . $item['id'],
-                ]
-            );
+        $currentLocation = self::getLocationRepository()->find($item['id']);
+
+        if (!$currentLocation instanceof Location) {
+            return 0;
```
carakas: wouldn't it be better to throw an exception here

tests/data/test_db.sql
```diff
@@ -2217,22 +2217,22 @@ UNLOCK TABLES;
...
-  `city` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
-  `country` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
-  `lat` float DEFAULT NULL,
-  `lng` float DEFAULT NULL,
-  `show_overview` tinyint(1) NOT NULL DEFAULT '1',
-  `created_on` datetime NOT NULL,
-  `edited_on` datetime NOT NULL,
-  PRIMARY KEY (`id`)
-) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+	`id` int(11) NOT NULL AUTO_INCREMENT,
```
carakas: why tabs?

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: You are correct 1b5b3c46b

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: 310c798be

src/Backend/Modules/Location/Domain/Location.php
StijnVrolijk: 310c798be

src/Backend/Modules/Location/Domain/LocationSetting.php
StijnVrolijk: I don't think so, you're still updating the setting

src/Backend/Modules/Location/Engine/Model.php
StijnVrolijk: 9482a995c

src/Backend/Modules/Location/Engine/Model.php
StijnVrolijk: The old implementation would return `0` if the `map_id` was non-existent.

tests/data/test_db.sql
StijnVrolijk: well it's a bit all over the place in the entire file so... you want me to change them all? I'd be glad to

StijnVrolijk: Done in 88acd04e9

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -149,142 +124,159 @@ public static function getCoordinates(
...
-     * @param array $item The data of the record to insert.
-     *
-     * @return int
-     */
     public static function insert(array $item): int
     {
-        $database = BackendModel::getContainer()->get('database');
+        $location = Location::fromArray($item);
+
+        $entityManager = self::getEntityManager();
```
carakas: it would be better to do these things in the repository

src/Backend/Modules/Location/Engine/Model.php
```diff
@@ -2,88 +2,63 @@
...
-        $database = BackendModel::getContainer()->get('database');
+        $location = self::getLocationRepository()->find($id);
 
-        // get item
-        $item = self::get($id);
+        if ($location instanceof Location) {
+            BackendModel::deleteExtraById($location->getExtraId());
 
-        BackendModel::deleteExtraById($item['extra_id']);
+            $entityManager = self::getEntityManager();
```
carakas: it would be better to do these things in the repository

src/Backend/Modules/Location/Domain/Location/Location.php
```diff
@@ -0,0 +1,331 @@
...
+        string $title,
+        string $street,
+        string $number,
+        string $zip,
+        string $city,
+        string $country,
+        float $latitude,
+        float $longitude,
+        bool $showInOverview = true,
+        ?int $extraId = null
```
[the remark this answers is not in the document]
StijnVrolijk: 70b2d4220

src/Backend/Modules/Location/Domain/Location/Location.php
```diff
@@ -0,0 +1,331 @@
...
+
+    public function getEditedOn(): DateTime
+    {
+        return $this->editedOn;
+    }
+
+    public function setExtraId(int $extraId): void
+    {
+        if ($this->extraId !== null) {
+            throw new Exception('You can only set the extra ID once');
```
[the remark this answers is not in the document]
StijnVrolijk: b2468f70d

src/Backend/Modules/Location/Domain/Location/Location.php
```diff
@@ -0,0 +1,331 @@
...
+     * @ORM\PreUpdate
+     */
+    public function preUpdate(): void
+    {
+        $this->editedOn = new DateTime();
+    }
+
+    public static function fromArray(array $item): self
+    {
+        if (APPLICATION === 'Frontend') {
```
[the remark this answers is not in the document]
StijnVrolijk: Maybe, but not in this PR since it's used throughout Fork

StijnVrolijk:
@jeroendesloovere @carakas I've processed your remarks. Could you take a look?

PR #2351, opened 2017-11-08
Feedback wcag 2 0

Katrienvh:
## Type

- Enhancement

## Pull request description

Fix some accessibility issues in the markup


src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
```diff
@@ -106,7 +106,7 @@
               {# Do not change the ID! #}
               <p class="current-template">{{ 'lbl.Template'|trans|ucfirst }}:
                 <span id="tabTemplateLabel">&nbsp;</span>
-                <a data-toggle="modal" data-target="#changeTemplate">
+                <a href="#" data-toggle="modal" data-target="#changeTemplate">
```
mathiashelin: I think this should be a button-element.

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
```diff
@@ -73,7 +73,7 @@
               {# Do not change the ID! #}
               <p class="current-template">{{ 'lbl.Template'|trans|ucfirst }}:
                 <span id="tabTemplateLabel">&nbsp;</span>
-                <a data-toggle="modal" data-target="#changeTemplate">
+                <a href="#" data-toggle="modal" data-target="#changeTemplate">
```
mathiashelin: I think this should be a button-element.

carakas: shouldn't this be a button ?

src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
carakas: shouldn't this be a button ?

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
Katrienvh: screen shot 2017-11-08 at 11 02 23
it isn't a button

src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
Katrienvh: screen shot 2017-11-08 at 11 02 23
it isn't a button

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
Katrienvh: screen shot 2017-11-08 at 11 02 23
it isn't a button

src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
Katrienvh: screen shot 2017-11-08 at 11 02 23
it isn't a button

carakas: but it works like a button, didn't they say that a link always needed to go to a page or resource?

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
carakas: but it works like a button, didn't they say that a link always needed to go to a page or resource?

src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
carakas: but it works like a button, didn't they say that a link always needed to go to a page or resource?

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
carakas: but it works like a button, didn't they say that a link always needed to go to a page or resource?

src/Backend/Modules/Pages/Layout/Templates/Edit.html.twig
mathiashelin: Maybe adding `role="button"` is the best solution here?

src/Backend/Modules/Pages/Layout/Templates/Add.html.twig
mathiashelin: Maybe adding `role="button"` is the best solution here?

PR #2352, opened 2017-11-08
Fix in mailmotor email validation

jeroendesloovere:
## Type

- Non critical bugfix

## Pull request description

When you want to "subscribe/unsubscribe" in the mailmotor. First the Email validation needs to be OK before we send feedback to any gateway API (will reduce page loading time also).
But it seems that all validations are checked first.
This is not necessary and potentially give an error. Like when you use MailChimp, you get an error returned that the "mail address" is wrong.

```
     * @Assert\NotBlank(message="err.FieldIsRequired")
     * @Assert\Email
     * @MailingListAssert\EmailSubscription(groups={"is_email"})
```

### DIFF
--- a/src/Frontend/Modules/Mailmotor/Domain/Subscription/Command/Subscription.php
+++ b/src/Frontend/Modules/Mailmotor/Domain/Subscription/Command/Subscription.php
@@ -13,7 +13,7 @@ final class Subscription
      *
      * @Assert\NotBlank(message="err.FieldIsRequired")
      * @Assert\Email
-     * @MailingListAssert\EmailSubscription
+     * @MailingListAssert\EmailSubscription(groups={"is_email"})
      */
     public $email;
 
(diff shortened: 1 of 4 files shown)


carakas:
I'm not really following what you are trying to do here, looks hacky

StijnVrolijk:
@carakas he's forcing the system to only check the subscription validity once the email is filled and valid.

StijnVrolijk:
I don't think this is a viable solution either.
As far as I can tell the issue only occurs if you skip the HTML5 validation. If you're someone who has no HTML5 validation (old browser) or just turns it off, I don't care about what errors you see.

Imo this is a non-issue

carakas:
@StijnVrolijk you shouldn't rely blindly on html5 validation
But I would add comments to explain what you are doing in the code, else it will be hard to understand 6 months from now

StijnVrolijk:
@carakas no one is relying on HTML5 validation blindly. The 'problem' is just that there's some extra validation in PHP that gets called regardless of whether the email is valid or not. This extra validation takes some more time which is 'wasted' if the email is invalid because it will fail anyway

carakas:
True, I do think it needs to be fixed but as discussed with @jeroendesloovere the fix needs to be within the MailingListAssert
That one needs to ignore everything if the input isnt an email

StijnVrolijk:
That sounds like a proper solution

jeroendesloovere:
Based upon the feedback I created a better fix.
https://github.com/forkcms/forkcms/pull/2352

PR #2354, opened 2017-11-09
Fix in "truncate" Twig modifier

jeroendesloovere:
## Type

- Enhancement

* Sometimes you have a variable that can be empty. Using truncate on it then throws an error.
* I discovered this in the search module where a variable in search_index table is empty.
 
## Pull request description

Allows "NULL" in truncate method

### DIFF
--- a/src/Common/Core/Twig/Extensions/BaseTwigModifiers.php
+++ b/src/Common/Core/Twig/Extensions/BaseTwigModifiers.php
@@ -221,7 +221,7 @@ class BaseTwigModifiers
      * Truncate a string
      *    syntax: {{ $string|truncate($max-length, $append-hellip, $closest-word) }}.
      *
-     * @param string $string The string passed from the template.
+     * @param null|string $string The string passed from the template.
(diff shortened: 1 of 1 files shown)


src/Common/Core/Twig/Extensions/BaseTwigModifiers.php
```diff
@@ -221,19 +221,23 @@ public static function showBool($status, bool $reverse = false): string
...
+     * @param null|string $string The string passed from the template.
      * @param int $length The maximum length of the truncated string.
      * @param bool $useHellip Should a hellip be appended if the length exceeds the requested length?
      * @param bool $closestWord Truncate on exact length or on closest word?
      *
      * @return string
      */
     public static function truncate(
-        string $string,
+        string $string = null,
```
carakas: wrong fix, now you have an optional parameter before a required parameter, the correct typehint is
`?string $string`

jeroendesloovere: Indeed, my bad.

PR #2361, opened 2017-11-20
Blog category entity

tijsverkoyen:
## Type
- Enhancement

## Pull request description
Use Doctrine entity for the blog categories.

### DIFF
--- a/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
+++ b/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
@@ -28,7 +28,7 @@ class CategoryRepository extends EntityRepository
         $this->getEntityManager()->flush($category);
     }
 
-    public function getUrl(string $url, ?int $id): string
+    public function getUrl(string $url, int $id = null): string
     {
         $query = $this
             ->createQueryBuilder('i')
--- a/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
+++ b/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
@@ -3,10 +3,25 @@
 namespace Backend\Modules\Blog\Domain\Category;
 
 use Common\Core\Model;
+use Common\Locale;
 use Doctrine\ORM\EntityRepository;
 
 class CategoryRepository extends EntityRepository
 {
+    public function findOneByUrl(string $url, Locale $locale): Category
+    {
+        return $this
+            ->createQueryBuilder('c')
+            ->innerJoin('c.meta', 'm')
+            ->andWhere('m.url = :url')
+            ->andWhere('c.locale = :locale')
+            ->setParameter(':url', $url)
+            ->setParameter(':locale', $locale)
+            ->getQuery()
+            ->getSingleResult()
+        ;
+    }
+
     public function remove(Category $category): void
     {
         $this->getEntityManager()->remove($category);
--- a/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
+++ b/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
@@ -7,7 +7,13 @@ use Doctrine\ORM\EntityRepository;
 
 class CategoryRepository extends EntityRepository
 {
-    public function getUrl(string $url, ?int $id)
+    public function remove(Category $category): void
+    {
+        $this->getEntityManager()->remove($category);
+        $this->getEntityManager()->flush($category);
+    }
+
+    public function getUrl(string $url, ?int $id): string
     {
         $query = $this
             ->createQueryBuilder('i')
--- a/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
+++ b/src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Backend\Modules\Blog\Domain\Category;
+
+use Common\Core\Model;
+use Doctrine\ORM\EntityRepository;
+
+class CategoryRepository extends EntityRepository
+{
+    public function getUrl(string $url, ?int $id)
+    {
+        $query = $this
+            ->createQueryBuilder('i')
+            ->select('COUNT(i)')
+            ->innerJoin('i.meta', 'm')
+            ->where('m.url = :URL')
+            ->setParameter('URL', $url);
+
+        if ($id !== null) {
+            $query
+                ->andWhere('i != :category')
+                ->setParameter(
+                    'category',
+                    $this->getEntityManager()->getReference(
+                        Category::class,
+                        $id
+                    )
+                );
+        }
+
+        if ((int) $query->getQuery()->getSingleScalarResult() === 0) {
+            return $url;
+        }
+
+        return $this->getURL(Model::addNumber($url), $id);
+    }
+}
--- a/src/Backend/Modules/Blog/Engine/Model.php
+++ b/src/Backend/Modules/Blog/Engine/Model.php
@@ -106,7 +106,12 @@ class Model
         // check if this action is allowed
         if (BackendAuthentication::isAllowedAction('Settings', 'Blog')) {
             // rss title
-            if (BackendModel::get('fork.settings')->get('Blog', 'rss_title_' . BL::getWorkingLanguage(), null) == '') {
+            $rssTitle = BackendModel::get('fork.settings')->get(
+                'Blog',
+                'rss_title_' . BL::getWorkingLanguage(),
+                null
+            );
+            if ($rssTitle == '') {
                 $warnings[] = [
                     'message' => sprintf(
                         BL::err('RSSTitle', 'Blog'),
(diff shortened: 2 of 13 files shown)


src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -864,34 +803,45 @@ public static function insertCompletePost(array $item, array $meta = [], $tags =
...
+        );
 
-        // meta given?
-        if ($meta !== null) {
-            $item['meta_id'] = $database->insert('meta', $meta);
+        if ($meta === null) {
+            $meta = BackendModel::get('fork.repository.meta')
+                                ->find($item['meta_id']);
+        } else {
+            $meta = new Meta(
```
carakas: Wouldn't it be better to do this in a static constructor?

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -864,34 +803,45 @@ public static function insertCompletePost(array $item, array $meta = [], $tags =
...
+                $meta['keywords'],
+                (isset($meta['keywords_overwrite'])) ? $meta['keywords_overwrite'] : false,
```
carakas: for this one and all the ones below
`$meta['keywords_overwrite'] ?? false`

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -1070,37 +1020,48 @@ public static function update(array $item): int
...
-     * @param array       $item The new data.
-     * @param array $meta The new meta-data.
-     *
-     * @return int
-     */
-    public static function updateCategory(array $item, array $meta = null): int
-    {
-        // get database
-        $database = BackendModel::getContainer()->get('database');
+    public static function updateCategory(
```
carakas: why not on a single line?

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -1070,37 +1020,48 @@ public static function update(array $item): int
...
+        array $item,
+        ?array $meta = null
```
carakas: why the additional useless `?` again?

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -1070,37 +1020,48 @@ public static function update(array $item): int
...
+    ): void {
+        $entityManager = BackendModel::get(
```
carakas: why not just use `BackendModel::get('doctrine.orm.default_entity_manager')->flush($category)` since you are only using the entity manager once (I've seen this in other places as well)

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -1070,37 +1020,48 @@ public static function update(array $item): int
...
-            // get current category
-            $category = self::getCategory($item['id']);
+        $metaEntity = $category->getMeta();
 
-            // update the meta
-            $database->update('meta', $meta, 'id = ?', [(int) $category['meta_id']]);
+        if ($meta !== null) {
+            $metaEntity->update(
+                $meta['keywords'],
+                (isset($meta['keywords_overwrite'])) ? $meta['keywords_overwrite'] : false,
```
carakas: same remark as above

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -440,24 +440,27 @@ public static function getCategory($id): array
...
-            'SELECT i.id
-             FROM blog_categories AS i
-             WHERE i.title = ? AND i.locale = ?',
-            [$title, $language]
-        );
+        $categories = BackendModel::get('blog.repository.category')
+                                  ->findBy(
+                                      [
+                                          'title' => $title,
+                                          'locale' => $language,
```
carakas: `Locale::fromString($language ?? BL::getWorkingLanguage())`
And then you can drop the if statement

You need to use the Locale here because it isn't a string but a Locale value object

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -440,24 +440,27 @@ public static function getCategory($id): array
...
+        }
 
-        return (int) BackendModel::getContainer()->get('database')->getVar(
-            'SELECT i.id
-             FROM blog_categories AS i
-             WHERE i.title = ? AND i.locale = ?',
-            [$title, $language]
-        );
+        $categories = BackendModel::get('blog.repository.category')
+                                  ->findBy(
```
carakas: why not findOneBy?

src/Backend/Modules/Blog/Engine/Model.php
tijsverkoyen: Fixed in 5762dc28fd712cccabfb1dad7fdf4173ae61b7e4

src/Backend/Modules/Blog/Engine/Model.php
tijsverkoyen: Fixed in 5762dc28fd712cccabfb1dad7fdf4173ae61b7e4

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -440,24 +440,27 @@ public static function getCategory($id): array
...
-     * Get a category id by title
-     *
-     * @param string $title    The title of the category.
-     * @param string $language The language to use, if not provided we will use the working language.
-     *
-     * @return int
+     * @deprecated
      */
-    public static function getCategoryId(string $title, string $language = null): int
+    public static function getCategoryId(string $title, ?string $language): int
```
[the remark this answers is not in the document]
tijsverkoyen: Fixed in 5762dc28fd712cccabfb1dad7fdf4173ae61b7e4
To be honest, I don't see the point of this method. It is also never used.

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -647,56 +631,11 @@ public static function getUrl(string $url, int $id = null): string
...
 
-    /**
-     * Retrieve the unique URL for a category
-     *
-     * @param string $url The string whereon the URL will be based.
-     * @param int    $id  The id of the category to ignore.
-     *
-     * @return string
-     */
     public static function getUrlForCategory($url, int $id = null): string
```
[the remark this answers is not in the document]
tijsverkoyen: Fixed in e1a30444f2858439c12fb88659292f551b1e924a

src/Backend/Modules/Blog/Engine/Model.php
```diff
@@ -864,34 +803,45 @@ public static function insertCompletePost(array $item, array $meta = [], $tags =
...
-    /**
-     * Inserts a new category into the database
-     *
-     * @param array $item The data for the category to insert.
-     * @param array $meta The metadata for the category to insert.
-     *
-     * @return int
-     */
-    public static function insertCategory(array $item, array $meta = null): int
+    public static function insertCategory(array $item, ?array $meta = null): int
```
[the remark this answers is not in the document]
tijsverkoyen: Fixed in e1a30444f2858439c12fb88659292f551b1e924a

src/Backend/Modules/Blog/Engine/Model.php
tijsverkoyen: Fixed in e1a30444f2858439c12fb88659292f551b1e924a

src/Backend/Modules/Blog/Engine/Model.php
tijsverkoyen: Fixed in e1a30444f2858439c12fb88659292f551b1e924a
Because my editor was set to 80chars maximum per line.

src/Backend/Modules/Blog/Engine/Model.php
tijsverkoyen: Fixed in e1a30444f2858439c12fb88659292f551b1e924a

src/Backend/Modules/Blog/Engine/Model.php
tijsverkoyen: Fixed in e1a30444f2858439c12fb88659292f551b1e924a
In my opinion my way was more readable.

src/Backend/Modules/Blog/Engine/Model.php
tijsverkoyen: @carakas I will add this in a future PR where I implement the entity for the Post

src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
```diff
@@ -7,7 +7,13 @@
...
 class CategoryRepository extends EntityRepository
 {
-    public function getUrl(string $url, ?int $id)
+    public function remove(Category $category): void
+    {
+        $this->getEntityManager()->remove($category);
+        $this->getEntityManager()->flush($category);
+    }
+
+    public function getUrl(string $url, ?int $id): string
```
carakas: should be `int $id = null` since you only need the id when you want to update an existing url and otherwise you'll have to pass null when you actually don't need the second parameter

src/Frontend/Modules/Blog/Engine/Model.php
```diff
@@ -874,29 +867,25 @@ public static function notifyAdmin(array $comment): void
...
+                                  $variables,
+                                  true
+                              );
             FrontendModel::get('mailer')->send($message);
         }
     }
 
     /**
      * Parse the search results for this module
-     *
```
carakas: did you do this on purpose?

because the empty line needs to be there according to the spec

src/Backend/Modules/Blog/Domain/Category/CategoryRepository.php
tijsverkoyen: Fixed in ae614a62d315d074c1532c1f207abb2bd14d4a99

src/Frontend/Modules/Blog/Engine/Model.php
tijsverkoyen: Fixed in ae614a62d315d074c1532c1f207abb2bd14d4a99

PR #2363, opened 2017-11-21
Specify form ID to avoid deleting all entries.

tommyvdv:
## Type

- Critical bugfix

## Pull request description

As a backend user I want to to save a blog post as a draft without every item for this `id` being deleted.


tommyvdv:
solves #2364.

PR #2365, opened 2017-11-22
Feature default template image

jonasdekeukelaere:
## Type
- Enhancement

## Pull request description
Adds default image to templates for pages. As per request of @Katrienvh


src/Backend/Modules/Extensions/Actions/EditThemeTemplate.php
```diff
@@ -346,6 +351,24 @@ private function validateForm(): void
                     $item['active'] = true;
                 }
 
+                $imagePath = FRONTEND_FILES_PATH . '/Templates/images';
+
+                if ($this->record['default_image'] !== null
+                    && ($this->form->getField('remove_default_image')->isChecked()
+                        || $this->form->getField('default_image')->isFilled())
```
carakas: while the indentation is correct, when you are reading it on first glance it doesn`t look correct
maybe close that second bracket on a new line so it is more clear?

src/Backend/Modules/Pages/Actions/Add.php
```diff
@@ -550,12 +554,33 @@ private function validateForm(): void
         }
     }
 
-    private function getImage(bool $allowImage): ?string
+    private function getImage(bool $allowImage, ?string $defaultImage): ?string
```
carakas: wouldn't it make more sense to use `string $defaultImage = null`

src/Backend/Modules/Pages/Actions/Add.php
```diff
@@ -550,12 +554,33 @@ private function validateForm(): void
...
     {
-        if (!$allowImage || !$this->form->getField('image')->isFilled()) {
+        if (!$allowImage) {
             return null;
         }
 
+        if ($defaultImage === null && !$this->form->getField('image')->isFilled()) {
```
carakas: why don't you group those if statements since they all return null?

src/Backend/Modules/Pages/Actions/Add.php
jonasdekeukelaere: Data comes straight of out of database, and will always be there, so no need to add default value I think.

src/Backend/Modules/Extensions/Actions/EditThemeTemplate.php
jonasdekeukelaere: Fixed 40e3308c2

src/Backend/Modules/Pages/Actions/Add.php
jonasdekeukelaere: Fixed 40e3308c2

jonasdekeukelaere:
Oh @carakas you changed target branch, no wonder I had more changes in the files overview.

jonasdekeukelaere:
I seem to be unable to easily fix the branch issues. Do I rebase on 6.0.0-dev or do I merge 6.0.0-dev in my branch (so also getting master changes in this branch)?

carakas:
Yea, you made database changes, so it had to be the 6.0.0
Just wait a bit for now, the changes from master still need to be merged into 5.1.0 and 6.0.0 branches
I waited with that because 5.0.7 should be released tomorrow or the day after that

jonasdekeukelaere:
So rebase to 6.0.0-dev? And remove the master changes?

carakas:
check this again after #2963 is merged

PR #2366, opened 2017-11-23
Add the correct error class with html5 validation and check validity …

mathiashelin:
## Type

- Enhancement

## Pull request description

Expand the custom html5 validation so it adds the correct error class to a form-group when the input is invalid.


carakas:
@mathiashelin isn't this more a bugfix? or does this rely on a feature that will be introduced in 5.1.0 ?

mathiashelin:
No it's not a bugfix. The previous version of this validation made it possible to set custom html5 error messages, this enhancement also changes the look accordingly.

PR #2368, opened 2017-11-24
Check on existence of translation keys

wouterverstuyf:
## Type

- Non critical bugfix

When you install with just one language (English) noticed are displayed in the backend.

noticed-languages

## Pull request description

Check on the existence of the keys in the translation array.


carakas:
seems like it doesn't work

PR #2376, opened 2017-12-01
Replaced `truncate` by method from twig/extensions library

jeroendesloovere:
## Type

- Non critical bugfix

Our current `truncate` function had a parameter `bool $closestWord`. But the use of it, didn't gave the wanted result... nothing happened.

## Pull request description

Since this method also used Spoon and we don't want to maintain all this code.
I decided to replace it by a working external library.


carakas:
this is more adding new features in a non backwards compatible way than fixing a bug, so switched it to the 6.0.0

PR #2383, opened 2017-12-21
typeahead.bundle.min.js updated

jeroendesloovere:
## Type

- Non critical bugfix

A client mentioned to me that in a "city"-search field, he typed the zip "9800". Normally he had to see multiple possible cities (and php had provided them), but only the first result was shown: "Astene - 9800".

So I googled and found out this was a bloodhound bug: https://stackoverflow.com/questions/30370496/typeahead-js-bloodhound-display-just-one-result

## Pull request description

Since the latest version 0.11.1 is the one we are using, I was forced to use the latest "master" version.


carakas:
well, the moment you build gulp this is gone, so don't see why we should merge it?

jeroendesloovere:
I have zero problems if you close this one.
Now it's online as a PR, other people can use this one as a fix if they need to (if they don't use gulp).

jeroendesloovere:
I discovered the bug in an older Fork CMS version 3.9.6 where typehead.bundle.min.js was not in the gulpfile.js, so I could fix this easily.

PR #2389, opened 2017-12-29
Skip rewriting and https enabling on localhost

jessedobbelaere:
## Type
- Enhancement

/

## Pull request description
There was already an option in .htaccess to skip https and url rewriting for .dev domains. Adding localhost should fix it for when you're using docker for example. 

Also, [Chrome 63 forces .dev domains to use https](https://ma.ttias.be/chrome-force-dev-domains-https-via-preloaded-hsts/) so using .dev for local development isn't so useful anymore FYI

It should skip on these hostnames:
```
localhost
localhost:3000
subdomain.localhost:3000
```


carakas:
.dev still works if you use firefox :D but thanks for the pr

PR #2395, opened 2018-01-10
Change return type of getAction to accept NULL

cirykpopeye:
## Type
- Critical bugfix

Functionality of creating urls like: /projects/{category_url}/{item_url} was impossible.

## Pull request description
NULL was no longer accepted as a return type for getAction. That broke some of the functionality of creating urls like: /projects/{category_url}/{item_url} since that requires the action field within ModuleExtra to be NULL.


carakas:
@cirykpopeye can you give more information on this? I'm rather reluctant to change this since we would need to add some extra checks in all the places where it assumes it is a string

carakas:
if it does seem possible to get a null on line 190 (the final return) I think the appropriate solution would be to add an extra check there and set the default action if needed

PR #2401, opened 2018-01-23
In some circumstances it could happen that first() would return a boo…

jacob-v-dam:
…lean instead of an object or null. This could cause some errors in PHP.

## Type
- Non critical bugfix

## Pull request description
In some situations it could happen that false was returned instead of null. Since php 7.2 uses type hinting to check for null or a MediaItem it would cause a PHP error.

This fix checks if the object is a instance of Collection but also if there is a first item. If both are true return the first item otherwise return always false.

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Domain/MediaGroup/MediaGroup.php
+++ b/src/Backend/Modules/MediaLibrary/Domain/MediaGroup/MediaGroup.php
@@ -163,7 +163,7 @@ class MediaGroup implements JsonSerializable
     {
         $connectedMediaItems = $this->getConnectedMediaItems();
 
-        return $connectedMediaItems instanceof Collection ? $this->getConnectedMediaItems()->first() : null;
+        return $connectedMediaItems->isEmpty() ? null : $connectedMediaItems->first();
     }
 
     public function hasConnectedItems(): bool


StijnVrolijk:
I think that the actual problem is the fact there's something in the collection that's not an object

jacob-v-dam:
That is not the case. I checked the database. A MediaGroup has been set but when I check the values in the table MediaGropMediaItem it would return 0 items.

StijnVrolijk:
btw, from what I can tell the error should occur sooner. The code is also a bit stupid, tbh.

```
    public function getConnectedMediaItems(): Collection
    {
        return $this->connectedItems->map(
            function (MediaGroupMediaItem $connectedItem): MediaItem {
                return $connectedItem->getItem();
            }
        );
    }
```

Afterwards it checks if it's a collection. It alwyas is because the return type of that method says so.
Also, in the method I pasted here, the return type says all items in the collection are a `MediaItem` object.
Furthermore, the `first()` method of a `Collection` uses `reset()` internally. `reset([])` returns false. 

Therefor we can conclude that `$this->connectedMediaItems` is an empty collection.

So the proper solution is as follows:
```
return $connectedMediaItems->isEmpty() ? null : $connectedMediaItems->first();
```

jeroendesloovere:
@jacob-v-dam Can you change your PR to the solution @StijnVrolijk has provided?

jacob-v-dam:
Changed it!

PR #2405, opened 2018-01-25
Make sure we always have a valid action

carakas:
## Type
- Non critical bugfix

closes #2395 

## Pull request description
Apparently it was possible for the action to be null in some cases. This should fix that

### DIFF
--- a/src/Frontend/Core/Engine/Block/ExtraInterface.php
+++ b/src/Frontend/Core/Engine/Block/ExtraInterface.php
@@ -99,12 +99,6 @@ class ExtraInterface extends KernelLoader implements ModuleExtraInterface
 
         // load the config file for the required module
         $this->loadConfig();
-
-        // is the requested action possible? If not we throw an exception.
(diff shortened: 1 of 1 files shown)


src/Frontend/Core/Engine/Block/ExtraInterface.php
```diff
@@ -149,6 +143,10 @@ public function execute(): void
     public function getAction(): ?string
     {
         if ($this->action !== null) {
+            if (!\in_array($this->action, $this->config->getPossibleActions(), true)) {
```
StijnVrolijk: what's the backslash for?

carakas: opcode optimalisations https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048

StijnVrolijk: moh, nice, didn't know that

PR #2415, opened 2018-01-29
A11y cancel buttons

StijnVrolijk:
## Type

- Enhancement

## Pull request description
I've added cancel buttons everywhere it seemed logical in the backend.


src/Backend/Core/Layout/Sass/components/_buttons.scss
```diff
@@ -32,6 +32,14 @@
     font-size: $font-size-small;
     padding: 4px 15px 5px;
   }
+
+  &.btn-cancel {
```
Katrienvh: Why not just use btn-default?

StijnVrolijk: @Katrienvh because the font-size of `.btn-text` of buttons that are _not_ in `.content-action-bar .btn-toolbar` get a font-size of 12px while those that do get a font-size of 14px.

Also the buttons near the bottom of the page (the action buttons) got a 14px font-size as well. This meant the cancel button was the only button with a smaller font-size.

I've discussed this with @mathiashelin and he told me you'd take care of it when converting to BS4.

carakas:
my only question is is this really a 4.1.1 or a 4.2.0

PR #2416, opened 2018-01-29
Show all groups linked to a profile but fade the ones that have expired

StijnVrolijk:
## Type
- Non critical bugfix

## Pull request description
Show all groups linked to a profile but fade the ones that have expired
This allows us to edit links and re-enabled them through changing the expiry date

### DIFF
--- a/src/Backend/Modules/Profiles/Engine/Model.php
+++ b/src/Backend/Modules/Profiles/Engine/Model.php
@@ -33,14 +33,15 @@ class Model
 
     /**
      * Browse groups for datagrid.
+     * The hidden field is added to display the expired group links as greyed out
      *
      * @var string
      */
     const QUERY_DATAGRID_BROWSE_PROFILE_GROUPS =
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Profiles/Engine/Model.php
```diff
@@ -37,10 +37,10 @@ class Model
      * @var string
      */
     const QUERY_DATAGRID_BROWSE_PROFILE_GROUPS =
-        'SELECT gr.id, g.name AS group_name, UNIX_TIMESTAMP(gr.expires_on) AS expires_on
+        'SELECT gr.id, g.name AS group_name, UNIX_TIMESTAMP(gr.expires_on) AS expires_on,
+          IF(gr.expires_on IS NOT NULL AND gr.expires_on <= NOW(), 1, 0) AS hidden
```
carakas: I would add a comment here that you name it hidden to gray it out in the datagrid

StijnVrolijk: I've amended it

PR #2417, opened 2018-01-29
Improve the layout of the form to add/edit profile groups

StijnVrolijk:
## Type
- Enhancement

## Pull request description
Makes the form add or edit the links between profile groups and the profile a bit clearer.


carakas:
@StijnVrolijk this could technically be a bugfix right? since that form field was missing a label?

StijnVrolijk:
@carakas that is correct

PR #2420, opened 2018-01-30
Media Library in CKEditor - uploaded video has no selection button

ohvitorino:
## Type

- Non critical bugfix

## Pull request description
In the media library popup screen within CKEditor the user has the possibility to add new videos. At this moment if the user adds a new video he won't be able to select the video as a link because no select button is present as it is the case for images and files. This PR solves this problem by correcting the code that should be adding this button.


StijnVrolijk:
Looks good to me, @jeroendesloovere can you confirm?

PR #2422, opened 2018-01-31
Core: Copy modules to other locale

jeroendesloovere:
## Type

- Feature

## Resolves the following issues

Fixes https://github.com/forkcms/forkcms/issues/2092

## Pull request description

Sometimes you need to enable a new locale for a client `bin/console forkcms:locale:enable`, but you also have to duplicate all content to that new locale. This code provides the basics to connect every module, to a uniform copy($fromLocale, $toLocale).

## How it works

The `ForkCMS\Utility\Module\CopyContentToOtherLocale\CopyContentFromModulesToOtherLocaleManager` is the core class that executes all copy-creating. 

You can add your own `copy command` using:
```yaml
services:
    content_blocks.copy:
        class: Backend\Modules\ContentBlocks\Domain\ContentBlock\Command\CopyContentBlocksToOtherLocale
        tags:
            - { name: forkcms.copy_module_to_other_locale }
```
> This will automatically be executed if somebody uses `bin/console forkcms:locale:copy` from the other PR https://github.com/forkcms/forkcms/pull/2423.

That `CopyContentBlocksToOtherLocale` extends `ForkCMS\Component\Module\CopyModuleToOtherLocaleCommand`.
In the class you must have (since it is declared abstract):
```php
    public function getModuleName(): string
    {
        return 'ContentBlocks';
    }
```
> You can also create a `__construct()` method where you can `$this->setPriority(50);` to be executed later. This allows us to first copy other modules and then you can use the new `id` or `extraId` in the Handler.

The Handler should implement `CopyModuleToOtherLocaleCommandHandlerInterface`.
In the handler class, `handle` method, you can set `id` (old -> new) and `extraId` (old -> new).

Modules changed:
* ContentBlocks (priority is 10)
* Location (priority is 10)
* Pages (priority is 50)

PR #2423, opened 2018-01-31
Core: Copy modules to other locale: console command

jeroendesloovere:
## Type

- Feature

Fixes https://github.com/forkcms/forkcms/issues/2092 and is an addition to https://github.com/forkcms/forkcms/pull/2422

## Pull request description

This PR adds the *console command* `bin/console forkcms:locale:copy` which copies content from the modules from one locale to another.


bin/console
```diff
@@ -44,6 +44,12 @@ $application = new Application($kernel);
 $application->add(new \Console\Locale\ImportLocaleCommand());
 $application->add(new \Console\Thumbnails\GenerateThumbnailsCommand());
 if ($kernel->getContainer()->get('fork.settings') instanceof ModulesSettings) {
+    $application->add(
+        new \Console\Locale\CopyLocaleCommand(
```
WouterSioen: I'm not really sure why we're not using use statements here?

src/Backend/Modules/Pages/Resources/config/services.yml
```diff
@@ -0,0 +1,6 @@
+services:
+    fork.pages.copy_module_to_other_locale:
+        class: Backend\Modules\Pages\Command\CopyModuleToOtherLocale
+        public: true
+        tags:
+            - { name: fork.copy_module_to_other_locale }
```
WouterSioen: It looks like your editor isn't set up to add an EOF character at the end of every file. This is a best practice. 

This should btw be covered by the editorconfig plugin (if you have it installed. PS: you should have it installed 😉 )

src/Console/Locale/CopyLocaleCommand.php
```diff
@@ -0,0 +1,103 @@
...
+            'To which locale would you like to copy it?',
+            $this->getToLocale($fromLocale)
+        ));
+    }
+
+    private function copyModules(CommonLocale $fromLocale, CommonLocale $toLocale)
+    {
+        $this->copyModulesToOtherLocaleManager->copy($fromLocale, $toLocale);
+
+        $this->io->success('All modules are copied from "' . $fromLocale . '" to "' . $toLocale . '".');
```
WouterSioen: Do all modules support this copy? I don't think that's the case so I would change the returned message to "All supported Modules ..."

src/Console/Locale/CopyLocaleCommand.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * @var SymfonyStyle
+     */
+    private $formatter;
+
+    /**
+     * @var SymfonyStyle
+     */
+    private $io;
```
WouterSioen: I don't like abbreviations in code.

Also: why do we have a formatter and an io which is basically the same object?

bin/console
jeroendesloovere: Fixed

WouterSioen:
@jeroendesloovere I don't think your merge did what it was expected to do 😉

src/Backend/Modules/Pages/Resources/config/services.yml
jeroendesloovere: Fixed

src/Console/Locale/CopyLocaleCommand.php
jeroendesloovere: Fixed

src/Console/Locale/CopyLocaleCommand.php
jeroendesloovere: Indeed, removed this variable entirely since we already are using this class as `private $formatter`

jeroendesloovere:
Anybody willing to review this one? @forkcms/core-contributors

jeroendesloovere:
Ping

StijnVrolijk:
Pong.

On a serious note, I'm really busy.

carakas:
is included in #2991

PR #2424, opened 2018-01-31
Make it possible to disable ck editor

carakas:
## Type

- Feature

## Pull request description
This setting makes it possible for users to disable ck editor and use plain html instead.


src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -342,6 +342,12 @@ jsBackend.pages.extras = {
     })
 
     $('#blockHtml').unbind('show.bs.modal').on('show.bs.modal', function (e) {
+      if (jsData.Core.preferred_editor !== 'ck-editor') {
+        $('#html').val(previousContent);
```
tijsverkoyen: I don't get this. 
Why restore the previousContent?

carakas: this loads the current html in the editor

PR #2425, opened 2018-02-01
Make the cropper more accessible

carakas:
## Type

- Enhancement

## Pull request description

This pull request makes the cropper more accessible since more control is now possible with the keyboard.

I also moved the checkbox to enable the cropper above the add image button with a helper text to make it more clear that the checkbox needs to be checked before pressing that button.

screen shot 2018-02-01 at 11 44 01
screen shot 2018-02-01 at 11 43 36


tijsverkoyen:
Just a general question, why not a separate button "Crop and Add", instead of the checkbox?

carakas:
Those buttons are controlled by a different js library, I tried that initially but failed

PR #2426, opened 2018-02-01
Fix collapse navigation button for pages and media library

carakas:
## Type

- Non critical bugfix

## Pull request description

The collapse arrow was gone on the pages and media library navigation because it uses a different template


carakas:
this is already fixed in a branch @Katrienvh  is working on

PR #2438, opened 2018-02-06
Use macro for button buttons

Katrienvh:
## Type
- Enhancement

## Pull request description
All buttons with text and icon are using the same macro now.

PR #2443, opened 2018-02-08
Don't publish blogposts created as draft

StijnVrolijk:
## Type
- Non critical bugfix

## Relies on
https://github.com/forkcms/forkcms/pull/2438 (This changes the markup of the buttons used in this PR, merge first)

## Pull request description
When clicking 'Save as draft' upon creating a blog post it would be saved with the `'active'` status, indicating it's published. This was caused by faulty JS that only worked on the Edit page (see `$('form#edit')`). Rather than fixing the JS I rewrote it to not use JS at all, instead focusing on using dual submit buttons, each of which has a different result.


src/Backend/Modules/Blog/Layout/Templates/Add.html.twig
```diff
@@ -155,8 +155,8 @@
           {{ macro.buttonIcon(geturl('Index'), 'times', 'lbl.Cancel'|trans|ucfirst, 'btn-default') }}
         </div>
         <div class="btn-group pull-right" role="group">
-          {{ macro.buttonIcon('', 'file-o', 'lbl.SaveDraft'|trans|ucfirst, 'btn-default', {"type":"submit", "id":"saveAsDraft"}) }}
-          {{ macro.buttonIcon('', 'plus-square', 'lbl.Add'|trans|ucfirst, 'btn-primary', {"id":"addButton", "type":"submit", "name":"add"}) }}
+          {{ macro.buttonIcon('', 'file-o', 'lbl.SaveDraft'|trans|ucfirst, 'btn-default', {"type":"submit", "name":"saveAsDraft"}) }}
+          {{ macro.buttonIcon('', 'plus-square', 'lbl.Publish'|trans|ucfirst, 'btn-primary', {"type":"submit", "name":"add"}) }}
```
mathiashelin: `name` should be `publish` and a translation instead of hardcoded text.

src/Backend/Modules/Blog/Layout/Templates/Edit.html.twig
```diff
@@ -213,8 +213,8 @@
           {{ macro.buttonIcon(geturl('Index'), 'times', 'lbl.Cancel'|trans|ucfirst, 'btn-default') }}
         </div>
         <div class="btn-group pull-right" role="group">
-          {{ macro.buttonIcon('', 'file-o', 'lbl.SaveDraft'|trans|ucfirst, 'btn-default', {"id":"saveAsDraft", "type":"submit"}) }}
-          {{ macro.buttonIcon('', 'check', 'lbl.Publish'|trans|ucfirst, 'btn-primary', {"id":"editButton", "type":"submit", "name":"edit"}) }}
+          {{ macro.buttonIcon('', 'file-o', 'lbl.SaveDraft'|trans|ucfirst, 'btn-default', {"type":"submit", "name":"saveAsDraft"}) }}
+          {{ macro.buttonIcon('', 'check', 'lbl.Publish'|trans|ucfirst, 'btn-primary', {"type":"submit", "name":"edit"}) }}
```
mathiashelin: `name` should be `publish` and a translation instead of hardcoded text.

src/Backend/Modules/Blog/Layout/Templates/Add.html.twig
StijnVrolijk: I've altered the name to `publish` but using a translation would jeopardize the workings of the PHP bit since it can no longer rely on consistent data.

src/Backend/Modules/Blog/Layout/Templates/Edit.html.twig
StijnVrolijk: I've altered the name to publish but using a translation would jeopardize the workings of the PHP bit since it can no longer rely on consistent data.

src/Backend/Modules/Blog/Layout/Templates/Add.html.twig
mathiashelin: ok

StijnVrolijk:
Tests will succeed once #2438 is merged. They fail because of a bug in that PR (that's been fixed but I can't be bothered to patch mine again)

PR #2447, opened 2018-02-09
Catch Doctrine's DBALException if the database connection fails

StijnVrolijk:
## Type
- Critical bugfix

## Pull request description
On step 4 of the installer, the database variables are guessed from the URL of the website.
Fork then tries to connect to this database to see if it worked. If it doesn't it catches a ConnectionException but since DoctrineBundle 1.6.8 it also throws a DBALException if it can't determine the server version. We now catch that exception as well and treat it like a ConnectionException.

For more information see https://github.com/doctrine/DoctrineBundle/issues/673

### DIFF
--- a/src/ForkCMS/Bundle/InstallerBundle/Service/InstallerConnectionFactory.php
+++ b/src/ForkCMS/Bundle/InstallerBundle/Service/InstallerConnectionFactory.php
@@ -6,6 +6,7 @@ use Doctrine\Bundle\DoctrineBundle\ConnectionFactory;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\DriverManager;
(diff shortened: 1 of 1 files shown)


src/ForkCMS/Bundle/InstallerBundle/Service/InstallerConnectionFactory.php
```diff
@@ -33,7 +34,7 @@ public function createConnection(
 
             //continue with regular connection creation using new params
             return parent::createConnection($params, $config, $eventManager, $mappingTypes);
-        } catch (ConnectionException $e) {
+        } catch (ConnectionException | DBALException $e) {
```
WouterSioen: TIL: very nice!

StijnVrolijk: Yup, new since PHP 7.1 
https://wiki.php.net/rfc/multiple-catch

WouterSioen: Sadly I'm not working on PHP7 atm 😢

PR #2449, opened 2018-02-09
Make nesting possible for CollectionType

cirykpopeye:
## Type
- Non critical bugfix

Nesting CollectionTypes caused an issue so it would add the prototype also to it's child CollectionType.

## Pull request description
Fix the issue with nesting CollectionTypes.

### DIFF
--- a/src/Backend/Core/Js/jquery/jquery.backend.js
+++ b/src/Backend/Core/Js/jquery/jquery.backend.js
@@ -1500,7 +1500,7 @@
     e && e.preventDefault()
 
     var collection = $('#' + selector)
-    var list = collection.find('ul.js-collection')
+    var list = $(collection.find('ul.js-collection')[0])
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Js/jquery/jquery.backend.js
```diff
@@ -1500,7 +1500,7 @@
     e && e.preventDefault()
 
     var collection = $('#' + selector)
-    var list = collection.find('ul.js-collection')
+    var list = $(collection.find('ul.js-collection')[0])
```
StijnVrolijk: I think a nicer solution would be `collection.find('ul.js-collection').first()`, could you try that?

cirykpopeye: I agree, changed it as you suggested.

PR #2455, opened 2018-02-12
Drag and drop handler fix

jacob-v-dam:
## Type
- Non critical bugfix

## Pull request description
New drag and drop handlers fixed

### DIFF
--- a/src/Backend/Core/Js/backend.js
+++ b/src/Backend/Core/Js/backend.js
@@ -1883,7 +1883,7 @@ jsBackend.tableSequenceByDragAndDrop = {
           placeholder: 'dragAndDropPlaceholder',
           forcePlaceholderSize: true,
           stop: function (e, ui) {
-            jsBackend.tableSequenceByDragAndDrop.saveNewSequence($(this))
+            jsBackend.tableSequenceByDragAndDrop.saveNewSequence($(this).closest('table.jsDataGrid'))
           }
         }
       )
@@ -1892,6 +1892,8 @@ jsBackend.tableSequenceByDragAndDrop = {
         var $this = $(this)
         var $row = $this.closest('tr')
         var direction = $this.data('direction')
+
+        e.preventDefault()
 
         if (direction === 'up') {
           $row.prev().insertAfter($row)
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Js/backend.js
```diff
@@ -1883,7 +1883,7 @@ jsBackend.tableSequenceByDragAndDrop = {
           placeholder: 'dragAndDropPlaceholder',
           forcePlaceholderSize: true,
           stop: function (e, ui) {
-            jsBackend.tableSequenceByDragAndDrop.saveNewSequence($(this))
+            jsBackend.tableSequenceByDragAndDrop.saveNewSequence($(this).closest('table'))
```
carakas: must be table.jsDataGrid

src/Frontend/Core/Header/Header.php
```diff
@@ -119,7 +119,12 @@ public function __construct(KernelInterface $kernel)
         $this->meta = new MetaCollection();
 
         // add some default CSS files
-        $this->addCSS('/src/Frontend/Core/Layout/Css/screen.css');
+        $themeCss = $container->getParameter('site.path_www') . '/src/Frontend/Themes/' . Theme::getTheme() . '/Core/Layout/Css/screen.css';
```
carakas: why is this commit in here?
If I remember it correctly this has been fixed in fork in a different way so you should drop this commit

src/Backend/Modules/MediaLibrary/Domain/MediaGroup/MediaGroup.php
```diff
@@ -163,6 +163,7 @@ function (MediaGroupMediaItem $connectedItem): MediaItem {
     public function getFirstConnectedMediaItem(): ?MediaItem
     {
         $connectedMediaItems = $this->getConnectedMediaItems();
+        $firstMediaItem = null;
```
carakas: this commit shouldn't be in here

jacob-v-dam: In deed, i noticed this i'm removing this pull request and use the right commits :)

src/Backend/Core/Js/backend.js
jacob-v-dam: Fixed

PR #2461, opened 2018-02-13
Update node engine requirement to v8

ohvitorino:
## Type

- Non-critical bugfix

## Pull request description

This PR changes de minimum version of the node engine requirement from version 7 to version 8.


carakas:
sumocoders specific, wrong target repository

PR #2464, opened 2018-02-13
Fix buttons over text in pages module

carakas:
## Type

- Non critical bugfix

## Pull request description

before
screen shot 2018-02-13 at 15 51 56

after
screen shot 2018-02-13 at 15 51 28


src/Backend/Core/Layout/Sass/components/_pages.scss
```diff
@@ -179,15 +179,9 @@
...
-      background: $white;
-      bottom: 0;
-      padding: 0 4px 4px 0;
-      position: absolute;
-      right: 0;
-      -webkit-transition: opacity 0.3s ease-in-out;
-      -moz-transition: opacity 0.3s ease-in-out;
-      -o-transition: opacity 0.3s ease-in-out;
-      transition: opacity 0.3s ease-in-out;
+      padding: 0.7em 0 0;
```
Katrienvh: Why not use px? we use pixels everywhere. This is not consistence.

carakas: habit, @Katrienvh would you like me to change it?

StijnVrolijk: @carakas go ahead and change it

PR #2468, opened 2018-02-14
Use the symfony form in the template instead of plain html

carakas:
## Type

- Non critical bugfix

## Pull request description

Re-add the html of the symfony form that was lost from https://github.com/forkcms/forkcms/pull/1950


src/Frontend/Modules/Mailmotor/Layout/Widgets/Subscribe.html.twig
```diff
@@ -9,19 +9,20 @@
...
+        {% endif %}
+        {{ form_label(form.email) }}
         <div class="input-group">
-          <input type="email" value="" id="email" name="email" class="form-control" placeholder="{{ 'lbl.YourEmail'|trans|ucfirst }}" />
+          {{ form_widget(form.email) }}
           <span class="input-group-btn">
-            <input id="send" class="btn btn-default" type="submit" name="send" value="{{ 'lbl.Subscribe'|trans }}" />
+            {{ form_widget(form.subscribe) }}
           </span>
         </div>
```
jeroendesloovere: Add form_rest, so if somebody updates the SubscribeType Form everything keeps working.

carakas: @jeroendesloovere the form_rest will automatically be executed by the form_end

PR #2470, opened 2018-02-16
Media library: Add Bootstrap "Carousel" widget

jeroendesloovere:
## Type

- Feature

## Pull request description

Add Bootstrap "Carousel" widget.
schermafbeelding 2018-02-20 om 10 58 44


carakas:
The frontenders here at sumocoders aren't a big fan of the bootstrap slider and usually use a different one

Katrienvh:
Indeed i usually use http://kenwheeler.github.io/slick/

Katrienvh:
Indeed i usually use http://kenwheeler.github.io/slick/

jeroendesloovere:
Fork CMS already has a `Slider` that is using `slick`.

But for lots of use cases a `Bootstrap` Carousel is great by default. Since we already load in the bootstrap.js, we don't need to load in an additional slick library.

My opinion is that they can run against each other...

carakas:
I don't see the benefit of adding an additional slider

jeroendesloovere:
We love Carousel more then slick and mostly choose for Carousel.
Sometimes if we want to have more options, we choose for slick.
It also does not extra JS and CSS to be loaded in, because it is already in the theme by default.

So we are definitely in favor of integrating Carousel.
Maby ask the opinion of other developers, other then the ones in SumoCoders.

jeroendesloovere:
If I'm correct, Slick has no support for captions, which is something that we need on a regularly base.

mathiashelin:
I don't really care, but would't it be more scalable if you would dynamically check which templates are available? Otherwise you'll have to add a widget for each type of slider you would like to implement.

jeroendesloovere:
@mathiashelin Can you provide more information about what you mean/want?
The module is already dynamically, it looks at src/Frontend/Modules/MediaLibrary/Widgets and uses al the widgets it can find there.

mathiashelin:
Something like this: https://github.com/sumocoders/fork-slideshows/blob/master/src/Backend/Modules/Slideshows/Actions/Edit.php#L102
Not sure if this is possible in the MediaLibrary, I'm not a php expert.

jeroendesloovere:
@mathiashelin thanks for your reply.

That really sounds like a good idea.

Unfortunately we currently need to use the PHP widgets (and not the templates), because the PHP files executes extra code like $this->addJS(), $this->addJSData(), ... $this->addOpenGraphImages, ...
[MediaWidget::getPossibleValues()](https://github.com/forkcms/forkcms/blob/master/src/Backend/Modules/MediaGalleries/Domain/MediaGallery/MediaGalleryType.php#L46)

But, it is not a bad idea. I'm sure we can make such thing possible with same magic.

jeroendesloovere:
Ok, ready to merge then?!

carakas:
well, I would say no, you said that it would be possible to make something more generic, in that case I see no reason why we sould merge it now already

carakas:
the code also still needs to be reviewed on top of that

jeroendesloovere:
@carakas some day, when have refactored a lot of code and controllers can be used properly ...
Maby, then... we can use a more generic solution.
But in the meanwhile, it is a "no go".

I really want the carousel.

carakas:
I'm just not a fan of filling fork with a kazilion widgets for the media library

jeroendesloovere:
The combination of this PR and this one https://github.com/forkcms/forkcms/pull/2485
will be so powerful.

carakas:
@jeroendesloovere slick is far superior to bootstrap from what I've seen. @forkcms/frontenders  what do you think?

mathiashelin:
Don't remove Slick. I don't care if Bootstrap Carousel is added, won't be using it anyway.

jeroendesloovere:
Closing this one since the interest in merging this one is nihil.

jeroendesloovere:
Opening this one again since Bootstrap Carousel is so nice.

carakas:
@jeroendesloovere is there somebody beside you interested?

bramds:
i m only using slick cause it is faster and easier to use for slideshows and lightboxes

freshface:
We use cycle2 :)

PR #2477, opened 2018-02-19
Media Library/Galleries: CamelCase ID's

jeroendesloovere:
## Type

- Enhancement

## Resolves the following issues

Fixes #2441

PR #2478, opened 2018-02-19
Media library: CamelCase html DOM ID's

### LINKED ISSUE #2411: CKEditor: Extra image options when inserting MediaItem from MediaLibrary in editor
## Type

- Feature

## Problem description

Since we have dropped ckfinder, we now include images in the editor through the media libray.  Now we can set the path and the alt text of the image.  In some cases it could be usefull that you re able to also set:
- image height and width ( recalculated like in ckfinder )
- class for the image

This way we can control the image more in the frontend. ( setting width and height with the css class alone wouldnt work, because the dimension would be hardcoded in css and may distort some images )

### LINKED ISSUE #2441: Id's in MediaLibrary not camelcased
## Type
- Enhancement

## Problem description

https://github.com/forkcms/forkcms/blob/master/src/Backend/Modules/MediaLibrary/Layout/Templates/MediaItemIndex.html.twig
In this file the id's are not consistent, sometimes they are camel cased and sometimes - separated. I think the default in fork is camel casing for id.

jeroendesloovere:
## Type

- Enhancement

Fixes #2441  (replaces https://github.com/forkcms/forkcms/pull/2477)


src/Backend/Modules/MediaLibrary/Resources/views/BackendMediaLibraryUpload.html.twig
```diff
@@ -141,7 +141,7 @@
           </div>
         </div>
       </div>
-      <div id="fine-uploader-gallery"></div>
+      <div id="fineUploader"></div>
```
carakas: why do you have the same id twice?

Katrienvh (approved):

carakas:
Refactor in fork 7 because too much has changed

PR #2479, opened 2018-02-19
The tags should be case insensitive unique

carakas:
## Type

- Non critical bugfix

## Pull request description

array_unique takes the case into account, adding a tag Ichtus and ichtus would result in a 500 error since the database would see it as a double entry but the php as separate entries

### DIFF
--- a/src/Backend/Modules/Tags/Engine/Model.php
+++ b/src/Backend/Modules/Tags/Engine/Model.php
@@ -235,8 +235,8 @@ class Model
             $tags = (array) explode(',', (string) $tags);
         }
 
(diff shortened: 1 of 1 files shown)


WouterSioen:
Was it previously possible to add the same tag with both cases or did you get a 500 error?

If it was possible, this is a BC break and should probably not be pushed in a minor release?

StijnVrolijk:
@WouterSioen I've just tested it on the demo

image

When saving this it will give a 500 error

WouterSioen:
Bugfix it is, let's merge this and put it in the next release?

PR #2481, opened 2018-02-20
Fix overview pages in blog module showing 404's when there are no blog posts

carakas:
## Type

- Non critical bugfix

## Pull request description

If there were no blog post the numer of pages was 0 and that resulted in a false 404

### DIFF
--- a/src/Frontend/Modules/Blog/Actions/Index.php
+++ b/src/Frontend/Modules/Blog/Actions/Index.php
@@ -29,6 +29,10 @@ class Index extends FrontendBaseBlock
         $limit = $this->get('fork.settings')->get($this->getModule(), 'overview_num_items', 10);
         $numberOfPages = (int) ceil($numberOfItems / $limit);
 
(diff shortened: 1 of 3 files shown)


src/Frontend/Modules/Blog/Actions/Index.php
```diff
@@ -29,6 +29,10 @@ private function buildPaginationConfig(): array
         $limit = $this->get('fork.settings')->get($this->getModule(), 'overview_num_items', 10);
         $numberOfPages = (int) ceil($numberOfItems / $limit);
 
+        if ($numberOfPages === 0) {
+            $numberOfPages = 1;
+        }
```
WouterSioen: Changing the exact same thing in two locations might indicate that we're missing a concept in our data model here. A pagination object/trait might be a nice cleanup?

carakas: This is just a small method to build the config. the biggest part of the pagination is already extracted

PR #2483, opened 2018-02-20
Add referrer security header

carakas:
## Type

- Enhancement

## Pull request description

More info see https://scotthelme.co.uk/a-new-security-header-referrer-policy/#strictoriginwhencrossorigin


jeroendesloovere:
Seems like the best and most secure solution.

jessedobbelaere:
Nice, discovered this header as well yesterday when doing a https://securityheaders.io scan 👍

PR #2485, opened 2018-02-21
Add MediaItemTranslation and make it editable

### LINKED ISSUE #2469: Media Library: Add option to add description to every MediaGroupMediaItem
## Type

- Feature

## Problem description

When creating a media gallery, you can add images.
Sometimes you feel the need to have the option to add a description to every image.

Example:
Title + Description + Link to the specific page.
schermafbeelding 2018-02-16 om 11 02 28

We should add a button to every image, which opens a pop-up, where we can change the "text".
schermafbeelding 2018-02-16 om 11 03 54

jeroendesloovere:
## Type

- Feature

Fixes #2469 
This PR handles the Backend part.

## Pull request description

MediaItem has now more descriptive fields and they can be filled in for every language:
* Title
* Caption
* CaptionLink
* Alt text
* Description
schermafbeelding 2018-02-21 om 13 22 03

schermafbeelding 2018-02-21 om 13 43 48


jeroendesloovere:
@forkcms/core-contributors 

I would like some advice about what you guys find the best way of integrating getTranslation() in back- and frontend. So we can access mediaItem.getTranslation().getTitle()/getCaption()/getDescription()/getAlt()/...

src/Backend/Modules/MediaLibrary/Actions/MediaItemEdit.php
```diff
@@ -75,12 +77,12 @@ public function execute(): void
             $this->getBackLink(
                 [
                     'report' => 'media-item-edited',
-                    'var' => $updateMediaItem->title,
+                    'var' => $updateMediaItem->url,
```
carakas: why did you change this?

src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
```diff
@@ -235,21 +258,28 @@ public static function createFromMovieUrl(
     public static function fromDataTransferObject(MediaItemDataTransferObject $mediaItemDataTransferObject): ?MediaItem
     {
         if (!$mediaItemDataTransferObject->hasExistingMediaItem()) {
-            throw new \BadFunctionCallException('This method can not be used to create a new media item');
+            throw new BadFunctionCallException('This method can not be used to create a new media item');
```
carakas: Exception thrown if a callback refers to an undefined function or if some arguments are missing. 

I think a RuntimeException would be better here

src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
```diff
@@ -235,21 +258,28 @@ public static function createFromMovieUrl(
...
+
+            $mediaItem->addTranslation(MediaItemTranslation::fromDataTransferObject($mediaItemTranslationDataTransferObject));
+        }
+
         return $mediaItem;
     }
 
     public function jsonSerialize(): array
     {
+        $locale = BackendLocale::workingLocale();
```
carakas: not a fan of this, I think a method to set the active language is better with a fallback like this for instance, but now if you would try to use this in the frontend it would fail

carakas: upi can then also have a method, getActiveTranslation

src/Backend/Modules/MediaLibrary/Actions/MediaItemEdit.php
jeroendesloovere: Before
MediaItem->getTitle();

After
MediaItem->getTranslation(locale)->getTitle();

That is why I'm now using $updateMediaItem->url, because ->title doesn't exist anymore

src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
jeroendesloovere: But,... some arguments are missing? That is why I throw error.

jeroendesloovere: Fixed. Changed to RuntimeException as requested.

src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
jeroendesloovere: Can you tell me how do you like to see Locale (Backend/Frontend) integrated?

jeroendesloovere: @carakas, any example?

jeroendesloovere: poke @carakas

carakas: just add a property for the current locale that isn't persisted to the entity, then add a method to set/change that locale and use that to get the correct translation

don't know the location of a code snippet by heart

PR #2486, opened 2018-02-22
Add the tags to the blogpost in the frontend in the model instead of the action

carakas:
## Type

- Enhancement

## Pull request description

That way it will also be available in the widgets

### DIFF
--- a/src/Frontend/Modules/Blog/Engine/Model.php
+++ b/src/Frontend/Modules/Blog/Engine/Model.php
@@ -762,6 +762,14 @@ class Model implements FrontendTagsInterface
             }
         }
 
+        if (isset($blogPost['id'])) {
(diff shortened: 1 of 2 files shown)


src/Frontend/Modules/Blog/Engine/Model.php
```diff
@@ -762,6 +763,14 @@ private static function completeBlogPost(array $blogPost)
             }
         }
 
+        if (isset($blogPost['id'])) {
+            $blogPost['tags'] = FrontendTagsModel::getForItem(
+                'Blog',
```
WouterSioen: Is there no constant for this? Something like `Module::BLOG` or `self::MODULE_NAME` might be more maintainable?

Issue #708, opened 2014-04-16
Replace SpoonThumbnail/SpoonFile by Imagine


jeroendesloovere:
When removing SpoonThumbnail, SpoonFile, ... we could maby replace it by Imagine.

[Packagist](https://packagist.org/packages/imagine/imagine)
[Documentation](http://imagine.readthedocs.org/en/latest/)

[Github](https://github.com/avalanche123/Imagine)
[Slidershare](http://www.slideshare.net/avalanche123/introduction-toimagine)

PR #2492, opened 2018-02-27
Refactor out SpoonThumbnail

jeroendesloovere:
## Type

- Feature
- Deprecation

Fixes https://github.com/forkcms/forkcms/issues/708

## Pull request description

Refactor out SpoonThumbnail.
We replaced the functionality with Imagine.


src/ForkCMS/Service/Thumbnails.php
```diff
@@ -0,0 +1,161 @@
...
+                $item['width'] = ($chunks[0] !== '') ? (int)$chunks[0] : null;
+                $item['height'] = ($chunks[1] !== '') ? (int)$chunks[1] : null;
+            }
+
+            $return[] = $item;
+        }
+
+        return $return;
+    }
+}
```
carakas: new line missing at the end of the file

StijnVrolijk:
I don't like the `Service` namespace. In an ideal (utopian) world, nearly everything should be a service. I'd give this a `Utility` namespace.

jeroendesloovere:
@carakas What do you think of Utils?

jeroendesloovere:
I'm using `Service` because that is the default in Symfony.
Check out this example https://symfony.com/doc/current/service_container.html#creating-configuring-services-in-the-container

carakas:
@jeroendesloovere That is not a default that is an example. The only rule is that the source code needs to be inside src, for the rest you are free (http://fabien.potencier.org/symfony4-directory-structure.html)

app/config/config.yml
```diff
@@ -198,6 +198,9 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse }
 
+    forkcms.thumbnails:
+        class: ForkCMS\Service\Thumbnails
```
carakas: you must declare it public

carakas:
Apart from that is the current recommendation to use the FQCN as service name.
You can always add an alias for `forkcms.utility.thumbnails`

But the benifit is that you can just do $this->getContainer()->get(Thumbnails::class)

src/ForkCMS/Service/Thumbnails.php
jeroendesloovere: Fixed

jeroendesloovere:
I prefer `forkcms.thumbnails`. @StijnVrolijk what do you think?
We do also use `fork.settings`?

app/config/config.yml
jeroendesloovere: Fixed

StijnVrolijk:
@jeroendesloovere settings arent really utilities, therefor it makes sense to me to have `fork.settings` but also `forkcms.utility.thumbnails`

jeroendesloovere:
Or should I use the fully qualified name and don't create an alias?

carakas:
I would do both, use the FQCN as service name and forkcms.utility.thumbnails as alias in case somebody wants to use it like that?

jeroendesloovere:
@carakas Can you re-review?

app/config/config.yml
```diff
@@ -198,7 +198,7 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse }
 
-    forkcms.thumbnails:
+    forkcms.utility.thumbnails:
```
carakas:
```yml
ForkCMS\Utility\Thumbnails:
    alias: forkcms.utility.thumbnails
    public: true
```

Can you try this instead

bin/console
```diff
@@ -42,7 +42,7 @@ $backend->initialize('Console');
 
 $application = new Application($kernel);
 $application->add(new \Console\Locale\ImportLocaleCommand());
-$application->add(new \Console\Thumbnails\GenerateThumbnailsCommand());
+$application->add(new \Console\Thumbnails\GenerateThumbnailsCommand($kernel->getContainer()->get('forkcms.utility.thumbnails')));
```
carakas: maybe you can register it in the yml config instead of hardcoded here?

src/Common/Core/Model.php
```diff
@@ -153,49 +151,26 @@ public static function generatePassword(
...
-            $thumbnail = new \SpoonThumbnail($sourceFile, $folder['width'], $folder['height']);
-            $thumbnail->setAllowEnlargement(true);
-
-            // if the width & height are specified we should ignore the aspect ratio
-            if ($folder['width'] !== null && $folder['height'] !== null) {
-                $thumbnail->setForceOriginalAspectRatio(false);
-            }
-            $thumbnail->parseToFile($folder['path'] . '/' . $filename);
-        }
+        self::get('forkcms.utility.thumbnails')->generate($path, $sourceFile);
```
carakas: when using the class name can also do Classname::class if you want

jeroendesloovere:
Why don't we use `ForkCMS\Utils` instead of `ForkCMS\Utility`?
Symfony always uses `Utils` in their examples, it would be unwise to use another naming then...
https://symfony.com/doc/current/best_practices/business-logic.html

StijnVrolijk:
Utils is an abbreviation of Utilities and abbreviations in code cause confusion (which is the reason we renamed `frm` to `form` for example).
Furthermore we never use plurals in our new namespaces (eg `Command`, `Event`, ..) so I think we should use the singular form: Utility.

jeroendesloovere:
Ok, I searched the web a little and read your opinion and I'm convinced.

Second question.
`ForkCMS\Utility\Thumbnails` or ForkCMS\Utility\GenerateThumbnails`?

bin/console
jeroendesloovere: Done

src/Common/Core/Model.php
jeroendesloovere: Done

app/config/config.yml
jeroendesloovere: Done

jeroendesloovere:
@StijnVrolijk and @carakas can you guys approve?

carakas:
I can approve if the code is good :) I'll see if I still have time today to go through it

src/Common/Core/Model.php
```diff
@@ -153,49 +152,26 @@ public static function generatePassword(
...
+        self::get(Thumbnails::class)->generate($path, $sourceFile);
     }
 
     /**
      * Delete thumbnails based on the folders in the path
      *
      * @param string $path The path wherein the thumbnail-folders exist.
      * @param string|null $thumbnail The filename to be deleted.
+     *
+     * @deprecated Please use the service `forkcms.utility.thumbnails` instead.
```
carakas: you removed the alias from the service definition. You need to update it here or add that alias again

src/Console/Thumbnails/GenerateThumbnailsCommand.php
```diff
@@ -28,41 +35,37 @@ protected function configure(): void
             );
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output): void
+    public function __construct(Thumbnails $thumbnails)
     {
-        // Get input values
-        $folderOption = $input->getOption('folder');
+        $this->finder = new Finder();
```
carakas: I wouldn't move this. It is only used in one method. I'm not sure if the constructor is always called or if there is an optimisation on it. But if it would always be called also when not using this command creating a new instance of finder is a waste of resources

src/ForkCMS/Utility/Thumbnails.php
```diff
@@ -0,0 +1,160 @@
...
+     *  - x128 as foldername to generate an image where the height will be
+     *      128px, the width will be calculated based on the aspect ratio.
+     *
+     * @param string $inPath The path wherein the thumbnail-folders will be stored.
+     * @param string $forSourceFile The location of the source file.
+     */
+    public function generate(string $inPath, string $forSourceFile): void
+    {
+        // get folder listing
+        $folders = $this->getFolders($inPath);
```
carakas: why don't you loop over the array that the method returns directly?

src/ForkCMS/Utility/Thumbnails.php
```diff
@@ -0,0 +1,160 @@
...
+
+        foreach ($this->finder->directories()->in($inPath)->name($nameFilter)->depth('== 0') as $directory) {
+            $dirname = $directory->getBasename();
+            $chunks = explode('x', $dirname, 2);
+
+            if (!$includeSourceFolder && count($chunks) !== 2) {
+                continue;
+            }
+
+            $width = (($dirname !== 'source') && ($chunks[0] !== '')) ? (int) $chunks[0] : null;
```
carakas: woulnd't it be better to just check here if the chunk is numerical?

src/ForkCMS/Utility/Thumbnails.php
```diff
@@ -0,0 +1,160 @@
...
+            $height = (($dirname !== 'source') && ($chunks[1] !== '')) ? (int) $chunks[1] : null;
```
carakas: same here

composer.json
```diff
@@ -86,5 +86,10 @@
             "**/tests/",
             "**/Test/"
         ]
+    },
+    "autoload-dev": {
```
carakas: why did you add this one?
Is it a new namespace?

src/Common/Core/Model.php
jeroendesloovere: Fixed by 8fd859c

src/Console/Thumbnails/GenerateThumbnailsCommand.php
jeroendesloovere: Fixed by c40f720

src/ForkCMS/Utility/Thumbnails.php
jeroendesloovere: Fixed

src/ForkCMS/Utility/Thumbnails.php
jeroendesloovere: Indeed, changed to `is_int(xxx)`.
Fixed by d1281a4

src/ForkCMS/Utility/Thumbnails.php
jeroendesloovere: Indeed, changed to is_int(xxx).
Fixed by d1281a4

composer.json
jeroendesloovere: Tests in Symfony 4 are in `/tests/`.
So I already provided this in the composer.json and placed the tests in the right position.
The namespace `ForkCMS\Tests` was not already existing, so that is probably OK then.

jeroendesloovere:
@carakas Fixed everything in your review.

jeroendesloovere:
@carakas can you re-review this if you find some time :-P

carakas:
If time could be found
Picked up from the ground
The richest on earth
Eternally crowned

PR #2501, opened 2018-03-05
Set aria label on subnavigation

Katrienvh:
## Type
- Enhancement

## Pull request description
Set aria label on subnavigation based on the active page

### DIFF
--- a/src/Backend/Core/Engine/Navigation.php
+++ b/src/Backend/Core/Engine/Navigation.php
@@ -27,7 +27,10 @@ final class Navigation extends KernelLoader
     public function parse(TwigTemplate $template): void
     {
         $template->assign('navigation', $this->navigation);
-        $template->assign('activeParent', $this->navigation[$this->getSelectedKey($this->navigation)]['label']);
+        $selectedKey = $this->getSelectedKey($this->navigation);
+        if ($selectedKey !== null && isset($this->navigation[$selectedKey])) {
+            $template->assign('activeParent', $this->navigation[$selectedKey]['label']);
+        }
     }
 
     private function addActiveStateToNavigation(array $navigation): array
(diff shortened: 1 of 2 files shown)


src/Backend/Core/Layout/Templates/base.html.twig
```diff
@@ -8,11 +8,13 @@
 
 {% include "Layout/Templates/header.html.twig" %}
 
+
```
bjorvack: newline isn't needed

src/Backend/Core/Engine/Navigation.php
```diff
@@ -27,6 +27,7 @@ public function __construct(KernelInterface $kernel)
     public function parse(TwigTemplate $template): void
     {
         $template->assign('navigation', $this->navigation);
+        $template->assign('activeParent', $this->navigation[$this->getSelectedKey($this->navigation)]['label']);
```
carakas: @Katrienvh dit is niet altijd beschikbaar, op sommige pagina's geeft dit dus een error kan je dit oplossen?

StijnVrolijk: @carakas can you give an example of a page without a selected key?

carakas: the login page in the backend for instance, those pages are the reason the tests are failing

StijnVrolijk: @Katrienvh 
```
$selectedKey = $this->getSelectedKey($this->navigation);
if ($selectedKey !== null && isset($this->navigation[$selectedKey])) {
  $template->assign('activeParent', $this->navigation[$selectedKey]['label']);
}
```
And of course add an `{% if %}{% else %}{% endif %}` in your template

PR #2506, opened 2018-03-06
Wcag update tags input help text

Katrienvh:
## Type
- Enhancement

## Pull request description
Screenreaders can't handle the enter symbol.
Added aria describedby so the help block is linked to the input field


src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -539,7 +539,7 @@ private function loadForm(): void
                 null,
                 'form-control js-tags-input',
                 'error js-tags-input'
-            );
+            )->setAttribute('aria-describedby', 'tags-info');;
```
StijnVrolijk: you have 2 `;`'s at the end of this line

PR #2507, opened 2018-03-06
Delete entries linked to a deleted group

ohvitorino:
## Type
- Non critical bugfix

## Pull request description

When a group is deleted, the linked settings and rights are not being deleted. This leaves some garbage in the database, which is not a good thing.

### DIFF
--- a/src/Backend/Modules/Groups/Engine/Model.php
+++ b/src/Backend/Modules/Groups/Engine/Model.php
@@ -53,6 +53,9 @@ class Model
     public static function delete(int $groupId): void
     {
         BackendModel::getContainer()->get('database')->delete('groups', 'id = ?', [$groupId]);
+        BackendModel::getContainer()->get('database')->delete('groups_settings', 'group_id = ?', [$groupId]);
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Groups/Engine/Model.php
```diff
@@ -53,6 +53,9 @@ public static function alreadyExists(string $groupName): bool
     public static function delete(int $groupId): void
     {
         BackendModel::getContainer()->get('database')->delete('groups', 'id = ?', [$groupId]);
+        BackendModel::getContainer()->get('database')->delete('groups_settings', 'group_id = ?', [$groupId]);
```
StijnVrolijk: Please add `BackendModel::getContainer()->get('database')` to a variable so you don't have to go through all those getters every time.

```
/** @var SpoonDatabase $database */
$database = BackendModel::getContainer()->get('database');
```

ohvitorino:
@StijnVrolijk I made the changes you requested 👍

PR #2508, opened 2018-03-06
Fix check if user belongs to a certain group

ohvitorino:
The group id coming from the database is a string and it must cast as an integer for the comparison.

## Type

- Non critical bugfix

## Pull request description

The comparison between the group id coming from the database and the one provided in the method `Model::isUserInGroup` of the Groups module should be done using integers.
The value coming from the database is now a string and the value passed into the method is an integer.


ohvitorino:
@StijnVrolijk I fixed the typo 😄

PR #2515, opened 2018-03-12
Dont load images that don't exist

### LINKED ISSUE #2490: Error 404 if no image loaded for home page, only background
## Type

- Non critical bug

## Problem description

Error 404 if no image loaded for home page
missing file - /src/Frontend/Files/Pages/images/source/1518760782.jpg

## Steps to reproduce

Use "Dev tools" Ctrl+Shift+I in browser to see the 404 error.

carakas:
## Type

- Non critical bugfix

fixes #2490

## Pull request description

We tried to load an image even if there was no image to be loaded resulting in a 404 on that call

### DIFF
--- a/src/Frontend/Themes/Fork/Core/Layout/Templates/Base.html.twig
+++ b/src/Frontend/Themes/Fork/Core/Layout/Templates/Base.html.twig
@@ -6,7 +6,7 @@
     {% include "Core/Layout/Templates/Cookies.html.twig" %}
   {% endblock %}
 
-  <div class="header header-blue" style="background-image: url('{{ FRONTEND_FILES_URL }}/Pages/images/source/{{ page.data.image }}');">
+  <div class="header header-blue" {% if page.data.image is defined and page.data.image is not empty %}style="background-image: url('{{ FRONTEND_FILES_URL }}/Pages/images/source/{{ page.data.image }}');"{% endif %}>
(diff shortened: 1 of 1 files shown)


src/Frontend/Themes/Fork/Core/Layout/Templates/Base.html.twig
```diff
@@ -6,7 +6,7 @@
     {% include "Core/Layout/Templates/Cookies.html.twig" %}
   {% endblock %}
 
-  <div class="header header-blue" style="background-image: url('{{ FRONTEND_FILES_URL }}/Pages/images/source/{{ page.data.image }}');">
+  <div class="header header-blue" {% if page.data.image is defined and page.data.image is not empty %}style="background-image: url('{{ FRONTEND_FILES_URL }}/Pages/images/source/{{ page.data.image }}');"{% endif %}>
```
Katrienvh: I think this can easier: {% if page.data.image %} ?

carakas: that will result in an error if page.data.image doesn't exist if I remember, or is that only when page for instance wouldn't exist?

Katrienvh: i think, but i am not sure

PR #2522, opened 2018-04-03
Only copy the location widgets to an other locale if the Location module is installed

StijnVrolijk:
## Type

- Non critical bugfix

## Pull request description
If it's not installed it would've thrown an error when copying pages to an other locale


jeroendesloovere:
I do not know if we need to merge this one,
since the PR https://github.com/forkcms/forkcms/pull/2422 already covers this...

StijnVrolijk:
@jeroendesloovere I made this PR since the PR you referenced is for Fork 6 and it might be a while before that gets merged.

PR #2525, opened 2018-04-13
Integrate Geolocation in Backend Location Model

jeroendesloovere:
## Type

- Enhancement

## Pull request description

Use an external well tested library (https://packagist.org/packages/jeroendesloovere/geolocation-php-api) for getting the location coordinates.

Added ForkCMS\Utility\Geolocation class

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,8 @@
         "spoon/library": "^3.0",
         "league/flysystem-aws-s3-v3": "^1.0.13",
         "league/flysystem-cached-adapter": "^1.0.6",
-        "pimple/pimple": "^3.2"
+        "pimple/pimple": "^3.2",
+        "jeroendesloovere/geolocation-php-api": "^2.1"
     },
     "require-dev": {
         "jdorn/sql-formatter": "1.2.17",
--- a/src/ForkCMS/Utility/Geolocation.php
+++ b/src/ForkCMS/Utility/Geolocation.php
@@ -26,7 +26,6 @@ class Geolocation
      * @param string|null $zip
      * @param string|null $country
      * @return array - Example: ['latitude' => 50.8864, 'longitude' => 3.42928]
-     * @throws Exception
      */
     public function getCoordinates(
         string $street = null,
--- a/src/ForkCMS/Utility/Geolocation.php
+++ b/src/ForkCMS/Utility/Geolocation.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace ForkCMS\Utility;
+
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -40,7 +40,8 @@
         "spoon/library": "^3.0",
         "league/flysystem-aws-s3-v3": "^1.0.13",
         "league/flysystem-cached-adapter": "^1.0.6",
-        "pimple/pimple": "^3.2"
+        "pimple/pimple": "^3.2",
+        "jeroendesloovere/geolocation-php-api": "^2.1"
```
carakas: why did you create your own package instead of using an existing/proven one?

jeroendesloovere: Can you point me in the direction of such a library?

carakas: https://packagist.org/

carakas:
@forkcms/core-contributors what do you think about this pr

src/ForkCMS/Utility/Geolocation.php
```diff
@@ -0,0 +1,56 @@
...
+        /** @var Coordinates $coordinates */
+        $coordinates = $this->api->getCoordinates(
+            $street,
+            $streetNumber,
+            $city,
+            $zip,
+            $country
+        );
+
+        return [
```
tijsverkoyen: What if the address can't be decoded?

jeroendesloovere: If any exception occurs in the vendor this will be thrown.
We just need to decide how we will handle this in Fork... 

Possible solution:
What about a try/catch around `$this->api->getCoordinates(` and return null values 
```
        return [
            'latitude' => null,
            'longitude' => null,
        ];
```
Any other suggestions?

jeroendesloovere: ... using null is exactly the same what we use now in https://github.com/forkcms/forkcms/pull/2525/files#diff-3d41d594577d7d3f3003016b06d01cd9L144

I'm going to update this PR so it matches the same behavior.

jeroendesloovere: Done. @tijsverkoyen can you review https://github.com/forkcms/forkcms/pull/2525/commits/60a0d41b90d135df5f2ec07f7b0cf93510c83bae

jeroendesloovere:
@carakas. Ready to merge?

carakas:
I would still wait a bit, @tijsverkoyen didn't reply yet to the thing you did after his review

tijsverkoyen:
ok for me

PR #2527, opened 2018-04-27
Fix google reCAPTCHA javascript link in production

jeroendesloovere:
## Type

- Critical bugfix

Using `$this->header->addJS(`, we add a link.
The second parameter is by default true and that means it will get minified in production.
That is wrong (because it is a link) and causes an error and stopping google reCAPTCHA from working.

## Pull request description

The JavaScript file is a link, so it has no content and therefore does not need to be minified.

### DIFF
--- a/src/Frontend/Modules/FormBuilder/Widgets/Form.php
+++ b/src/Frontend/Modules/FormBuilder/Widgets/Form.php
@@ -331,7 +331,7 @@ class Form extends FrontendBaseWidget
         $this->template->assign('successMessage', false);
 
         if ($this->hasRecaptchaField) {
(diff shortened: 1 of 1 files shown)


StijnVrolijk:
@jeroendesloovere if what you say is correct, the solution is incorrect. In the `addJS` method there's a check for external urls. This url is external so it shouldn't get minified, even if `$minify` is set to `true`.

See `\Frontend\Core\Header\minify:173`

jeroendesloovere:
Thanks @StijnVrolijk for the reply.

I integrated reCAPTCHA in a Fork CMS 4.5,
that is why the bug occured.

Closing this one.

carakas:
this was indeed a bug in the addJs method that has been fixed already

PR #2528, opened 2018-04-30
Set the language for the Frontend

ChristopheSR:
## Type

- Non critical bugfix

## Pull request description

Set the language so it is available to be assigned in Common/Core/Twig/BaseTwigTemplate.php to the frontend variables. For the Backend this was already set. Because this was not set the isEN or other language checks would not work anymore in the frontend templates.


carakas:
@ChristopheSR can you fix the issue with the test? Maybe try to see if you can do a clean install, I think the issue might be situated there

PR #2530, opened 2018-05-03
Updated Twigtemplate so language gets set

ChristopheSR:
## Type

- Non critical bugfix

## Pull request description

Set the language so it is available to be assigned in Common/Core/Twig/BaseTwigTemplate.php to the frontend variables. For the Backend this was already set. Because this was not set the isEN or other language checks would not work anymore in the frontend templates.


carakas:
@ChristopheSR I think it would be best to take a look at this on your computer and go through a clean install. I think the code fails during the install but I'm not sure. You can also run the tests locally and get a more in-depth look like that

PR #2543, opened 2018-05-28
Fix docker image not building because libpng12-dev was not found

carakas:
Replaced it with libpng-dev instead

## Type

- Non critical bugfix

## Pull request description

Creating a new docker image didn't work anymore, this fixes it


jessedobbelaere:
Looks good, 👍 I also had to do this recently in a non-forkcms project. Apparently that libpng12-dev package is dropped after Ubuntu 16.04

PR #2548, opened 2018-05-29
increase log level for the dockerstdout since it also impacts the console commands

carakas:
## Type

- Enhancement

## Pull request description

We got a lot of stuff that we didn't need mixed in with the console commands because of the log level for docker


StijnVrolijk:
Ahh, no more notice spam 😌

PR #2555, opened 2018-06-01
We've already fetched the extras, no need to fetch them again

StijnVrolijk:
## Type
- Enhancement

## Pull request description
We fetched the extras twice in the page edit action. Now we only do it once, +1 performance points!

### DIFF
--- a/src/Backend/Modules/Pages/Actions/Edit.php
+++ b/src/Backend/Modules/Pages/Actions/Edit.php
@@ -635,7 +635,7 @@ class Edit extends BackendBaseActionEdit
         $this->template->assign('templates', $this->templates);
         $this->template->assign('positions', $this->positions);
         $this->template->assign('extrasData', json_encode(BackendExtensionsModel::getExtrasData()));
-        $this->template->assign('extrasById', json_encode(BackendExtensionsModel::getExtras()));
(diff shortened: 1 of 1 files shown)


jeroendesloovere:
Nice catch

src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -635,7 +635,7 @@ protected function parse(): void
         $this->template->assign('templates', $this->templates);
         $this->template->assign('positions', $this->positions);
         $this->template->assign('extrasData', json_encode(BackendExtensionsModel::getExtrasData()));
-        $this->template->assign('extrasById', json_encode(BackendExtensionsModel::getExtras()));
+        $this->template->assign('extrasById', json_encode($this->extras));
```
WouterSioen: The question is, can these extras be changed in between fetching them and assigning them here? 🤔

StijnVrolijk: well.. they can but they aren't. They're used to check if a page contains more than 1 block or if a block is placed on the homepage. Then they get assigned.

WouterSioen: Let's 🚢 then!

PR #2556, opened 2018-06-01
Remove duplicate id="fine-uploader-gallery" element

jeroendesloovere:
## Type

- Enhancement

`<div id="fine-uploader-gallery"></div>` was twice in this template, so I removed one.


carakas:
dit you test all possible ways this is used to make sure that nothing breaks?

jeroendesloovere:
I have already implemented this fix in 3 production sites.

PR #2557, opened 2018-06-05
Fix the ck editor custom icons

carakas:
## Type

- Non Critical bugfix

## Pull request description

It will fix the image and link icons


Katrienvh:
Are there no icons of ckeditor itself like the other icons? There is a difference now and the new icons are not aligned correct

carakas:
there are icons, but then we need to manually use the correct offset in the sprite, and if the sprite changes these problems return. I tried extending those classes but that didn't work

carakas:
wouldn't it be possible to fix the alignment or something?

Katrienvh:
Yeah you can add padding and maybe make them larger and the give them the color of the other icons

jeroendesloovere:
Is it possible to change all ckeditor icons to one of fontawesome?

carakas:
@jeroendesloovere no that is not possible
@Katrienvh  I've updated it
screen shot 2018-06-26 at 08 36 37

PR #2558, opened 2018-06-05
Fix form token error

carakas:
## Type

- Non critical bugfix

## Pull request description

Ok, I have no clue why, this validate function seems to be in the wrong place, since it validates the form before it is submitted.

I have no clue why we are suddenly seeing this invalid token error when we open some pages but this seems to fix it

@tijsverkoyen can you shine a light on this?

### DIFF
--- a/src/Backend/Core/Engine/Form.php
+++ b/src/Backend/Core/Engine/Form.php
@@ -173,7 +173,6 @@ class Form extends \Common\Core\Form
     public function parse($tpl): void
     {
         parent::parse($tpl);
-        $this->validate();
 
         // if the form is submitted but there was an error, assign a general error
         if ($this->useGlobalError && $this->isSubmitted() && !$this->isCorrect()) {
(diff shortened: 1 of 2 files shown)


jeroendesloovere:
So we can close this PR and open a new PR on forkcms/library

carakas:
that will still keep the issue of the validation of the form running when the form isn't submitted

jeroendesloovere:
@carakas 

The `validate()` comment states => `Validates the form. This is an alternative for isCorrect, but without retrieve the status of course.`.

1. Check if there is a token set
2. Checks if the fields added to the form are valid (= do they have the getErrors method)
```
// loop objects
foreach($this->objects as $oElement)
{
    // check, since some objects don't have this method!
    if(is_callable(array($oElement, 'getErrors'))) $errors .= $oElement->getErrors();
}
```

So I guess `$this->validate()` should not be deleted and this PR should be closed.

jeroendesloovere:
A better name ;-P for the method should actually be something like... `checkIfFormContainsAllAppropiateFields()`

PR #2571, opened 2018-06-27
Fix array instead of file name passed on to the function

carakas:
## Type

- Critical bugfix

## Pull request description

The upload module action was broken because of typehints and faulty data being passed on to that method


carakas:
indeed, maybe it did, but I caught it red-handed

PR #2572, opened 2018-06-27
parent::execute should always be called

carakas:
Cleaned up the code so that the parent::execute function works

## Type

- Critical bugfix

## Pull request description

Fixes js issues because the parent wasn't called


carakas:
@jeroendesloovere thanks

PR #2577, opened 2018-06-29
A button should be a button, not an anchor

StijnVrolijk:
## Type
- Critical bugfix

## Pull request description
A button should be a button, not an anchor
Buttons don't have urls, silly you!


carakas:
@jeroendesloovere why? is there a reason for it?

jeroendesloovere:
@carakas I have no idea...
I just know that there are "zero search results" for `macro.buttonIcon(null,`.
So to keep everything uniform I suggest we also use `macro.buttonIcon('',` or change everything...

jeroendesloovere:
```
  {# When url is an empty string, use a button, otherwise use a hyperlink/anchor #}
  {% if url %}
```

StijnVrolijk:
Well, the code doesn't do what the comment says it does so something needs to change

jeroendesloovere:
@StijnVrolijk Can you update the comment and change `null` to `''`. Thanks.

carakas:
I actually think that null makes more sense

StijnVrolijk:
@jeroendesloovere updating the comment still allows null :)

PR #2580, opened 2018-07-02
Wrong btn class

freshface:
The btn was green, but should have a default class

## Type

- Enhancement

### DIFF
--- a/src/Backend/Core/Js/jquery/jquery.backend.js
+++ b/src/Backend/Core/Js/jquery/jquery.backend.js
@@ -1246,7 +1246,7 @@
         '</div>' +
         '<div class="input-group">' +
         '<input class="form-control dontSubmit" id="addValue-' + id + '" name="addValue-' + id + '" type="text" />' +
-        '<span class="input-group-btn"><button id="addButton-' + id + '" class="btn btn-success">' +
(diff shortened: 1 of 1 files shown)


carakas:
can you give a bit more context or a screenshot?

freshface:
This btn is green. https://jmp.sh/Clmz7I8
Kinda weird for a green button in that place, after the changes it looks like this https://jmp.sh/dBWTVjF
Makes more sense?

jeroendesloovere:
The current colors make more sence to me.
Red = delete
Green = add

schermafbeelding 2018-07-02 om 16 15 30

jeroendesloovere:
Perhaps we should change the "Publish" color to green too.

schermafbeelding 2018-07-02 om 16 17 27

freshface:
I see green as a feedback/success color, not as a primary button color.

jeroendesloovere:
@freshface thanks for your PR!
I believe this points out a greater problem in the UI colors.

- We practically always use "red" for "deletions". Which is good.
- But we do use "green", "grey" and "blue" for additions/savings... This could indeed be better and be made more uniform. 

@Katrienvh Do you know if there was any particular reason this was done like this?

freshface:
A red button is fine by me because it removes something which has more impact then adding something.

carakas:
not going to change this one

PR #2597, opened 2018-07-31
Style collections as tabs

carakas:
## Type

- Feature

## Pull request description

Styling to make a collections look like a tab instead of a box within a box

Before
screen shot 2018-07-31 at 13 14 35
screen shot 2018-07-31 at 13 18 56

After adding the following to the form type
```php
        $builder->add(
            'downloads',
            CollectionType::class,
            [
                …
                'attr' => [
                    'class' => 'tab-pane fade tab-pane-collection',
                ],
            ]
        );
```

screen shot 2018-07-31 at 13 19 26
screen shot 2018-07-31 at 13 14 15


src/Backend/Core/Layout/Sass/components/_tabs.scss
```diff
@@ -77,3 +77,11 @@
     background-color: $white;
   }
 }
+
+.tab-pane-collection {
+  margin: -31px -26px -56px;
```
jonasdekeukelaere: Those are some very specific numbers

carakas: they are needed to cancel out the padding and margin of the tab container

PR #2612, opened 2018-08-07
Better flow for adding media

### LINKED ISSUE #2566: Media library: add media change destination folder after uploading
## Type

- Non critical bug

## Problem description

You can change the destination folder after uploading media. But the media is alreay uploaded in an other folder.

I suggest the easiest way is to disabled the destination folder after uploading or find a way to make it clear for the use that the first uploaded item is saved in an other folder when they edit the folder after uploading something.

carakas:
## Type

- Enhancement

Fixes #2566 

## Pull request description

Before:
screen shot 2018-08-07 at 11 02 05
After:
screen shot 2018-08-07 at 10 55 12
screen shot 2018-08-07 at 10 55 52

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Resources/views/BackendMediaLibraryUpload.html.twig
+++ b/src/Backend/Modules/MediaLibrary/Resources/views/BackendMediaLibraryUpload.html.twig
@@ -76,128 +76,141 @@
   </div>
 </script>
 
-{# Step 0 - Choose type #}
-<div id="uploadMediaTypeBox" class="row fork-module-content">
-  <div class="col-md-12">
-    <div class="panel panel-default">
-      <div class="panel-heading">
-        <h2 class="panel-title">{{ 'msg.MediaUploadThisType'|trans }}</h2>
-      </div>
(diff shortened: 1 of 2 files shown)


Katrienvh:
I should place the first two panels in a row, each in a .col-md-6

src/Backend/Modules/MediaLibrary/Resources/views/BackendMediaLibraryUpload.html.twig
```diff
@@ -76,128 +76,141 @@
...
-          </a>
-        </p>
+    </div>
+  </div>
+
+  <div class="row fork-module-actions">
+    <div class="col-md-12">
+      <div class="btn-toolbar">
+        <div class="btn-group pull-right" role="group">
+          {{ macro.buttonIcon('', '', 'lbl.Next'|trans|ucfirst, 'btn-primary', {"data-role":"uploadMediaGoToStep2", "type":"button"}) }}
```
Katrienvh: All buttons have an icon, add here one two

PR #2613, opened 2018-08-08
Sanitize querystring before redirect after profile login

jonasdekeukelaere:
## Type
- Security

## Pull request description
Only accepts internal pages as queryString for redirect after login.


jonasdekeukelaere:
Backend added 1f6686b1f

PR #2625, opened 2018-08-28
 Create a TitleType to easily add titles to your Fork CMS form!

StijnVrolijk:
## Type

- Feature

## Pull request description
Create a TitleType to easily add titles to your Fork CMS form!

Provides everything from the required classes to the label. All overwritable, of course.

### DIFF
--- a/src/Common/Form/TitleType.php
+++ b/src/Common/Form/TitleType.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Common\Form;
+
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class TitleType extends TextType
+{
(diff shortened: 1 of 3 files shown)


src/Common/Form/TitleType.php
```diff
@@ -0,0 +1,24 @@
...
+class TitleType extends TextType
+{
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        parent::configureOptions($resolver);
+
+        $resolver->setDefaults(
+            [
+                'label' => 'lbl.Title',
+                'required' => true,
```
carakas: is required not always true by default?

StijnVrolijk: It is but I decided to explicitly define it because it was defined in some other places (like ContentBlocks) and it's also clearer for users.

PR #2635, opened 2018-09-05
Don't require a database password

### LINKED ISSUE #2623: Application install - database password is required
## Type

- Enhancement
 
## Problem description

When install application in step 4 database password is required. When I have user "root" without password on my local dev environment I cannot continue with the installation. I undrestand that use no password is not save but for dev usage it should not be a problem. So to finish installation I have to update the file src\ForkCMS\Bundle\InstallerBundle\Form\Type\DatabaseType.php and add required to false.

 ->add(
                'databasePassword',
                PasswordType::class,
                [
                    'required' => false,
                ]

## Steps to reproduce

Process the installation to the step 4.

Database password is not required.

carakas:
## Type

- Non critical bugfix

Fixes #2623

## Pull request description

Don't require a password for the database, though you really should use one 👍

### DIFF
--- a/src/ForkCMS/Bundle/InstallerBundle/Form/Type/DatabaseType.php
+++ b/src/ForkCMS/Bundle/InstallerBundle/Form/Type/DatabaseType.php
@@ -23,35 +23,26 @@ class DatabaseType extends AbstractType
         $builder
             ->add(
                 'databaseHostname',
-                TextType::class,
-                [
-                    'required' => true,
-                ]
+                TextType::class
             )
             ->add(
                 'databasePort',
-                TextType::class,
-                [
-                    'required' => true,
-                ]
+                TextType::class
             )
             ->add(
                 'databaseName',
-                TextType::class,
-                [
-                    'required' => true,
-                ]
+                TextType::class
             )
             ->add(
                 'databaseUsername',
-                TextType::class,
-                [
-                    'required' => true,
-                ]
+                TextType::class
             )
             ->add(
                 'databasePassword',
-                PasswordType::class
+                PasswordType::class,
+                [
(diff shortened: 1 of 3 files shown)


carakas:
@tijsverkoyen would it be an idea to only allow this in case fork is running in debug?

carakas:
I'm not sure if safari has html5 validation or not at the moment, but if not it actually is possible to do so in safari at the moment

tijsverkoyen:
Still not a fan. 
1. I think we can expect someone who wants to develop with Fork CMS is able to set a password on their database.
2. I you only allow it in debug, we will get questions on why it is not allowed in production
3. In my opinion it is our duty to preach best practices, and not implement features against them.

jamius19:
Today I was trying out Fork for the first time in my local dev environment and the required password field was a problem. Almost quit on Fork because of it.

With all due respect, I don't think anyone needs to be bound to use a password on their dev machine only because somebody else thought so. I want my software to be up and running quickly, not making a fuss about my security practices on my local dev machine. If I wanted that, then I'd be using local **.dev** domains with my own certificates for SSL instead of simply using **.tess**.

Sure it is a good practice, but maybe some people want this convenience on our local machines. We can sometimes leave the good practices for a production server, right?

Changing the root user password could break myriads of other Softwares running on my system. Technically I can make a new user for just the fork's database, then again I can't quite get why it has to be made so difficult whereas even Wordpress allows for a passwordless user.

I think giving a polite reminder about using password is the right way. Or maybe even showing a popup reminding that setting a password is the recommended way is a lot better than requiring me to use password **exclusively for Fork.**

Anyway, just my two cents. Have a nice day.

PR #2636, opened 2018-09-06
When using the modulesettings in your services the installation won't work anymore.

carakas:
## Type

- Non critical bugfix

## Pull request description

This pr fixes that issue by using actual implementations instead of stdClass

### DIFF
--- a/app/config/config_install.yml
+++ b/app/config/config_install.yml
@@ -50,15 +50,8 @@ services:
     mailer_configurator:
         class: stdClass
 
-    # dummy service during the installation
-    fork.settings:
-        class: stdClass
-        public: true
-
-    # dummy service during the installation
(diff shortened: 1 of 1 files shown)


app/config/config_install.yml
```diff
@@ -50,15 +50,8 @@ services:
     mailer_configurator:
         class: stdClass
 
-    # dummy service during the installation
-    fork.settings:
```
jonasdekeukelaere: Is the service not missing now?

carakas: No, we were overwriting it

carakas:
@jonasdekeukelaere it is fixed now, no clue why it needs that service like that but when I add it with the same config as in the `config.yml` it works

PR #2639, opened 2018-09-07
Phpstan integration

ohvitorino:
## Type

- Enhancement

## Pull request description
This PR aims to add static code analysis to ForkCMS.
The travis configuration was modified to have two stages. One that executes the unit and functional tests in parallel, and another stage (Code Quality) that runs PHPCS and PHPStan.

PHPStan does at this moment only level 0 analysis and my idea is to increase the level with time, so that the code can become more robust.


phpstan.neon
```diff
@@ -0,0 +1,22 @@
+includes:
```
carakas: is this file already blocked by the .htaccess ?

ohvitorino:
Yes @justcarakas

phpstan.neon
ohvitorino: Yes @carakas

PR #2642, opened 2018-09-13
Fix mail template

bart-webleads:
## Type
- Non critical bugfix

## Pull request description
Prevent html/javascript injection in form builder e-mail template.


carakas:
@bart-webleads can you fix this in a way that this still works
https://github.com/forkcms/forkcms/pull/1854

carakas:
@bart-webleads maybe adding `|striptags('<br>')` might be a solution?

bart-webleads:
Drop this pull request. I'll create a new one with more fixes.

PR #2644, opened 2018-09-13
Media library: Delete multiple items on index

### LINKED ISSUE #2560: Media library: Delete multiple items on index
## Type
- Feature

## Problem description

I think it's strange that you can't choose delete as action to do with al the selected items. 

## Steps to reproduce

https://demo.fork-cms.com/private/en/media_library/media_item_index
If you select some items, you can do the same action on all the items.

I my head it's logic that at least you can delete them all? No the only action you can do is move.

thibautvincent:
## Type

- Feature

fixes #2560

## Pull request description

This PR adds the functionality of deleting multiple media items at a time.
There is a new option in the dropselect for deleting media.
After selecting the option, you'll have to confirm your action before the media actually gets deleted.


carakas:
@thibautvincent thank you for this pr

PR #2646, opened 2018-10-02
Preserve sorted order of images

jacob-v-dam:
## Type
Non critical bugfix

## Pull request description
When a user sorted the images it would save perfectly. But when the
user reloads the edit page the order would be gone. By this the order
is preserved and can be used as expected.

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Ajax/MediaItemGetAllById.php
+++ b/src/Backend/Modules/MediaLibrary/Ajax/MediaItemGetAllById.php
@@ -34,6 +34,12 @@ class MediaItemGetAllById extends BackendBaseAJAXAction
             return [];
         }
 
-        return $this->get('media_library.repository.item')->findById($ids);
+        $mediaItems = [];
+
+        foreach ($ids as $id) {
+            $mediaItems[] = $this->get('media_library.repository.item')->findOneById($id);
+        }
+
+        return $mediaItems;
     }
 }


src/Backend/Modules/MediaLibrary/Ajax/MediaItemGetAllById.php
```diff
@@ -34,6 +34,12 @@ protected function getMediaItems(): array
             return [];
         }
 
-        return $this->get('media_library.repository.item')->findById($ids);
+        $mediaItems = [];
```
carakas: why not just sort them in the method itself?

jacob-v-dam: The id's are passed in a given order. This is the best way to sort the images in my opinion.

carakas: the current implementation will get all items in 1 query, you are running a separate query per item, so that solution is not acceptable.
You can use the initial implementation and then sort it by the order of the ids.

jacob-v-dam: Ok, but wouldn't it be better to retrieve the images by the given library instead of all the id's? It would be easier to preserve the order and it makes more sense instead of sending all the id's.

carakas:
Let me know if you would like to finish this one

PR #2648, opened 2018-10-03
Don't show description in autocomplete suggestion

### LINKED ISSUE #2647: Search widget autosuggest returns full html
## Type

- Non critical bug

## Problem description

The autosuggest in the search widget returns the entire page content as HTML

## Steps to reproduce

go to the forkcms demo site, type a search in the search widget (top of page) see the result.

Return the matching word instead of the complete html

vanMeerdervoort:
## Type
- Non critical bugfix

fixes #2647 

## Pull request description
Removes the full HTML description rendering in the autosuggest dropdown or the search widget.

### DIFF
--- a/src/Frontend/Modules/Search/Js/Search.js
+++ b/src/Frontend/Modules/Search/Js/Search.js
@@ -70,7 +70,7 @@ jsFrontend.search =
       source: searchEngine,
       templates: {
         suggestion: function (data) {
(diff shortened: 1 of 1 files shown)


carakas:
not a big fan of this solution, I think it would be better to pass a proper short description fo pages to the search indexes

carakas:
I'll take a look if I can find a solution that works in all cases

PR #2649, opened 2018-10-05
Only build the navigation cache if vital information has changed

StijnVrolijk:
## Type
- Enhancement

## Pull request description
Only build the navigation cache if vital information has changed. When editing a page we don't need to rebuild the navigation cache if we only changed a typo or a block. It should only be rebuilt if the title, navigation_title or published properties changed.

This is a rather small enhancement but might be useful for sites with a lot of pages.

### DIFF
--- a/src/Backend/Modules/Pages/Actions/Edit.php
+++ b/src/Backend/Modules/Pages/Actions/Edit.php
@@ -765,7 +765,17 @@ class Edit extends BackendBaseActionEdit
 
         $this->saveTags($page['id']);
 
-        BackendPagesModel::buildCache(BL::getWorkingLanguage());
+        $cacheShouldBeUpdated = !(
+            $this->record['title'] === $page['title']
+            && $this->record['navigation_title'] === $page['navigation_title']
+            && $this->record['navigation_title_overwrite'] === $page['navigation_title_overwrite']
+            && $this->record['hidden'] === $page['hidden']
+        );
+
+        // build cache
+        if ($cacheShouldBeUpdated) {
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -765,7 +765,17 @@ private function validateForm(): void
...
+            $this->record['title'] === $page['title']
+            && $this->record['navigation_title'] === $page['navigation_title']
+            && $this->record['navigation_title_overwrite'] === $page['navigation_title_overwrite']
+            && $this->record['hidden'] === $page['hidden']
+        );
+
+        // build cache
+        if ($cacheShouldBeUpdated) {
+            BackendPagesModel::buildCache(BL::getWorkingLanguage());
+        }
```
WouterSioen: This looks like something that could perfectly be done as a side effect using an event + listener.

StijnVrolijk: yup, sure :) I only made this PR because I had to do this for an older project. This should definitely be handled by an event listener/subscriber in the future.

WouterSioen: Might be a good moment to do this? 😄

StijnVrolijk: A good time for Fork, yes. A good time for me? No :p too much work unfortunately

ohvitorino: It's Hacktober fest @StijnVrolijk, don't you want a new t-shirt?

PR #2650, opened 2018-10-05
Bugfix // Fix search settings error

wouterverstuyf:
## Type

- Non critical bugfix

- Saving the search settings with not all modules enabled generates an error on the searchable column (expects int and saved as 'null')

## Pull request description
- casting to int to save a 0 instead of 'null'

### DIFF
--- a/src/Backend/Modules/Search/Actions/Settings.php
+++ b/src/Backend/Modules/Search/Actions/Settings.php
@@ -162,7 +162,7 @@ class Settings extends BackendBaseActionEdit
 
                 // module search
                 foreach ((array) $this->modules as $module) {
-                    $searchable = $this->form->getField('search_' . $module['module'])->getChecked();
+                    $searchable = (int) $this->form->getField('search_' . $module['module'])->getChecked();
                     $weight = $this->form->getField('search_' . $module['module'] . '_weight')->getValue();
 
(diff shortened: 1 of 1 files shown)


StijnVrolijk:
@wouterverstuyf can you show us the error you are describing?

```php
BackendSearchModel::insertModuleSettings(string $module, string $searchable, string $weight)```

The method expects a string so if there is an issue with inserting the value into the database the issue is either in that method or the `getChecked()` method of `SpoonFormCheckbox`

wouterverstuyf:
Hmz, the searchable columns expects an integer?

schermafbeelding 2018-10-05 om 11 32 07

See below the error:

schermafbeelding 2018-10-05 om 11 31 17

StijnVrolijk:
@wouterverstuyf you're receiving an empty string as `$searchable`, can you check what the `->getChecked()` method in your form returns?

wouterverstuyf:
Yeah, I know, the first is a bool(true) the second one is empty, cause the second item in the search modules is not checked.
So casting it to an integers makes it 0 so the error is gone.

schermafbeelding 2018-10-05 om 11 39 04

But this is strange. The function is expecting a string like you said, but the db searchable column field is a Tinyint...

carakas:
@wouterverstuyf it should be a boolean value. Booleans are saved as integers being 0 or 1.
The string thing is probably a left over from when we used `Y` and `N`

PR #2653, opened 2018-10-11
Prevent installation of a theme if it contains .tpl files

StijnVrolijk:
## Type
- Enhancement

## Pull request description
Prevent installation of a theme if it contains .tpl files

### DIFF
--- a/src/Backend/Modules/Extensions/Installer/Data/locale.xml
+++ b/src/Backend/Modules/Extensions/Installer/Data/locale.xml
@@ -751,15 +751,6 @@
       <item type="error" name="IncompatibleTheme">
         <translation language="nl"><![CDATA[Het thema "%1$s" bevat bestanden die niet compatibel zijn met deze versie van Fork CMS.]]></translation>
         <translation language="en"><![CDATA[Theme "%1$s" contains files that are incompatible with this version of Fork CMS.]]></translation>
-        <translation language="fr"><![CDATA[Theme "%1$s" contains files that are incompatible with this version of Fork CMS.]]></translation>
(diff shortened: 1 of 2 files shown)


src/Backend/Modules/Extensions/Installer/Data/locale.xml
```diff
@@ -748,6 +748,19 @@
         <translation language="en"><![CDATA[The template file is missing.]]></translation>
         <translation language="pl"><![CDATA[Brakuje pliku szablonu.]]></translation>
       </item>
+      <item type="error" name="IncompatibleTheme">
+        <translation language="nl"><![CDATA[Het thema "%1$s" bevat bestanden die niet compatibel zijn met deze versie van Fork CMS.]]></translation>
+        <translation language="en"><![CDATA[Theme "%1$s" contains files that are incompatible with this version of Fork CMS.]]></translation>
+        <translation language="fr"><![CDATA[Theme "%1$s" contains files that are incompatible with this version of Fork CMS.]]></translation>
+        <translation language="ru"><![CDATA[Theme "%1$s" contains files that are incompatible with this version of Fork CMS.]]></translation>
+        <translation language="es"><![CDATA[Theme "%1$s" contains files that are incompatible with this version of Fork CMS.]]></translation>
+        <translation language="de"><![CDATA[Theme "%1$s" contains files that are incompatible with this version of Fork CMS.]]></translation>
```
jonasdekeukelaere: Maybe lose the English translations for the locales that you didn't translate?

StijnVrolijk: Shouldn't we at least default to something that some people might actually understand rather than having no translation at all?

carakas: Doesn't fork default to english when the translation is missing?

PR #2663, opened 2018-10-17
Add better breadcrumb support

### LINKED ISSUE #2629: Media library breadcrumbs
## Type

- Enhancement

## Problem description

I should add the folder where you are working in in the breadcrumb. That would make the view way more user friendly.

## Steps to reproduce

https://demo.fork-cms.com/private/en/media_library/media_item_index?folder=1&token=vq51skpbw5

StijnVrolijk:
## Type
- Feature

Fixes #2629

## Pull request description
Adds a simple method to the backend `Header` class with which we can easily add a breadcrumb item to the breadcrumbs. This is useful for display what you are uploading files in for example.

Before:
image

After:
image


src/Backend/Core/Layout/Templates/breadcrumb.html.twig
```diff
@@ -31,4 +31,7 @@
       </li>
     {% endif %}
   {% endfor %}
+  {% if breadcrumbDetail %}
```
carakas: have you tried this on a page where it wasn't defined?

src/Backend/Modules/MediaLibrary/Actions/MediaItemUpload.php
```diff
@@ -44,9 +44,11 @@ protected function parse(): void
         /** @var int|null $mediaFolderId */
         $mediaFolderId = ($this->mediaFolder instanceof MediaFolder) ? $this->mediaFolder->getId() : null;
 
+
```
carakas: useless whitespace

PR #2664, opened 2018-10-22
Fix related tags type error

bart-webleads:
## Type
- Critical bugfix

## Pull request description
The related tags widget did not work, because of an error.
Type error: Argument 2 passed to Frontend\Modules\Tags\Engine\Model::getForItem() must be of the type integer, null given, called in src/Frontend/Modules/Tags/Widgets/Related.php on line 152

The getIdForTags function returns an id and not a record as assumed in the related widget code.
Also removed useless comments in this function.


carakas:
maybe try removing the comments in a separate commit the next time, now it was hard to find the difference @bart-webleads but thanks for the fix 😄

PR #2666, opened 2018-10-23
Use a valid date format

Katrienvh:
## Type
- Enhancement

## Pull request description
Improve readability of the time datetime property format. In certain cases the w3c validator throws an error.

### DIFF
--- a/src/Frontend/Modules/Blog/Layout/Templates/Archive.html.twig
+++ b/src/Frontend/Modules/Blog/Layout/Templates/Archive.html.twig
@@ -41,7 +41,7 @@
             {% for item in items %}
               <tr>
                 <td class="date">
-                  <time itemprop="datePublished" datetime="{{ item.publish_on|date('Y-m-d\TH,i,s' ) }}">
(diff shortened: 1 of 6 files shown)


src/Frontend/Modules/Blog/Layout/Templates/Archive.html.twig
```diff
@@ -41,7 +41,7 @@
             {% for item in items %}
               <tr>
                 <td class="date">
-                  <time itemprop="datePublished" datetime="{{ item.publish_on|date('Y-m-d\TH,i,s' ) }}">
+                  <time itemprop="datePublished" datetime="{{ item.publish_on|date('Y-m-d\\TH:i:s.vP' ) }}">
```
carakas: do there need to be 2 backslashes?

StijnVrolijk: Yes, for some reason, if you use `\T` if will output something like `CEST`.

PR #2670, opened 2018-10-31
Catch mailmotor spamming errors

carakas:
## Type

- Non critical bugfix

## Pull request description

When spammers fill in the mailmotor subscribe they can trigger the following error:
```json
{"type":"http://developer.mailchimp.com/documentation/mailchimp/guides/error-glossary/","title":"Invalid Resource","status":400,"detail":"tcr10112@gmail.com has signed up to a lot of lists very recently; we're not allowing more signups for now","instance":"xxxxxxxxx"}
```
this will catch it and keep spammers from messing up your error log

### DIFF
--- a/src/Frontend/Modules/Mailmotor/Actions/Subscribe.php
+++ b/src/Frontend/Modules/Mailmotor/Actions/Subscribe.php
@@ -69,6 +69,8 @@ class Subscribe extends FrontendBaseBlock
                 throw $exception;
             }
 
+            $this->getContainer()->get('logger')->error('Mailmotor Subcribe Mailchimp error: ' . $reason);
+
             $this->template->assign('mailmotorSubscribeHasFormError', true);
             $this->template->assign('form', $form->createView());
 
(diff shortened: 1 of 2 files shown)


src/Frontend/Modules/Mailmotor/Actions/Subscribe.php
```diff
@@ -61,6 +62,20 @@ public function execute(): void
...
 
             $doubleOptIn = false;
+        } catch (Exception $exception) {
+            $reason = json_decode($exception->getMessage());
+            // check if the error is one from mailchimp
+            if ($reason === false) {
+                throw $exception;
+            }
+
+            $this->template->assign('mailmotorSubscribeHasFormError', true);
```
tijsverkoyen: Wouldn't it be better to have a separate error in the template for Mailchimp specific errors? And also passing the reason to it?

carakas: We could, but this would work in all custom themes as well without modification.
Apart from that, I think it is a little bit dangerous to just parse the errors from MailChimp since I just get a bad request back, but you never know what info will be included. On top of that are the returned errors in English, so on a site in a different language you would get multiple languages on a page with no proper way of translating it

tijsverkoyen: maybe log the error into a error log, because with your fix you will never know what the error was.

PR #2674, opened 2018-11-09
Use find method instead of magic method

ohvitorino:
## Type

- Enhancement

## Pull request description
This PR replaces the usage of the magic method `findOneById` with the method `find` in the MediaLibrary and MediaGallery modules.


carakas:
@StijnVrolijk I think this one can be included in 5.4.2. Any reason why you changed it to 5.5.0?

StijnVrolijk:
@carakas sure but since it's an enhancement I thought it'd be better to release it in a feature release.

PR #2687, opened 2018-12-14
Remove spoon implementation for backend csv

### LINKED ISSUE #2099: Change the implementation of Backend/Core/Engine/Csv away from spoon library
## Type

- Enhancement

## Problem description

We are using Spoon csv, since spoon is deprecated we should swap it out

Keep the same external api but let it work on a different library or implementation

Senjutsuu:
## Type
- Enhancement

fixes #2099 

## Pull request description
This pull request will no longer implement spoon library for CSV handling in Backend/Core/Engine/Csv.
Instead we will use the Symfony CSV-encoder and serializer.


src/Backend/Core/Engine/Csv.php
```diff
@@ -54,6 +60,7 @@ public static function outputCSV(
         );
     }
 
+    // we are not using the line ending for now.
```
WouterSioen: Not sure if it's safe to drop this feature? Will all software handling csv's on multiple OS'es still be able to handle this?

StijnVrolijk: @WouterSioen Symfony's [CSV encoder](https://api.symfony.com/3.3/Symfony/Component/Serializer/Encoder/CsvEncoder.html) uses `fputcsv` to write CSV files so I assume it relies on PHP to properly detect the correct line endings. Not too sure if that's a good idea, though.

@Senjutsumo can you test on some different machines (OS X, Windows (with Excel), ...) to see if the CSV encoder always write readable files?

If it doesn't we could look at a CSV generating package on packagist.

carakas:
I would use https://csv.thephpleague.com instead

carakas:
replaced by #3160

PR #2695, opened 2018-12-19
2681 change image preview

### LINKED ISSUE #2681: Change image preview
## Type

- Enhancement

## Problem description

If you change an image in a module, the preview doesn't change, only the name of the file changes. It would be user-friendlier to change the preview at the same time. 

## Steps to reproduce

Change an image in a module

Make the preview the same as the uploaded file immediately.

Senjutsuu:
## Type
- Enhancement

fixes #2681 

## Pull request description
Show an image preview after upload when replacing an existing image.
For Symfony imageType fields the data attribute is always added to the imageType.
I have added support for the current blog module and image field in pages.

### DIFF
--- a/src/Backend/Core/Js/backend.js
+++ b/src/Backend/Core/Js/backend.js
@@ -1197,6 +1197,25 @@ jsBackend.forms = {
     jsBackend.forms.meta()
     jsBackend.forms.datePicker()
     jsBackend.forms.bootstrapTabFormValidation()
+    jsBackend.forms.imagePreview()
+  },
+
+  imagePreview: function () {
+    $('input[type=file]').on('change', function () {
+      // make sure we are uploading an image by checking the name attribute
+      if ($(this).get(0).getAttribute('data-fork-cms-role') === 'image-field' && $(this).get(0).files && $(this).get(0).files[0]) {
+        // get the image closest to the upload button
+        let imagePreview = $(this).closest('.panel-body').find('.img-thumbnail')
+        // use FileReader to get the url
+        let reader = new FileReader()
+
+        reader.onload = function (event) {
+          imagePreview.attr('src', event.target.result)
+        }
+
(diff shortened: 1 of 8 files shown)


src/Backend/Core/Js/backend.js
```diff
@@ -1197,6 +1197,25 @@ jsBackend.forms = {
     jsBackend.forms.meta()
     jsBackend.forms.datePicker()
     jsBackend.forms.bootstrapTabFormValidation()
+    jsBackend.forms.imagePreview()
+  },
+
+  imagePreview: function () {
+    $('input[type=file]').on('change', function () {
+      // make sure we are uploading an image by checking the name attribute
+      if ($(this).get(0).getAttribute('data-fork-cms-role') === 'image-field' && $(this).get(0).files && $(this).get(0).files[0]) {
```
StijnVrolijk: It would be a good idea to put `$(this).get(0)` in a variable.

src/Backend/Core/Js/backend.js
```diff
@@ -1197,6 +1197,25 @@ jsBackend.forms = {
...
+        // get the image closest to the upload button
+        let imagePreview = $(this).closest('.panel-body, .tab-pane').find('.img-thumbnail, .img-responsive')
```
StijnVrolijk: I'd like to see some `data-fork-cms-roles` here as well, rather than relying on bootstrap classes.

src/Backend/Core/Js/backend.js
```diff
@@ -1197,6 +1197,26 @@ jsBackend.forms = {
     jsBackend.forms.meta()
     jsBackend.forms.datePicker()
     jsBackend.forms.bootstrapTabFormValidation()
+    jsBackend.forms.imagePreview()
+  },
+
+  imagePreview: function () {
```
carakas: @Senjutsuu can you add this to the frontend as well? That way it will also work for the image types there

PR #2703, opened 2018-12-20
Improve the thumbnail generation console command

StijnVrolijk:
## Type
- Enhancement

## Pull request description
Improvements made to UX as well as inner workings

image


Katrienvh:
I tested the basic usage and it works fine

PR #2706, opened 2018-12-20
Blog. Make blog image clickable.

Sellato:
## Type

- Enhancement
- Feature

## Pull request description

Small fix to make blog index image clickable :)


src/Frontend/Modules/Blog/Actions/Detail.php
```diff
@@ -173,7 +173,7 @@ private function addTwitterCard(): void
         $this->header->setTwitterCard(
             $this->blogPost['title'],
             $this->blogPost['meta_description'],
-            FRONTEND_FILES_URL . '/Blog/images/source/' . $this->blogPost['image']
+            SITE_URL . '/src/Frontend/Files/Blog/images/source/' . $this->blogPost['image']
```
carakas: wrong solution, probably you need `SITE_URL . '/' . FRONTEND_FILES_URL  . '/Blog/images/source/' . $this->blogPost['image']`
Hardcoding `/src/Frontend/Files` is not allowed

Sellato: I made changes. 
This peace of cod `'/'` is unneeded :)

PR #2727, opened 2019-01-14
Translate google maps

carakas:
## Type

- Enhancement

## Pull request description

Translates the google maps used on the site to the interface language or the language that the site is shown in instead of the browser language

### DIFF
--- a/src/Backend/Modules/Location/Actions/Edit.php
+++ b/src/Backend/Modules/Location/Actions/Edit.php
@@ -47,10 +47,8 @@ class Edit extends BackendBaseActionEdit
                 $this->redirect(BackendModel::createUrlForAction('Index', 'Settings'));
             }
 
-            $interfaceLanguage = BL::getInterfaceLanguage();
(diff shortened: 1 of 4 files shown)


src/Backend/Modules/Location/Actions/Edit.php
```diff
@@ -47,8 +47,11 @@ public function execute(): void
                 $this->redirect(BackendModel::createUrlForAction('Index', 'Settings'));
             }
 
-            // add js
-            $this->header->addJS('https://maps.googleapis.com/maps/api/js?key=' . $apikey);
+            $interfaceLanguage = BL::getInterfaceLanguage();
+            $googleLanguage = $interfaceLanguage . '_' . strtoupper($interfaceLanguage);
```
StijnVrolijk: Where did you find the information that states the language should be in the `nl_NL` format?

I found it's not necessary ( https://developers.google.com/maps/documentation/javascript/localization ) which is fortunate because this code will generate stuff like `en_EN` which does not exist.

carakas: Changed in bb88e9f

PR #2732, opened 2019-01-17
Copy page button

### LINKED ISSUE #2721: Please add a Duplicate Page button
## Type

- Enhancement
- Feature

## Problem description

> Please can we add a "Copy" button to the CMS so that existing landing pages can be easily duplicated?  This means that when we want to create 3 very similar landing pages, we can copy and paste all the content from an existing page in one click, similar to the functionality available in the WordPress Divibuilder.

## Steps to reproduce

> When logged into Fork, under the Pages tab, please add a "Duplicate page" option.

> By adding the button we will be able to  copy and paste all the content from an existing page into a new page, in one click, similar to the functionality available in the WordPress Divibuilder.

carakas:
## Type

- Feature

fixes #2721

## Pull request description

This pr adds the option to create a new page based on the page you are currently viewing/editing.
During this implementation the missing authentication tab on page add was also added like it exists on the page edit action
screenshot 2019-01-24 at 10 08 27
screenshot 2019-01-24 at 10 08 13


freshface:
Would 'Copy this page' not be a better translation?

carakas:
The reason why I called it clone is because of the definition of the words

    clone - create something new based on something that exists.
    copying - copy from something that exists to something else (that also already exists).

mathiashelin:
Maybe use `duplicate`?

carakas:
Changed it back to copy after the feedback

PR #2738, opened 2019-01-25
Check only for the existence of the Facebook API ID

ohvitorino:
## Type

- Non critical bugfix

## Pull request description

When loading the facebook snippet Fork is checking if one of the fields Facebook Admin IDs or Application ID is filled in. If we only fill in Admin IDs and no Application ID, this leads to an application error (500) on the frontend. This PR changes the verification to only take in consideration the Application ID, since the Admin IDs don't seem to be necessary to load the FB script.

### DIFF
--- a/src/Frontend/Core/Engine/Footer.php
+++ b/src/Frontend/Core/Engine/Footer.php
@@ -44,12 +44,10 @@ class Footer extends KernelLoader
         $this->template->assignGlobal('footerLinks', $footerLinks);
 
         $siteHTMLFooter = (string) $this->get('fork.settings')->get('Core', 'site_html_footer', null);
-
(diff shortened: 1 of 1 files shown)


carakas:
is the admin id used somewhere else? Otherwise it might be an idea to remove the option to fill it in?

ohvitorino:
@carakas I don't think it gets used anywhere else. But I wonder if it isn't a good idea to let it stay since developers adding new modules can still use this setting without having to add it to the its module preferences of changing the core to re-add the setting.

carakas:
Ok, but maybe check for the setting in the module that indicates that that key needs to be requested

ohvitorino:
I think that that check is not being done for facebook or twitter.

PR #2743, opened 2019-01-29
Delete space between asterisk and label

sarahmey:
## Type
- Non critical bugfix

## Pull request description
Delete the space between asterisk and label for form fields


src/Backend/Core/Layout/Templates/FormLayout.html.twig
```diff
@@ -54,8 +54,11 @@
 {% endblock checkbox_radio_label %}
 
 {%- block label_asterisk -%}
-  {% import 'Layout/Templates/macros.html.twig' as macro %}
-  {{ macro.required }}
+  {% spaceless %}
+    {% import 'Layout/Templates/macros.html.twig' as macro %}
+    {{ macro.required }}
+  {% endspaceless %}
+
```
carakas: Is this empty line required?

carakas:
recreated here since the tests here kept failing without an obvious reason while they work when I recreated the PR
https://github.com/forkcms/forkcms/pull/2761

PR #2745, opened 2019-01-30
Autocomplete attributes on Formbuilder input fields

Senjutsuu:
## Type
- Enhancement

## Pull request description
Autocomplete attributes are available for the input text and date fields in the Formbuilder module
screen shot 2019-01-30 at 16 15 07
screen shot 2019-01-30 at 16 15 43

### DIFF
--- a/src/Backend/Modules/FormBuilder/Actions/Edit.php
+++ b/src/Backend/Modules/FormBuilder/Actions/Edit.php
@@ -174,6 +174,21 @@ class Edit extends BackendBaseActionEdit
         $this->form->addText('datetime_classname');
         $this->form->addText('datetime_error_message');
 
+        // map the values to replace them with the backend translations
+        // use array combine to set the keys as the autocomplete values instead of key values
+        $this->form->addDropdown(
+            'datetime_autocomplete',
+            array_map(
+                function ($value) {
+                    return $value . ' (' . BL::getLabel('Autocomplete_' . str_replace('-', '_', $value)) . ')';
+                },
+                array_combine(
+                    Autocomplete::POSSIBLE_VALUES,
+                    Autocomplete::POSSIBLE_VALUES
+                )
+            )
+        );
+
         // dropdown dialog
         $this->form->addText('dropdown_label');
         $this->form->addText('dropdown_values');
--- a/src/Backend/Modules/FormBuilder/Actions/Edit.php
+++ b/src/Backend/Modules/FormBuilder/Actions/Edit.php
@@ -97,7 +97,7 @@ class Edit extends BackendBaseActionEdit
         $this->form->addText('textbox_validation_parameter');
         $this->form->addText('textbox_error_message');
 
-        $this->form->addDropdown('textbox_autocomplete', Autocomplete::getAutocompleteValues());
+        $this->form->addDropdown('textbox_autocomplete', Autocomplete::getValuesForDropdown());
 
         // textarea dialog
         $this->form->addText('textarea_label');
@@ -161,7 +161,7 @@ class Edit extends BackendBaseActionEdit
         $this->form->addText('datetime_classname');
         $this->form->addText('datetime_error_message');
 
-        $this->form->addDropdown('datetime_autocomplete', Autocomplete::getAutocompleteValues());
+        $this->form->addDropdown('datetime_autocomplete', Autocomplete::getValuesForDropdown());
 
         // dropdown dialog
         $this->form->addText('dropdown_label');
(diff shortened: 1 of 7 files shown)


src/Backend/Modules/FormBuilder/Engine/Autocomplete.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace Backend\Modules\FormBuilder\Engine;
+
+final class Autocomplete
```
StijnVrolijk: You might want to add some PHPDoc stating where this information comes from

src/Backend/Modules/FormBuilder/Actions/Edit.php
```diff
@@ -96,6 +97,21 @@ private function loadForm(): void
         $this->form->addText('textbox_validation_parameter');
         $this->form->addText('textbox_error_message');
 
+        // map the values to replace them with the backend translations
+        // use array combine to set the keys as the autocomplete values instead of key values
+        $this->form->addDropdown(
+            'textbox_autocomplete',
+            array_map(
+                function ($value) {
```
StijnVrolijk: You can use `function (string $value): string {` here

src/Backend/Modules/FormBuilder/Actions/Edit.php
```diff
@@ -174,6 +174,21 @@ function ($value) {
         $this->form->addText('datetime_classname');
         $this->form->addText('datetime_error_message');
 
+        // map the values to replace them with the backend translations
+        // use array combine to set the keys as the autocomplete values instead of key values
+        $this->form->addDropdown(
+            'datetime_autocomplete',
+            array_map(
```
StijnVrolijk: Since you do this map twice (once in Add, once in Edit), it might be a good idea to extract it to a method in Autocomplete, `Autocomplete::getValuesForDropdown()` for example.

src/Backend/Modules/FormBuilder/Ajax/SaveField.php
```diff
@@ -247,6 +252,10 @@ public function execute(): void
             $settings['classname'] = \SpoonFilter::htmlspecialchars($classname);
         }
 
+        if ($autocomplete != '') {
```
StijnVrolijk: you can simplify this to `$settings['autocomplete'] = in_array($autocomplete, Autocomplete::POSSIBLE_VALUES) ? $autocomplete : '';`

This also removes the need for lines 61-63.

src/Backend/Modules/FormBuilder/Actions/Edit.php
Senjutsuu: The edit action is meant for editing a formbuilder form and not the field.
That means the dropdown is only added in the edit action.

StijnVrolijk: My bad! Either way, you are doing the same processing of values twice so you should extract it to a single method in the `Autocomplete` class.

src/Backend/Modules/FormBuilder/Engine/Autocomplete.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Backend\Modules\FormBuilder\Engine;
 
+use Backend\Core\Language\Language as BL;
```
StijnVrolijk: Don't alias the `Language` class... I know it's done all over Fork but it's really pretty useless. Just use `Language::lbl` :)

src/Backend/Modules/FormBuilder/Actions/Edit.php
```diff
@@ -97,20 +97,7 @@ private function loadForm(): void
...
-                function (string $value): string {
-                    return $value . ' (' . BL::getLabel('Autocomplete_' . str_replace('-', '_', $value)) . ')';
-                },
-                array_combine(
-                    Autocomplete::POSSIBLE_VALUES,
-                    Autocomplete::POSSIBLE_VALUES
-                )
-            )
-        );
+        $this->form->addDropdown('textbox_autocomplete', Autocomplete::getAutocompleteValues());
```
StijnVrolijk: As far as I can tell, the `getAutocompleteValues()` method isn't implemented yet so you might want to push that to git :) Also, `getAutocompleteValues()` is not a very good name since it's not very specific. `getValuesForDropdown()` might be a better one.

src/Backend/Modules/FormBuilder/Engine/Autocomplete.php
carakas: The alias is because of legacy.

PR #2748, opened 2019-02-05
Prevent mixed content because of livereload on https in debug mode

carakas:
## Type

- Non critical bugfix

## Pull request description

Livereload.js was hardcoded to http, this resulted in mixed content when using https during development


carakas:
Apparently it doesn't work with https … http://livereload.com/limitations/https/

PR #2753, opened 2019-02-06
Get the display name from the profile array

Senjutsuu:
## Type
- Critical bugfix

We get an error when we add a profile in the backend. The error states that the display name given to the breadcrumb is null.
screen shot 2019-02-06 at 09 32 57

## Pull request description
Change the array name to the correct profile array to fix this issue.
screen shot 2019-02-06 at 09 37 55


carakas:
Is there already a test that checks if the profile edit page works? If not I would add it so this doesn't happen again in the future

PR #2757, opened 2019-02-11
Restyle image upload

sarahmey:
## Type
- Enhancement

When uploading an image, it shows super large.

## Pull request description
Restyle image upload for blog and other modules. Image preview is smaller.


src/Backend/Core/Layout/Templates/FormLayout.html.twig
```diff
@@ -214,24 +214,26 @@
 
 {% block fork_image_widget %}
   {% spaceless %}
-    <div class="form-group">
-      {{ form_widget(form.file) }}
-      {% if help_text_message %}
-        <div class="help-text">{{ help_text_message|trans|format(help_text_argument) }}</div>
+    <div class="row">
```
carakas: does it also need to change for symfony form image uploads?

Katrienvh:
I guess the image upload for a page also need the new layout? Can you also add it there?

PR #2758, opened 2019-02-14
Add redirect url to form builder

jonasdekeukelaere:
## Type
- Feature

## Pull request description
Add option to use redirect url instead of success message for form builder forms.


carakas:
Changed to fork 6 because there is a database change

carakas:
Closing this one since we merged #2764 since it is more user friendly

PR #2759, opened 2019-02-14
Use and display a file or image maxSize constraint if the value is lower than the server value

Senjutsuu:
## Type
- Enhancement

When a maxSize constraint is set for a file or image then the default argument text is still displayed as the server max filesize

## Pull request description
Get the value of a maxSize constraint and compare it to the server max filesize setting from the php ini file. If the constraint is higher then the server value we use the server value otherwise we display and use the constraint value.

screen shot 2019-02-14 at 11 17 48

### DIFF
--- a/app/config/form.yml
+++ b/app/config/form.yml
@@ -23,14 +23,12 @@ services:
             - "@translator"
         tags:
             - { name: form.type, alias: meta}
-    form.type.image:
-        class: Common\Form\ImageType
+    Common\Form\ImageType:
         arguments:
             - "@validator"
         tags:
             - { name: form.type, alias: image}
-    form.type.file:
-        class: Common\Form\FileType
+    Common\Form\FileType:
         arguments:
             - "@validator"
         tags:
--- a/app/config/form.yml
+++ b/app/config/form.yml
@@ -29,3 +29,9 @@ services:
             - "@validator"
         tags:
             - { name: form.type, alias: image}
+    form.type.file:
+        class: Common\Form\FileType
+        arguments:
+            - "@validator"
+        tags:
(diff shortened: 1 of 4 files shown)


src/Common/Form/FileType.php
```diff
@@ -173,33 +189,101 @@ function ($optionName) use ($options, &$view) {
...
+        }
+
+        if (is_numeric($constraintValue)) {
+            return $this->prettyPrintFileSize($constraintValue);
+        }
+
+        return $this->prettyPrintFileSize($serverValue);
+    }
+
+    private function prettyPrintFileSize(int $fileSize): string
```
ohvitorino: Can't this function be placed in somewhere else since it's also be being used in ImageType?

app/config/form.yml
```diff
@@ -23,3 +23,15 @@ services:
             - "@translator"
         tags:
             - { name: form.type, alias: meta}
+    form.type.image:
```
carakas: Can you use the new format for this?
```
    Common\Form\ImageType:
        arguments:
            - "@validator"
        tags:
            - { name: form.type, alias: image}
```

carakas: can you use the class name as service name instead please?

app/config/form.yml
```diff
@@ -23,3 +23,15 @@ services:
...
+        class: Common\Form\ImageType
+        arguments:
+            - "@validator"
+        tags:
+            - { name: form.type, alias: image}
+    form.type.file:
```
carakas: idem

carakas:
Thank you, looks good 💯

PR #2780, opened 2019-03-16
Bump local Docker to php 7.4

jessedobbelaere:
## Type

- Enhancement

## Pull request description

Bump the local developent Dockerfile to PHP 7.3
Two issues here:

- `configure: error: Please reinstall the libzip distribution`: Fixed by adding libzip-dev package
-  Composer:`As there is no 'unzip' command installed zip files are being unpacked using the PHP zip extension.`. Fixed by adding the zip and unzip package


carakas:
I would do this in the 6 and keep it at 7.1 for now since that is the minimum version we require. Unless somebody has an argument on why we should already increase it?

carakas:
We'll see what Symfony does with Symfony 5 and follow them

PR #2783, opened 2019-03-27
Improve checking the use of SSL

StijnVrolijk:
## Type
- Enhancement

## Pull request description
Servers that are behind a proxy will always have `%{HTTPS} === off`.
To prevent infinite redirect loops in this scenario this commit adds a
check on the
[X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)
header before defaulting back to the `%{HTTPS}` check. This should cover
about 99% of possible htaccess/ssl related scenarios.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -80,7 +80,8 @@ deny from env=stayout
     # <IfModule headers_module>
     #   Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
     # </IfModule>
-    # RewriteCond %{HTTPS} !=on
+    # RewriteCond %{HTTP:X-Forwarded-Proto} !https
+    # RewriteCond %{HTTPS} off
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -80,7 +80,8 @@ deny from env=stayout
     # <IfModule headers_module>
     #   Header set Strict-Transport-Security "max-age=31536000; includeSubDomains" env=HTTPS
     # </IfModule>
-    # RewriteCond %{HTTPS} !=on
+    # RewriteCond %{HTTP:X-Forwarded-Proto} !https
+    # RewriteCond %{HTTPS} off
```
carakas: why did you change this one as well?
I don't remember the exact reason but I seem to remember there was a reason why we didn't have off but !=on maybe on some servers that is was configured as 0

PR #2788, opened 2019-04-09
Fix bad value for time element

bart-webleads:
## Type
- Non critical bugfix

## Pull request description
validator.W3.org reports this error:
Bad value 2019-02-04CET09,35,00 for attribute datetime on element time: The literal did not satisfy the time-datetime format.
Also removed the milliseconds modifier, second should be fine.

### DIFF
--- a/src/Frontend/Modules/Blog/Layout/Templates/Archive.html.twig
+++ b/src/Frontend/Modules/Blog/Layout/Templates/Archive.html.twig
@@ -41,7 +41,7 @@
             {% for item in items %}
               <tr>
                 <td class="date">
-                  <time itemprop="datePublished" datetime="{{ item.publish_on|date('Y-m-d\\TH:i:s.vP' ) }}">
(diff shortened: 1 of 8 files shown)


carakas:
@bart-webleads can you remove the commit that adds the milliseconds?
It gets through the validator with the milliseconds, unless you have a specific reason why they should be removed cause they were recently added

bart-webleads:
@carakas 

When you add a blog post in the backend, the publish_on field (time) can be entered as "hh:mm" Seconds are not allowed here!

So all datetime stamps for the publish_on column in the blog table look like this:
2018-09-05 07:35:00
2018-09-05 07:37:00
2018-09-05 07:39:00
2016-09-05 07:31:00
2017-09-05 07:31:00
........
Seconds and also milliseconds are not used and are all set to "zero".

So, why should be bother to print milliseconds when these are always set to zero???

carakas:
Oki

PR #2801, opened 2019-04-23
Partly fixed deprecations of Twig #2777

### LINKED ISSUE #2777: Twig deprecations
## Type

> If you discover any security related issues, please email core@fork-cms.com instead of using the issue tracker.

- Enhancement

## Problem description

Since twig 2.7 some classes we extend are deprecated. We need to change them.
## Steps to reproduce

Look in the debug bar for the deprecations

Don't use deprecated twig classes

aalwash:
## Type
- Enhancement

Partly fixes #2777

## Pull request description
I'm not done with all deprecations, some of them are a bit harder, like
`User Deprecated: The "Symfony\Bridge\Twig\Extension\FormExtension" class implements "Twig\Extension\InitRuntimeInterface" that is deprecated since Twig 2.7, to be removed in 3.0.`

### DIFF
--- a/src/Common/Core/Twig/Extensions/TwigFilters.php
+++ b/src/Common/Core/Twig/Extensions/TwigFilters.php
@@ -5,9 +5,10 @@ namespace Common\Core\Twig\Extensions;
 /**
  * Contains all Forkcms filters for Twig
  */
-use Twig_Environment;
-use Twig_SimpleFilter;
-use Twig_SimpleFunction;
+
+use Twig\Environment;
+use \Twig\TwigFilter;
+use \Twig\TwigFunction;
 
 class TwigFilters
 {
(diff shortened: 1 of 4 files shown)


src/Common/Core/Twig/Extensions/TwigFilters.php
```diff
@@ -5,83 +5,84 @@
...
 {
     /**
      * //http://twig.sensiolabs.org/doc/advanced.html#id2
      * returns a collection of Twig SimpleFilters
      *
-     * @param Twig_Environment $twig
+     * @param Environment $twig
      * @param string $app
      */
-    public static function addFilters(Twig_Environment $twig, string $app): void
```
carakas: is this change backwards compatible?

aalwash: It's definitely not backward compatible

But I've checked the usage of TwigFilters, that only happens in the Frontend/Backend Engine
And both are rewritten to pass the new Twig\Environment class

Do extensions call the Frontend/Backend Engine them self?

aalwash: Btw, I forgot to mention that I've done some random filter tests after the Twig changes and it worked for me so far

PR #2816, opened 2019-05-21
Fix youtube and vimeo in the media gallery

### LINKED ISSUE #2791: Youtube in media gallery not working
## Type
- Non critical bug

## Problem description
When adding a Youtube video for the media gallery, it's not working in the frontend.

## Steps to reproduce
Add a media gallery in the backend, add an online movie to it and add Youtube video.
Add the widget from the media gallery to a page.
The YT video won't be shown: 
Screen Shot 2019-04-16 at 07 22 54

Youtube thumbnail shown with link to YT video (in modal or something).

carakas:
## Type

- Non critical bugfix

fixes #2791 

## Pull request description

The media gallery widgets didn't work with youtube and vimeo video's. That has been fixed now.
Also the video's now have thumbnails in the backend.

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Component/StorageProvider/YoutubeStorageProvider.php
+++ b/src/Backend/Modules/MediaLibrary/Component/StorageProvider/YoutubeStorageProvider.php
@@ -10,4 +10,9 @@ class YoutubeStorageProvider extends MovieStorageProvider
     {
         return '<iframe width="560" height="315" src="' . $this->includeUrl . $mediaItem->getUrl() . '" frameborder="0" allowfullscreen></iframe>';
     }
+
+    public function getWebPath(MediaItem $mediaItem): string
+    {
+        return 'https://img.youtube.com/vi/' . $mediaItem->getUrl() . '/maxresdefault.jpg';
+    }
 }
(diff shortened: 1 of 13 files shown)


src/Backend/Modules/MediaLibrary/Component/StorageProvider/YoutubeStorageProvider.php
```diff
@@ -8,7 +8,7 @@ class YoutubeStorageProvider extends MovieStorageProvider
 {
     public function getIncludeHTML(MediaItem $mediaItem): string
     {
-        return '<iframe width="560" height="315" src="' . $this->includeUrl . $mediaItem->getUrl() . '" frameborder="0" allowfullscreen></iframe>';
+        return '<iframe width="100%" height="100%" src="' . $this->includeUrl . $mediaItem->getUrl() . '" frameborder="0" allowfullscreen></iframe>';
```
StijnVrolijk: Does `height="100%"` even work? I'd run this by a frontender

carakas: it worked when I tried it here. It was full in the gallery and a small player when displayed as a single item.
If there is a better way a frontender may always suggest it
@forkcms/frontenders

jeroendesloovere:
The only thing of feedback I can provide is that it should be better if we move all HTML outside of the PHP classes. Perhaps using a translation with %1$s to replace filename, etc...

Example, replacing
```php
return '<iframe src="' . $this->includeUrl . $mediaItem->getUrl() . '?color=ffffff&title=0&byline=0&portrait=0&badge=0" width="100%" height="100%" frameborder="0" webkitallowfullscreen mozallowfullscreen allowfullscreen></iframe>';
``` 

by
```php
return sprintf(
    $this->htmlForInclusion,
    $this->includeUrl . $mediaItem->getUrl()
);
```

carakas:
@jeroendesloovere going to leave it like this. You can always change it in the php if needed. And with the width stuff it will take the size of it's parent so you can control it like that

PR #2819, opened 2019-05-27
Add the current tag to the page title

StijnVrolijk:
## Type

- Enhancement

## Pull request description
This simply adds the current tag to the page title, useful for SEO and the likes.

### DIFF
--- a/src/Frontend/Modules/Tags/Actions/Detail.php
+++ b/src/Frontend/Modules/Tags/Actions/Detail.php
@@ -56,6 +56,7 @@ class Detail extends FrontendBaseBlock
 
     private function parse(): void
     {
+        $this->header->setPageTitle($this->tag['name']);
         $this->template->assign('tag', $this->tag);
         $this->template->assign('tagsModules', $this->tagsModules);
         $this->breadcrumb->addElement($this->tag['name']);


Sellato:
I think it's need to change the order of fields. Looks not so good.

<title>hat - Tags -  Etsy Shop Goods and Gifts</title>

where:
hat - is tag
Tags - is a module name 
Etsy Shop Goods and Gifts - is site Title

Sellato:
> This simply adds the current tag to the page title, useful for SEO and the likes.

This lines kill the SEO :) 
at lines 64-65

>         // tag-pages don't have any SEO-value, so don't index them
        $this->header->addMetaData(['name' => 'robots', 'content' => 'noindex, follow'], true);

StijnVrolijk:
@Sellato hehe, you are correct, I missed that.

I wonder why that decision was made, though? To me it seems pretty useful to have search engines index a page where you can find all sorts of other pages with a certain tag. 

@tijsverkoyen I see you added that line way back in 2011, can you tell me why?

tijsverkoyen:
I can remember Bseen (now part of Wijs) had audited Fork on SEO. So probably feedback from them.

PR #2829, opened 2019-06-12
Fix check to see if the cache should be updated because of strict comparison between 0 and false

carakas:
## Type

- Non critical bugfix

## Pull request description

There was a strict comparison between 0 and false and because of that the code never worked as intended

### DIFF
--- a/src/Backend/Modules/Pages/Actions/Edit.php
+++ b/src/Backend/Modules/Pages/Actions/Edit.php
@@ -770,8 +770,8 @@ class Edit extends BackendBaseActionEdit
         $cacheShouldBeUpdated = !(
             $this->record['title'] === $page['title']
             && $this->record['navigation_title'] === $page['navigation_title']
-            && $this->record['navigation_title_overwrite'] === $page['navigation_title_overwrite']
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Pages/Actions/Edit.php
```diff
@@ -770,8 +770,8 @@ private function validateForm(): void
         $cacheShouldBeUpdated = !(
             $this->record['title'] === $page['title']
             && $this->record['navigation_title'] === $page['navigation_title']
-            && $this->record['navigation_title_overwrite'] === $page['navigation_title_overwrite']
-            && $this->record['hidden'] === $page['hidden']
+            && $this->record['navigation_title_overwrite'] == $page['navigation_title_overwrite'] // can be 0 or false
+            && $this->record['hidden'] == $page['hidden'] // can be 0 or false
```
WouterSioen: If only `$this->record` was an entity with an `->equals` method 🙈

carakas: soon™

PR #2841, opened 2019-06-25
Use Symfony's Dotenv so Symfony Server is more convenient

StijnVrolijk:
## Type

- Enhancement

## Pull request description
We've been using [Symfony server](https://symfony.com/doc/current/setup/symfony_server.html) in favor of [Docker](https://www.docker.com/) for the past few months now as it boasts quite a bit of advantages (not least of all a serious speed improvement).

The main disadvantage here was how Fork handles environments and debugging. In Docker, a virtual machine would be booted containing some environment variables: `FORK_ENV` and `FORK_DEBUG`. Fork would then read these using PHP's `getenv()` method. This poses a problem for Symfony server since it doesn't run a virtual machine but rather a local PHP instance, thus setting environment variables would be very cumbersome and not very dynamic. 

Using [Symfony's Dotenv](https://symfony.com/doc/current/components/dotenv.html) we can set environment variables for each project or even environment or machine individually.

A perfect example might be something like this: By default, Fork ships in production mode without debugging. If you'd like to run in dev mode with debugging on, simply create a `.env.local` file with this content:

```
FORK_ENV=dev
FORK_DEBUG=1
```

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -1,6 +1,9 @@
 <?php
 
+// @TODO rename this file to bootstrap.php to better reflect its function
+
 use Doctrine\Common\Annotations\AnnotationRegistry;
+use Symfony\Component\Dotenv\Dotenv;
 
 // use vendor generated autoloader
 $loader = require __DIR__ . '/vendor/autoload.php';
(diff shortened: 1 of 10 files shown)


autoload.php
```diff
@@ -1,6 +1,9 @@
 <?php
 
+// @TODO rename this file to bootstrap.php to better reflect its function
```
jessedobbelaere: Plan on doing it still in this PR or in a next release?

docs/01. installation/07. docker.md
```diff
@@ -153,12 +153,19 @@ You can easily use different credentials by adjusting them in the `docker-compos
 
 It's easy to setup PHPStorm in combination with Docker and Xdebug, one of the most popular tools to debug your PHP application.
 
-- First copy the .env.dist to .env in the root folder of your project
+- First create a file called .env.local in the root folder of your project and paste these lines in them
 ```
-cp .env.dist .env
+# Use this file to set your environment variables for Docker
+
+# XDebug variables
```
jessedobbelaere: Docker-compose only looks at .env currently (https://docs.docker.com/compose/environment-variables/#the-env-file). 

I guess we can set the .env.local in the `docker-compose.yml`:

```
    env_file:
     - .env.local
```

To make it point to .env.local, and people can move the xdebug variables there.

autoload.php
carakas: next major

docs/01. installation/07. docker.md
WouterSioen: You can even add multiple env files in the docker-compose env_file directive

StijnVrolijk: @jessedobbelaere I hadn't considered people still using Docker for a webserver. I'll add both `.env` and `.env.local`

StijnVrolijk:
@jeroendesloovere I've just implemented environment variables in exactly the same way Symfony does, see https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/config/bootstrap.php 

The code I added to autoload.php is basically the same.

Also, here's an excerpt from https://symfony.com/doc/current/components/dotenv.html:

> Sensitive information and environment-dependent settings should be defined as environment variables (as recommended for twelve-factor applications). Using a .env file to store those environment variables eases development and CI management by keeping them in one "standard" place and agnostic of the technology stack you are using (Nginx vs PHP built-in server for instance).

docs/01. installation/07. docker.md
StijnVrolijk: Unfortunately Docker doesn't like loading environment variables from files that don't exist and `.env.local` doesn't always exist so I created a Docker specific `.env` file in `var/docker`

carakas:
@StijnVrolijk could you update the docs as well? https://docs.fork-cms.com/documentation/detail/installation/docker#phpstorm-configuration

PR #2852, opened 2019-07-01
2830 add page context to symfony debug bar

### LINKED ISSUE #2830: Page context - which action and/or widget(s) are being rendered on the current page
## Type

- Feature

## Problem description

Many times while working on existing projects which I'm not yet familiar with, it would be nice to be able to know which actions and/or widgets are being rendered in the frontend on the page I'm currently visiting.

## Proposed solutions

An extra item could be added to the Symfony debug toolbar with the name of the action and/or widgets being rendered on the page and the also the module name, similarly to the Request/Response button of the current toolbar.

image

carakas:
## Type

- Feature

closes #2830 
## Pull request description

This will add a section to the Symfony Debug Toolbar and give you some extra information about the page you are watching.

#### When there is no data (Backend)
Screenshot 2019-07-01 at 18 32 01

#### When there is both a block and a widget
Screenshot 2019-07-01 at 18 31 31

Screenshot 2019-07-01 at 18 30 41

#### When there is both a no block but there is a widget
Screenshot 2019-07-01 at 18 31 14
Screenshot 2019-07-01 at 18 30 53


ohvitorino:
@carakas a counter on the button with the amount of widgets would also be nice :)

WouterSioen:
But do you also contain the main block into account then? It would probably be weird to have `0` as count on a page with only a module and no widgets?

An aggregated count of all blocks on the page might be nice though 👌

carakas:
Screenshot 2019-07-02 at 08 45 05
I implemented it now like this.
No visible count if there are no widgets. Just like the block info is also not shown if it is null.

carakas:
@WouterSioen do you think it is oke like that? cause I'm not sure if also counting things like the html from the editors in the backend adds value

WouterSioen:
It's a pretty good improvement already. I think you'll get a feeling real quick if that would provide value once you start using this part of the debug bar.

PR #2858, opened 2019-07-09
Show session timeout warning 15min after page load

Katrienvh:
## Type
- Enhancement

Losing changes because your session expired without any warning.

## Pull request description
Now there will appear a warning 15 minutes after page load, with the message to recommend the user to (temporary) save the changes that are made.

### DIFF
--- a/src/Backend/Core/Engine/Header.php
+++ b/src/Backend/Core/Engine/Header.php
@@ -221,6 +221,7 @@ final class Header extends KernelLoader
 
         $this->jsData->add('site', 'domain', SITE_DOMAIN);
         $this->jsData->add('editor', 'language', $this->getCKEditorLanguage());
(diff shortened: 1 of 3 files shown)


carakas:
@Katrienvh is session timeout on all servers the same? Wouldn't it be better to look at the server config and then subtract like 5 minutes or so?

StijnVrolijk:
What if the "save" action uses an ajax call? Wouldn't the message have to be shown again after x minutes?

Katrienvh:
> What if the "save" action uses an ajax call? Wouldn't the message have to be shown again after x minutes?

Katrienvh:
> What if the "save" action uses an ajax call? Wouldn't the message have to be shown again after x minutes?
@StijnVrolijk fixed in 9e03fd6

PR #2871, opened 2019-08-03
Fix broken translation in Pagination template

jessedobbelaere:
## Type

- Non critical bugfix

## Pull request description

The translation in the base pagination template was broken and was showing literal string "lbl.NextPage". This should fix it


carakas:
Thanks

PR #2874, opened 2019-08-07
Convert + to %20 in mailto urls

StijnVrolijk:
## Type
- Non critical bugfix

## Pull request description
Previously, if you created a mailto link with spaces in the subject and/or body, the url would contain "+" instead of spaces (because of `url.toString()`) and those "+"'s would end up in the mail clients.

This pull request converts those "+"'s to "%20" which get converted to actual spaces by the mail clients.

### DIFF
--- a/src/Backend/Core/Js/ckeditor/plugins/medialibrary/dialogs/linkDialog.js
+++ b/src/Backend/Core/Js/ckeditor/plugins/medialibrary/dialogs/linkDialog.js
@@ -208,8 +208,9 @@ CKEDITOR.dialog.add(
                                     url.searchParams.append('body', this.getDialog().getValueOf('tab', 'content'));
                                 }
 
-                                element.setAttribute('href', url.toString());
-                                element.setAttribute('data-cke-saved-href', url.toString());
+                                var href = url.toString().replace(/\+/g, '%20');
+                                element.setAttribute('href', href);
+                                element.setAttribute('data-cke-saved-href', href);
                             }
                         },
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Js/ckeditor/plugins/medialibrary/dialogs/linkDialog.js
```diff
@@ -208,8 +208,9 @@ CKEDITOR.dialog.add(
                                     url.searchParams.append('body', this.getDialog().getValueOf('tab', 'content'));
                                 }
 
-                                element.setAttribute('href', url.toString());
-                                element.setAttribute('data-cke-saved-href', url.toString());
+                                var href = url.toString().replace(/\+/g, '%20');
```
WouterSioen: Can't this have problems when e.g. query parameters contain a plus? Not sure what's exactly possible here but this seems quite risky 🤔

StijnVrolijk:
```
var url = new URL('https://www.fork-cms.com')
url.searchParams.append('foo', 'bar + bar = 20')
console.log(url.toString()) // https://www.fork-cms.com/?foo=bar+%2B+bar+%3D+20
console.log(url.toString().replace(/\+/g, '%20')) // https://www.fork-cms.com/?foo=bar%20%2B%20bar%20%3D%2020
```

So it shouldn't cause any issues

PR #2901, opened 2019-09-16
Add trim to user content template's text value

jessedobbelaere:
## Type

- Enhancement

When my user content template contains this (auto-formatted) html:

```
<h5 class="text-blue-darkest font-medium uppercase text-lg" data-ft-label="Naam" data-ft-type="text">
    Firstname Lastname
</h5>
```

There are spaces in the html. However, these spaces are copied in the rendering of the field:

image

## Pull request description

- ✅ Add a `trim()` to the value of the user content text field html method
- ✅ Use a ES6 [Template Literals](https://caniuse.com/#feat=template-literals) to return the snippet of HTML. We lose IE11 support by doing that, but I guess we should stop supporting IE11 in the backend? ⚠️   This is the backend and any admin is most likely using an evergreen browser? Should I make all the methods that return html in this class, use template literals? 

After the fix: no added whitespace in the default values visible 

image

### DIFF
--- a/src/Backend/Modules/Pages/Js/Pages.js
+++ b/src/Backend/Modules/Pages/Js/Pages.js
@@ -744,22 +744,27 @@ jsBackend.pages.extras = {
    * Creates the html for a text field
    */
   getTextFieldHtml: function (text, label, key) {
-    return `
-      <div class="panel panel-default" id="user-template-text-${key}">
-        <div class="panel-heading">
-          <h2 class="panel-title">${label}</h2>
-        </div>
-        <div class="panel-body">
-          <div class="form-group last">
-            <input data-ft-label="${label}" type="text" class="form-control" value="${text.trim()}" />
-          </div>
-        </div>
-      </div>
-    `
+    var html = '<div class="panel panel-default" id="user-template-text-' + key + '">'
+
+    html += '<div class="panel-heading">'
+    html += '<h2 class="panel-title">' + label + '</h2>'
+    html += '</div>'
+
+    html += '<div class="panel-body">'
+
+    html += '<div class="form-group last">'
+    html += '<input data-ft-label="' + label + '" type="text" class="form-control" value="' + text.trim() + '" />'
+    html += '</div>'
+
+    html += '</div>'
+
+    html += '</div>'
+
+    return html
   },
 
   /**
-   * Creates the html for an editor
+   * Creates the html for a text area
    */
   getTextAreaFieldHtml: function (text, label, key) {
     var html = '<div class="panel panel-default panel-editor" id="user-template-textarea-' + key + '">'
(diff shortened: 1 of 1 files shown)


jessedobbelaere:
Possibly also needed for the textarea field?

src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -744,23 +744,18 @@ jsBackend.pages.extras = {
...
-    html += '<div class="form-group last">'
-    html += '<input data-ft-label="' + label + '" type="text" class="form-control" value="' + text + '" />'
-    html += '</div>'
-
-    html += '</div>'
-
-    html += '</div>'
-
-    return html
+    return `
```
carakas: is this supported in all browsers?

carakas: It's this right? https://caniuse.com/#feat=template-literals

jessedobbelaere: Yeah indeed

> ✅ Use a ES6 Template Literals to return the snippet of HTML. We lose IE11 support by doing that, but I guess we should stop supporting IE11 in the backend? ⚠️ This is the backend and any admin is most likely using an evergreen browser? Should I make all the methods that return html in this class, use template literals?


I wonder if we should allow more modern Javascript, knowing that this is the backend of the CMS, and should we still support IE11?

mathiashelin: Yes, we should support IE11.
I understand that you want to use modern JS, but this is too small of an advantage to justify dropping IE11 support.

Katrienvh: I share Mathias opinion, i think we should not drop the support for IE11

ohvitorino: This PR is a good solution for an annoying problem but dropping the support for IE11 seems to me a little bit too early.

jessedobbelaere: Ok, sounds good. Yeah, this MR is a pretty small change, and I wasn't aware that IE11 support is still needed in the backend of the CMS, so I wanted to discuss it here. Some CMS still support IE11 (enterprise-focused ones and wordpress), some more modern CMS systems drop their support. Like [Craft CMS requirements](https://docs.craftcms.com/v3/requirements.html#control-panel-browser-requirements) or [Shopify requirements](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers). But I understand that we don't want to bundle/transpile the core js code + each module's javascript file either, to make it easier for devs, so we have to keep our javascript like it's 2008 😁. 

Anyway, I'll make some changes to just use trim without ES2015 features :-)

PR #2903, opened 2019-09-18
Media Helper JS triggered twice

cirykpopeye:
## Type
- Non critical bugfix

## Pull request description
When opening the model the fineuploader gets triggered once again, but it's already been trigger the moment the page was loaded. This causes two dropzone's and when dragging in images it uploads each image twice or more depending on how many times you trigger the modal.


carakas:
Nice catch

PR #2909, opened 2019-09-24
Secure cookies

carakas:
## Type

- Security

## Pull request description

Secure all the cookies when using https

### DIFF
--- a/src/Backend/Core/Js/utils.js
+++ b/src/Backend/Core/Js/utils.js
@@ -75,7 +75,7 @@ utils.cookies = {
 
     var expireDate = new Date()
     expireDate.setDate(expireDate.getDate() + days)
-    document.cookie = name + '=' + escape(value) + ';expires=' + expireDate.toUTCString() + ';path=/;secure=' + secure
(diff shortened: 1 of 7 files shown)


src/Backend/Core/Js/utils.js
```diff
@@ -69,12 +69,13 @@ utils.cookies = {
...
 
-  setCookie: function (name, value, days) {
+  setCookie: function (name, value, days, secure) {
     if (typeof days === 'undefined') days = 7
+    if (typeof secure === 'undefined') secure = location.protocol === 'https:'
 
     var expireDate = new Date()
     expireDate.setDate(expireDate.getDate() + days)
-    document.cookie = name + '=' + escape(value) + ';expires=' + expireDate.toUTCString() + ';path=/'
+    document.cookie = name + '=' + escape(value) + ';expires=' + expireDate.toUTCString() + ';path=/;secure=' + secure
```
jessedobbelaere: On [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#Security) and other examples I never see a boolean value being used, e.g. `;secure=true`. Is that supported?

If I run this snippet in my browser console:

```
document.cookie = "user=John; secure=false"
```

then the secure flag is toggled anyways? 
image

PR #2913, opened 2019-10-02
More secure headers

carakas:
## Type

- Enhancement

## Pull request description

Add some improvements to the default headers to make things more secure etc

### DIFF
--- a/src/Common/EventListener/ResponseSecurer.php
+++ b/src/Common/EventListener/ResponseSecurer.php
@@ -14,13 +14,19 @@ class ResponseSecurer
      */
     public function onKernelResponse(FilterResponseEvent $event): void
     {
-        $cacheControl = 'max-age=0, must-revalidate, private, no-cache';
+        $cacheControl = [
+            'max-age=0',
+            'must-revalidate',
+            'private',
+            'no-cache',
+        ];
+
         if (defined('APPLICATION') && APPLICATION === 'Backend') {
-            $cacheControl .= ', no-store';
+            $cacheControl[] = 'no-store';
         }
 
         $headers = [
-            'Cache-Control' => $cacheControl,
+            'Cache-Control' => implode(', ', $cacheControl),
             'X-Frame-Options' => 'deny',
             'Expires' =>  '0',
             'Pragma' => 'no-cache',
--- a/src/Common/EventListener/ResponseSecurer.php
+++ b/src/Common/EventListener/ResponseSecurer.php
@@ -25,6 +25,7 @@ class ResponseSecurer
             'Expires' =>  '0',
             'Pragma' => 'no-cache',
             'referrer' => 'strict-origin-when-cross-origin',
+            'Referrer-Policy' => 'strict-origin-when-cross-origin',
             'X-XSS-Protection' => '1; mode=block',
             'X-Content-Type-Options' => 'nosniff',
         ];
--- a/src/Common/EventListener/ResponseSecurer.php
+++ b/src/Common/EventListener/ResponseSecurer.php
@@ -21,9 +21,22 @@ class ResponseSecurer
             'X-Content-Type-Options' => 'nosniff',
         ];
 
+
+        $responseHeaders = $event->getResponse()->headers;
         foreach ($headers as $header => $value) {
-            if (!$event->getResponse()->headers->has($header)) {
-                $event->getResponse()->headers->set($header, $value);
+            if (!$responseHeaders->has($header)) {
+                $responseHeaders->set($header, $value);
+            }
+        }
+
+        // Don't leak server config
+        $blockedHeaders = [
+            'x-powered-by',
+            'Server',
+        ];
+        foreach ($blockedHeaders as $blockedHeader) {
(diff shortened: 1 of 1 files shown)


src/Common/EventListener/ResponseSecurer.php
```diff
@@ -14,16 +14,38 @@ class ResponseSecurer
...
-            if (!$event->getResponse()->headers->has($header)) {
-                $event->getResponse()->headers->set($header, $value);
+            if (!$responseHeaders->has($header)) {
+                $responseHeaders->set($header, $value);
+            }
+        }
+
+        // Don't leak server config
+        $blockedHeaders = [
+            'x-powered-by',
```
jessedobbelaere: Q: Do we need to check in a case-insensitive way? 
I think this header is often used in Nginx with `X-Powered-By` (camelcased version) next to the lowercase version.

src/Common/EventListener/ResponseSecurer.php
```diff
@@ -14,16 +14,38 @@ class ResponseSecurer
...
+            'Server',
+        ];
+        foreach ($blockedHeaders as $blockedHeader) {
+            if ($responseHeaders->has($blockedHeader)) {
+                $responseHeaders->remove($blockedHeader);
```
jessedobbelaere: Q: Do we need to check in a case-insensitive way? 
I think this header is often used in Nginx with `X-Powered-By` (camelcased version) next to the lowercase version.

carakas: A: the keys are lowercased

```php
    /**
     * Returns true if the HTTP header is defined.
     *
     * @param string $key The HTTP header
     *
     * @return bool true if the parameter exists, false otherwise
     */
    public function has($key)
    {
        return \array_key_exists(str_replace('_', '-', strtolower($key)), $this->all());
    }
```
and

```php
/**
     * {@inheritdoc}
     */
    public function remove($key)
    {
        $uniqueKey = str_replace('_', '-', strtolower($key));
        unset($this->headerNames[$uniqueKey]);

        if ('set-cookie' === $uniqueKey) {
            $this->cookies = [];

            return;
        }

        parent::remove($key);

        if ('cache-control' === $uniqueKey) {
            $this->computedCacheControl = [];
        }

        if ('date' === $uniqueKey) {
            $this->initDate();
        }
    }
```

src/Common/EventListener/ResponseSecurer.php
```diff
@@ -14,16 +14,38 @@ class ResponseSecurer
      */
     public function onKernelResponse(FilterResponseEvent $event): void
     {
+        $cacheControl = 'max-age=0, must-revalidate, private, no-cache';
```
tijsverkoyen: I prefer
```$cacheControl = [
'max-age=0',
'must-revalidate', 
'private', 
'no-cache',
];```, and where line 19, adds it to the array, and on line 23, it can be imploded

File: src/Common/EventListener/ResponseSecurer.php [hunk 4]
```diff
@@ -14,16 +14,38 @@ class ResponseSecurer
...
+        if (defined('APPLICATION') && APPLICATION === 'Backend') {
+            $cacheControl .= ', no-store';
+        }
+
         $headers = [
+            'Cache-Control' => $cacheControl,
             'X-Frame-Options' => 'deny',
+            'Expires' =>  '0',
+            'Pragma' => 'no-cache',
```
tijsverkoyen: is this needed, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma
> Use Pragma only for backwards compatibility with HTTP/1.0 clients.

carakas: I added it here because it came back in an audit done on a fork project

src/Common/EventListener/ResponseSecurer.php
carakas: changed in df776be9d

src/Common/EventListener/ResponseSecurer.php
tijsverkoyen: ok.

wouterverstuyf:
When I doe a reinstall I get the following error:
Notice: Use of undefined constant APPLICATION - assumed 'APPLICATION' in /users/wouter/Sites/forkcms/src/Backend/Core/Language/Language.php on line 246

I suppose this is related to this PR?

Schermafbeelding 2019-10-03 om 11 43 30

carakas:
no, it has to do with a fix from in the last release, but I'll fix it

src/Common/EventListener/ResponseSecurer.php
jessedobbelaere: thx, didn't know!

wouterverstuyf:
> @wouterverstuyf what fork version?

Latest - 5.6.2

carakas:
@wouterverstuyf it wasn't 5.7.2 bu the master branch, you caught a bug before release, thank you, I have no clue why I flipped it around
https://github.com/forkcms/forkcms/commit/3dce90d8cc59f8225931359cd9e392602b7ef6a4

PR #2928, opened 2019-10-10
Fix env variables loading in test mode

ohvitorino:
## Type

- Non critical bugfix

## Pull request description

This PR sets the values of `FORK_ENV` and `FORK_DEBUG` in the PHPUnit configuration.
The values of `$_SERVER['APP_ENV]` and `$_SERVER['APP_DEBUG]` are also set to the same value as `$_SERVER['FORK_ENV]` and `$_SERVER['FORK_DEBUG]` respectively. The Symfony kernel makes some initialisation decisions based on these variables.

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -32,8 +32,8 @@ if ('prod' !== $_SERVER['FORK_ENV']) {
     }
 }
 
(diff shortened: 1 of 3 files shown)


ohvitorino:
I just noticed loading `.env.test` is not correct. I'll have a look at this again, because it's not correct.

PR #2930, opened 2019-10-14
Make adding a timestamp in the php work

Katrienvh:
## Type
- Non critical bugfix

If the addTimestamp option was set true with the addJS function i wasn't working because of the way we rendered the script tags in the template


carakas:
can you fix it for the css as well

PR #2931, opened 2019-10-14
Fix session timeout popup

jonasdekeukelaere:
## Type
- Non critical bugfix

## Pull request description
In https://github.com/forkcms/forkcms/pull/2858 the session timeout popup was added, but the `session_timout` variable is saved in the jsData (see https://github.com/forkcms/forkcms/pull/2858/files#diff-4fb0e3a8ed70071edbb74cdf222a7af6R224) not in `jsBackend`.


carakas:
nice catch :)

PR #2938, opened 2019-10-17
Allow Core actions for frontend ajax calls

jonasdekeukelaere:
## Type
- Non critical bugfix

## Pull request description
When adding frontend ajax calls, the AjaxAction allows for Core ajax actions (see https://github.com/forkcms/forkcms/blob/master/src/Frontend/Core/Engine/AjaxAction.php#L32), but the frontend ajax handler doesn't allow it, this fixes the problem.

### DIFF
--- a/src/Frontend/Core/Engine/Ajax.php
+++ b/src/Frontend/Core/Engine/Ajax.php
@@ -122,6 +122,10 @@ class Ajax extends KernelLoader implements ApplicationInterface
     public function setAction(string $action): void
     {
         $ajaxActionClass = 'Frontend\\Modules\\' . $this->getModule() . '\\Ajax\\' . $action;
+        if ($this->getModule() === 'Core') {
+            $ajaxActionClass = 'Frontend\\Core\\Ajax\\' . $action;
+        }
+
         if (!class_exists($ajaxActionClass)) {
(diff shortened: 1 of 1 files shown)


carakas:
Fork doesn't have any core ajax actions in the frontend as far as I can see, so I don't see how this is a bug

jonasdekeukelaere:
But if you want to add one, some code allows for it, but other parts of it not.

jonasdekeukelaere:
You'll get the following error: `{"code":500,"data":null,"message":"Action class Frontend\\Modules\\Core\\Ajax\\SaveRating does not exist"}`

carakas:
you shouldn't be adding it there but to a module imho

jonasdekeukelaere:
I've used it for multiple ajax call that was used in different modules and the domain for it was placed in Common namespace.

StijnVrolijk:
There's a bigger issue here, I believe.

I happen to know what code lead you to this problem and the issue is you added code in the `Common` namespaces that's common between modules _in_ the Frontend.

Imo, `Common` is a namespace for stuff that's both used in Front- and Backend. If a bunch of code is a dependency of several *Frontend* modules, it should be placed in a separate module _in_ the frontend.

PR #2953, opened 2019-10-22
Add language to url

ohvitorino:
## Type

- Enhancement

## Pull request description

Add language to the API call parameters and change resource name

`/api/content-blocks/{language}/{id}`
`/api/content-blocks`

### DIFF
--- a/src/Backend/Modules/ContentBlocks/Api/ApiController.php
+++ b/src/Backend/Modules/ContentBlocks/Api/ApiController.php
@@ -38,12 +38,8 @@ final class ApiController
 
     /**
      * @Rest\Get("/content-blocks/{language}/{id}")
-     *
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/ContentBlocks/Api/ApiController.php
```diff
@@ -22,6 +23,9 @@ public function __construct(ContentBlockRepository $blockRepository, SerializerI
         $this->serializer = $serializer;
     }
 
+    /**
+     * @Rest\Get("/content-blocks")
```
carakas: I would add the language here as well

ohvitorino: Good point, added 308152d8

PR #2959, opened 2019-10-25
Formbuilder confirmation mail message

wouterverstuyf:
## Type
- Enhancement

## Pull request description
There was already a field for the subject of the formbuilder confirmation mail (to the visitor). Perhaps it is useful to also make the message customizable.

Schermafbeelding 2019-10-25 om 21 19 00


wouterverstuyf:
Nice extension possible:
Would be great if we can use the fields as a token in the confirmation mail :)

wouterverstuyf:
Update - Completely rebuild the mail flow in #2961

PR #2960, opened 2019-10-26
Bugfix: Make medialibrary modal scrollable

jessedobbelaere:
## Type

- Non critical bugfix

## Pull request description

When you create a media gallery and upload a bunch of files, the modal becomes bigger than the screen and does not allow for scrolling towards the Save button.

### Before
before

### After
after


carakas:
Love it, just checking with some frontenders if this is the correct way to fix this bug

PR #2961, opened 2019-10-26
Formbuilder multiple emails

wouterverstuyf:
## Type
- Feature

## Pull request description
In the current formbuilder module you can send 1 e-mail to the value of a certain field, in addition to the confirmation e-mail to the site owner.
With this extension it is possible to set multiple e-mails per form, choose destinations (multiple addresses or value of a field), choose to add the form data and also specify the message itself.
Tested but errors are possible, as well as possible improvements to code.

Schermafbeelding 2019-10-26 om 21 50 24

### DIFF
--- a/src/Backend/Modules/FormBuilder/Actions/Add.php
+++ b/src/Backend/Modules/FormBuilder/Actions/Add.php
@@ -36,17 +36,7 @@ class Add extends BackendBaseActionAdd
     {
         $this->form = new BackendForm('add');
         $this->form->addText('name')->makeRequired();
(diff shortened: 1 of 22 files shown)


carakas:
I like the idea, I'll have a look at the code later.
It does contain a breaking change so it will be for the next major release

wouterverstuyf:
@carakas 
Can this be merged if it's still interesting enough? Or only in Fork 6?

carakas:
Only in Fork 6, because of a BC break

PR #2965, opened 2019-11-02
Add select-all checkbox to MediaGallery connect datagrid

jessedobbelaere:
## Type

- Enhancement

In the current MediaGallery module you need to browse through your media and "connect" images. This looks like this:
Screen Shot 2019-11-02 at 17 39 12

However, if you have 100+ images to add to a gallery, it's quite annoying to select them all manually, as there is no select-all checkbox. 

## Pull request description

This PR adds a check-all checkbox to the datagrid of the MediaGallery modal to connect media to a gallery. It looks just like any other datagrid used in Fork CMS

image

### Changes done

* Added `.change()` to the bindMassCheckbox event handler in `backend.js` because programmatically checking checkboxes does not trigger any events (change event). Without triggering events, the medialibrary modal would not actually add any images even though they're selected. It needs the event to add them to the `currentMediaItemIds` array
* Added a `<thead>` in `MediaLibraryHelper.js` with the checkbox to select all checkboxes. 
* Changed the event handler on checkboxes from `click` to `change` which makes more sense for checkboxes. Added an `isChecked` check to the conditionals that populate the `currentMediaItemIds`

### DIFF
--- a/src/Backend/Core/Js/backend.js
+++ b/src/Backend/Core/Js/backend.js
@@ -874,7 +874,7 @@ jsBackend.controls = {
       var $this = $(this)
 
       // check or uncheck all the checkboxes in this datagrid
-      $this.closest('table').find('td input:checkbox').prop('checked', $this.is(':checked'))
+      $this.closest('table').find('td input:checkbox').prop('checked', $this.is(':checked')).change()
 
       // set selected class
       if ($this.is(':checked')) {
--- a/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
+++ b/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
@@ -613,8 +613,11 @@ jsBackend.mediaLibraryHelper.group = {
     })
 
     $(mediaItemTypes).each(function (index, type) {
-      var mediaTableHtml = '<thead><tr><th class="check"><span><input type="checkbox" name="toggleChecks" value="toggleChecks"></span></th>' +
-        '<th>&nbsp;</th><th>&nbsp;</th><th>&nbsp;</th></tr></thead>' +
+      var mediaTableHtml = '<thead><tr><th class="check"><span><input type="checkbox" name="toggleChecks" value="toggleChecks" title="Select all"></span></th>' +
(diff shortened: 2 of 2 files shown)


ohvitorino:
I would replace the checkbox with a link label `select/deselect all` to improve usability. I know the checkbox is self explanatory for the most of us, but a link with text is more explicit.

carakas:
> I would replace the checkbox with a link label `select/deselect all` to improve usability. I know the checkbox is self explanatory for the most of us, but a link with text is more explicit.

@jessedobbelaere what do you think about this?

jessedobbelaere:
To me it does not matter :-) . 
I'll have a look to change it to a text link soon 👍

ohvitorino:
> To me it does not matter :-) .
> I'll have a look to change it to a text link soon 👍 Should we put it right above the datagrid?

I would replace the checkbox 🙂

carakas:
@jessedobbelaere can you check the feedback?

jessedobbelaere:
Yep! Lost track of this. Not sure if I understand it right. Would you place a text label "select/deselect all" at the top or bottom of the datagrid? Or replace the checkbox in the table header `<thead>` of the datagrid with a link label?

jessedobbelaere:
What do you think about the placement of the `select/deselect all` link @ohvitorino or @carakas ? 👀

carakas:
@Katrienvh ?

Katrienvh:
Me and @sarahmey discussed this and we both agree that there should be a better way to add a select all checkbox/link, but as we can see the way Jesse added the checkbox now is the default in fork (see screenshot). This is also a common used way of doing it in a lot of web apps (people recognize it). For now we should leave it like this. There should be a title attribute with value 'select all'. Also adding the titles of the collum (preview, image name, ...) in de top row of the datagrid.

image

jessedobbelaere:
Thx @Katrienvh ! 
I added (locale) titles to the column headers + added title attribute on the checkbox.
image

Preview: https://share.jessedobbelae.re/gp4kn1

_Note: I had to add a conditional on `type === "image"` because the images datagrid had a column extra (and needed a extra column header)_

PR #2973, opened 2019-11-12
Allow get parameters to be set when creating an action for the DeleteType

StijnVrolijk:
## Type

- Feature

## Pull request description

While I was working on a feature in a project, the delete action my DeleteType was POSTing to didn't know what to redirect to after performing the delete but the action where I created the DeleteType did! By allowing the GET parameters of the action in the DeleteType to be set I was able to send a `queryString` parameter along to the Delete action. This allowed me to have the Edit action tell the Delete action where to redirect to, useful stuff!

This could be used for a multitude of other things as well.

### DIFF
--- a/src/Backend/Form/Type/DeleteType.php
+++ b/src/Backend/Form/Type/DeleteType.php
@@ -12,7 +12,14 @@ class DeleteType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options): void
     {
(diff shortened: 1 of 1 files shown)


ohvitorino:
@StijnVrolijk any chance we can document this anywhere? `get_parameters` doesn't really give much context to what it does.

carakas:
@ohvitorino fixed in a8b1a81

PR #2986, opened 2019-11-18
Small fixes

carakas:
## Type

- Non critical bugfix

## Pull request description

A combination of small fixes from changes that were newly merged

### DIFF
--- a/src/Backend/Core/Installer/Data/locale.xml
+++ b/src/Backend/Core/Installer/Data/locale.xml
@@ -7817,7 +7817,7 @@
       </item>
       <item type="label" name="PageTitle">
         <translation language="nl"><![CDATA[paginatitel]]></translation>
-        <translation language="en"><![CDATA[pagetitle]]></translation>
+        <translation language="en"><![CDATA[page title]]></translation>
         <translation language="hu"><![CDATA[oldalcím]]></translation>
         <translation language="ru"><![CDATA[Заголовок]]></translation>
         <translation language="it"><![CDATA[titolo di pagina]]></translation>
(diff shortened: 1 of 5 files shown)


src/Backend/Core/Installer/Data/locale.xml
```diff
@@ -7817,7 +7817,7 @@
       </item>
       <item type="label" name="PageTitle">
         <translation language="nl"><![CDATA[paginatitel]]></translation>
-        <translation language="en"><![CDATA[pagetitle]]></translation>
+        <translation language="en"><![CDATA[page title]]></translation>
```
WouterSioen: Shouldn't the Dutch translation also be split up?

src/Common/Doctrine/Type/LocaleType.php
```diff
@@ -18,13 +18,17 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
     }
 
     /**
-     * @param string $locale
+     * @param string|null $locale
```
WouterSioen: Can't you type hint this inside the method signature?

src/Backend/Core/Installer/Data/locale.xml
carakas: Nope, it is apparently one word in Dutch

src/Common/Doctrine/Type/LocaleType.php
carakas: Not yet, only once php 7.3 is the minimum version of fork since the declaration wouldn't be the same as the base class.

PR #2997, opened 2019-12-04
Fix cropping for images with height lower than requested folder height

dreadnip:
## Type

- Non critical bugfix

## Pull request description

This commit fixes thumbnail generation for images that have a base height lower than what is requested for that folder.

Before:
oeps

After:
beter


ohvitorino:
@dreadnip can you add a pull request description? Thx

PR #3004, opened 2019-12-09
Add documentation about the SingleMediaGroupType

carakas:
## Type

- Documentation

## Pull request description

Add some documentation for the MediaLibrary about the usage of the SingleMediaGroupType

### DIFF
--- a/docs/07. media library/03. integrating in your module.md
+++ b/docs/07. media library/03. integrating in your module.md
@@ -51,14 +51,6 @@ Symfony FormType class:
 use Backend\Modules\MediaLibrary\Domain\MediaGroup\MediaGroupType;
 ```
 
-Or if you only want a single media item to be connected you can use the SingleMediaGroupType
-```php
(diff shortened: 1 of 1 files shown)


docs/07. media library/03. integrating in your module.md
```diff
@@ -51,6 +51,15 @@ Symfony FormType class:
...
 
+Or if you only want a single media item to be connected you can use the SingleMediaGroupType
+```php
+use Backend\Modules\MediaLibrary\Domain\MediaGroup\SingleMediaGroupType;
+```
+
+This will set the options minimum_items and maximum_items to 1 and add the constraint to enforce it.
+These options are also available on the MediaGroupType if you would want a different range.
+To help the end user there is also some client side validation that will indicate when more items need to be connected or too many items are connected.
+
```
ohvitorino: The description looks fine but the example below doesn't match the form type you just described. I would put this text under the example and add an example of the SingleMediaGroupType.

carakas: fixed in a02f2e551

PR #3006, opened 2019-12-17
Fix timestamps for caching incorrect

carakas:
## Type

- Enhancement
- Deprecation

## Pull request description

Use the filetime of the actual file instead of that of the parameters.yml file for caching assets
That way you don't need to touch the parameters.yml file to make sure your assets get a new url

### DIFF
--- a/src/Common/Core/Header/Asset.php
+++ b/src/Common/Core/Header/Asset.php
@@ -64,6 +64,6 @@ final class Asset
         // check if we need to use a ? or &
         $separator = mb_strpos($this->file, '?') === false ? '?' : '&';
 
-        return $this->file . $separator . 'm=' . filemtime(__DIR__ . '/../../../../' . $this->file);
+        return $this->file . $separator . 'm=' . @filemtime(__DIR__ . '/../../../../' . $this->file);
     }
 }
--- a/src/Common/Core/Header/Asset.php
+++ b/src/Common/Core/Header/Asset.php
@@ -26,7 +26,7 @@ final class Asset
         $this->createdOn = new DateTimeImmutable();
     }
 
-    public function compare(Asset $asset)
+    public function compare(Asset $asset): int
     {
         $comparison = $this->priority->compare($asset->priority);
 
@@ -62,6 +62,8 @@ final class Asset
         }
 
         // check if we need to use a ? or &
-        return $this->file . (mb_strpos($this->file, '?') === false ? '?' : '&') . 'm=' . LAST_MODIFIED_TIME;
+        $separator = mb_strpos($this->file, '?') === false ? '?' : '&';
+
+        return $this->file . $separator . 'm=' . filemtime(__DIR__ . '/../../../../' . $this->file);
     }
 }
--- a/src/Backend/Core/Engine/Form.php
+++ b/src/Backend/Core/Engine/Form.php
@@ -82,16 +82,12 @@ class Form extends \Common\Core\Form
 
         // add the internal link lists-file
         if (is_file(FRONTEND_CACHE_PATH . '/Navigation/editor_link_list_' . BackendLanguage::getWorkingLanguage() . '.js')) {
-            $timestamp = @filemtime(
-                FRONTEND_CACHE_PATH . '/Navigation/editor_link_list_' . BackendLanguage::getWorkingLanguage() . '.js'
-            );
             $this->header->addJS(
-                '/src/Frontend/Cache/Navigation/editor_link_list_' . BackendLanguage::getWorkingLanguage(
-                ) . '.js?m=' . $timestamp,
+                '/src/Frontend/Cache/Navigation/editor_link_list_' . BackendLanguage::getWorkingLanguage() . '.js',
                 null,
                 false,
                 true,
-                false
+                true
             );
         }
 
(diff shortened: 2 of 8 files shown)


src/Common/Core/Header/Asset.php
```diff
@@ -62,6 +62,8 @@ public function __toString(): string
         }
 
         // check if we need to use a ? or &
-        return $this->file . (mb_strpos($this->file, '?') === false ? '?' : '&') . 'm=' . LAST_MODIFIED_TIME;
+        $separator = mb_strpos($this->file, '?') === false ? '?' : '&';
+
+        return $this->file . $separator . 'm=' . filemtime(__DIR__ . '/../../../../' . $this->file);
```
tijsverkoyen: I don't know if this is a correct solution:
1. [The results of this function are cached. See clearstatcache() for more details](https://www.php.net/filemtime)
2. In order to have a reliable modified time you need to execute `clearstatcache` before, only once per request.

The reason why we have chosen the modified time from parameters.yml (or globals.php in much older versions) was because we could touch a single file in deploy-scripts to change the version on all assets.

carakas: @tijsverkoyen I looked at that function because I initially thought the same.
I actually tested it locally and it works.
When you look deeper in the clearstatcache function it says that it is withing the same request.
It is there in case a file can be changed during the request itself. And in case of the assets that won't happen

tijsverkoyen: I don't get it. Again: In order to have a reliable modified time you need to execute clearstatcache before, only once per request. 

So will you call clearstatcache on each request, to have reliable timestamps? If so: this will kill performance as the stat cache will be cleared a lot and therefore be useless. Of not: the timestamps won't be reliable. It probably works locally as it is correctly implemented on MacOS. 
As I said above we chose in the past to touch the parameters file for that reason.
But maybe others should give their opinion.

carakas: Ah oke, I see now, I thought that cache was only kept during the current request

carakas: What would you suggest instead? hashing the file? But that would probably be rather slow

tijsverkoyen: Some ideas:

Use the **latest commit hash** as version string, but that would require everyone to use git on their server which is not the case

Set a **version number in parameters**, and change this on deploy. Somewhat like is used in Symfony: https://symfony.com/doc/3.3/reference/configuration/framework.html#reference-framework-assets-version

Create a **hash on the first request** and store it the cache-folder. 

But the most usefull implementation would be to implement Encore and don't ship with precompiled assets, so they have a unique name every time it is compiled: https://symfony.com/doc/current/frontend/encore/versioning.html#asset-versioning-and-deployment

carakas: The problem with this is that it won't work for user uploaded assets. You can basically wrap any asset in this and it will work. If we use the proposed solutions it won't work with user uploaded or dynamically generated assets

carakas: I took a look in the vendor code. Spoon library is using that function in the file upload. Doctrine is also using it but they clear the cache specifically for that file. Maybe we could do that?
Symfony even runs it like `clearstatcache();` when asking the size of an uploaded file

tijsverkoyen: If the MediaLibrary is used for all user uploaded assets you could use the `editedOn`-field of a MediaItem as a version-string.

In my opinion your current solution isn't reliable, but I won't oppose against it. Caching is hard, and removing something from cache even harder.

tijsverkoyen: They probably call `clearstatcache` on upload to get an non-cached version of the size. Uploading files does not happen on each request, so this won't be a performance killer.

I wasn't aware Spoon/Symfony did a `clearstatcache` on upload, so in essence this means that your method will work for assets uploaded thru Spoon/Symfony (in essence: all modules + media library).

For assets added thru SSH/FTP/deployment it can happen that the cache is not cleared if there is no file uploaded. But that could easily be fixed by adding a `clearstatcache` on deployment.

carakas: Good point about the media library.
We would indeed then need to add it to the deployment, or we can add it to the `forkcms:clear:cache` command?

tijsverkoyen: It would make sense to add in into `forkcms:clear:cache`.

Also, another setback:
> Note that, unlike APCu and Opcache, the file status cache is per-process rather than stored in shared memory. This means that running stat:clear against PHP-FPM will only affect whichever FPM worker responds to the request, not the whole pool.
Source: https://gordalina.github.io/cachetool/

`stat:clear` essential runs `clearstatcache`. So on servers running PHP FPM (which most hosters are using) it probably still isn't reliable ;-)

I would think there is a reason why Symfony does not have a default cache strategy based on modified time

I think the most realistic approach is:

1. Using the `editedOn`-field as a version string for assets that are uploaded with the MediaLibrary.
2. Convert core modules to use the MediaLibrary instead of own upload mechanism.
3. Implement Encore and enable versioning

StijnVrolijk: +1 for adding it to `forkcms:cache:clear`. It makes the most sense.

carakas: fixed in e5feef65c
@tijsverkoyen but that issue already exists as we use the exact same system on the parameters.yml afaik

PR #3035, opened 2020-02-06
Add api call for a page block

carakas:
## Type

- Feature

## Pull request description

Adds an api call to get the content that the widget or action would've sent to the twig template

example call
`/api/pages/en/1/page-block/9.json`

There is a lot of stuff in the way that we render things that can be refactored but that is out of scope for this pr

### DIFF
--- a/src/Backend/Modules/Pages/Domain/ModuleExtra/ModuleExtraType.php
+++ b/src/Backend/Modules/Pages/Domain/ModuleExtra/ModuleExtraType.php
@@ -71,5 +71,4 @@ final class ModuleExtraType implements JsonSerializable
     {
         return (string) $this;
     }
-
 }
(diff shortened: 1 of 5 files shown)


StijnVrolijk:
@carakas why do we need this feature? How is the request built up? What is `9`? Is it the revision ID?

carakas:
9 is the module extra id and 1 is the page.
It is used to get the data that the widget would pass to the template if that module extra was rendered on that page. If you want to build a headless site you need this to get the data from the widget/action that is added in the backend to the page

src/Backend/Modules/Pages/Domain/ModuleExtra/ModuleExtraType.php
```diff
@@ -56,4 +71,5 @@ public function jsonSerialize(): string
     {
         return (string) $this;
     }
+
```
StijnVrolijk: This whitespace should be removed according to PSR-2

StijnVrolijk:
@carakas how do you plan on fetching something like a blogpost that gets its information from the URL? e.g. `/en/blog/detail/my-blog-post` won't work with this implementation because the block on the detail page requires `my-blog-post` to know what to fetch.

StijnVrolijk (changes requested):

PR #3047, opened 2020-02-24
Google Tag Manager

tijsverkoyen:
## Type

- Enhancement
- Feature

### New Google Analytics options

We removed automatically adding Google Analytics code if you link the Analytics
module. From now on you have two options to let Fork add Google tracking:

Interface for setting the Google ID's

You can select them both, but that is not something I would recommend. In this case both the Google Analytics and Google Tag Manager code will be added.

#### Google Analytics
If you don't want to use Google Tag Manager you can just set the Tracking ID for Google Analytics. With this we will add the Google Analytics code directly.

#### Google Tag Manager
Google Tag Manager is a very powerful product from Google where you can manage
everything yourself, or allow external users access to manager the scripts
for you. With Google Tag Manager you don't need a Fork CMS developer to add
scripts for you.

If you want to add Google Tag Manager to your Fork CMS website you need to fill
in the Google Tag Manager Container Id in the Settings screen under Google Tag 
Manager Container Id. You can find this Id by entering Google Tag Manager, in the 
overview it is listed under Container ID.

##### anonymize IP

If you are targeting users in the EU, you have probably heard of GDPR. GDPR is all
about protecting the personal data of people. 

Google has made it possible to anonimize the IP's that are send, so by default we 
will anonimize them.<sup>1</sup> If the user allows it we will not anonymize them anymore. At the moment this happens as they agree on the current Cookie Bar implementation.

To make this possible we need to configure a variable in Google Tag Manager:

* Open Variables
* Click "New" next to "User-Defined Variables"
* Name it "DataLayer - anonymizeIp"
* Choose "Data Layer Variable" for the type
* Use "anonymizeIp" as the value for "Data Layer Variable Name"
* Enable "Set Default Value", and set the Default Value to "true"

Configuration of the anonymizeIp variable

<sup>**1**</sup>: In Belgium this is not enough, the Belgian DPA has said that anonimyzeIp is not 
enough. So if you are a Belgian company you should ask the visitor permision before loading 
Google Analytics.

##### Google Analytics

The next step is to add Google Analytics thru Google Tag Manager.

**Configure a variable which contains the Tracking ID, this is optional but advised** 
* Open Variables
* Click "New" next to "User-Defined Variables"
* Name it "GA - Tracking ID"
* Choose "Google Analytics setting" as the type
* Enter the Tracking ID into the correct field, it should look like UA-xxxxxxx-y

**Add the Universal Analytics**
* Open Tags
* Click "New"
* Choose "Google Analytics: Universal Analytics" as the type
* Choose the "GA - Tracking ID" variable for Google Analytics settings
* Enable "Enable overriding settings in this tag"
* Leave "Tracking ID" empty
* Click "More settings" --> "Field to Set"
* Add the field "anonymizeIp", and set the value to the variable DataLayer - anonymizeIp.
* Choose "All pages" for the trigger

Don't forget to publish the changes.


carakas:
@tijsverkoyen Would it make sense to deprecate the analytics module in Fork after this PR is merged since the whole setup has moved somewhere else? it shows some graphs but google analytics shows them better and you can do more with it.

tijsverkoyen:
I think the integrated analytics module has it value for a lot of people who are not that experienced in Google Analytics. On the other hand it would mean that there is a lot of code not to be maintained anymore.
Maybe ask the opinion on this in the (Slack) community?

bart-webleads:
We (webleads) never load Google Analytics directly.
We always use the Google Tag Manager.

So I would say, remove the GA code from ForkCMS.

jacob-v-dam:
I think removing GA could be ok. But for the time being add both to Fork until version 6.0. If you remove GA a good manual should be available on how to add GA in combination with GTM.

Somewhere else I noticed a discussion about the graphs in Fork, if these should stay or be removed. At first I thought remove them from Fork. A week ago the api or something had an error this resulted in a 500 error, and users where unable to login.

But on the other hand users look at these graphs since GA is sometimes to complicated. If users get more used to GA they prefer this above the Fork stats.

tijsverkoyen:
I think there are several questions to be answered:

**1. Should we remove the Analytics module.** 
As with this PR the Google Analytics code won't be added automatically anymore. The Analytics module uses the Google Analytics API and that API is prune to change, it means maintaining code.
Jelmers argument is that Google Analytics has a better interface, more features, ... The current Analytics module is a very simple interface which only allows the admin to see very little and basic information.

**2. Should we provide the option to choose between Google Analytics and Google Tag Manager**
This PR provides the user with 3 options:
1. Use the scripts-field to add code yourself
2. Use Google Analytics
3. Use Google Tag manager

**3. Some question about the removal of graphs**
See, the comment from Jacob: https://github.com/forkcms/forkcms/pull/3047#issuecomment-591315837


My personal opinion:

**1:** I think the current Analytics has value for non-professional users or users that only need basic stats. At this moment I would vote against removing the module.
**2:** The user should be able to choose, with the 3 options you can anything you want. Google Tag Manager is complicated if you don't are experienced with it.
**3:** Not relevant for this PR, so if you want to discuss this: create a PR/Issue.

carakas:
I would keep all three options as well. I find the arguments of @tijsverkoyen convincing.
Is there still a lot of work that needs to be done for this PR @tijsverkoyen? Any idea on a timeline?
I think it might be a good idea to release 5.9.0 ones this one and #3048 are finished if it is in the near future

tijsverkoyen:
@carakas For this PR there is no work left. For https://github.com/forkcms/forkcms/pull/3048 I need to review the comments again and check to finish up the code, I will try to finish it this month.

Let me know if you merge these PR's, in that case I will create a blogpost based on the PR descriptions, so we can use it as a reference/docs for these features.

src/Frontend/Core/Engine/Footer.php
```diff
@@ -43,26 +43,26 @@ public function parse(): void
...
             $searchUrl = FrontendNavigation::getUrlForBlock('Search');
             $url404 = FrontendNavigation::getUrl(Model::ERROR_PAGE_ID);
             if ($searchUrl !== $url404) {
-                $siteHTMLFooter .= $this->getSiteLinksCode($searchUrl);
+                $siteHTMLEndOfBody .= $this->getSiteLinksCode($searchUrl);
             }
         }
 
         // assign site wide html
-        $this->template->assignGlobal('siteHTMLFooter', $siteHTMLFooter);
```
carakas, line 65: for now I would assign it to the both of them for BC reasons. There are people that upgrade to the latest version of fork and this will break the theme that they use

src/Frontend/Core/Header/Header.php
```diff
@@ -351,14 +352,36 @@ public function parse(): void
...
         $this->jsFiles->parse($this->template, 'jsFiles');
 
-        $siteHTMLHeader = (string) $this->get('fork.settings')->get('Core', 'site_html_header', '') . "\n";
-        $siteHTMLHeader .= new GoogleAnalytics(
-            $this->get('fork.settings'),
-            Model::getRequest()->getHttpHost(),
-            $this->get('fork.cookie')
-        );
-        $siteHTMLHeader .= "\n" . $this->jsData;
-        $this->template->assignGlobal('siteHTMLHeader', trim($siteHTMLHeader));
```
carakas, line 361: for now I would assign it to the both of them for BC reasons. There are people that upgrade to the latest version of fork and this will break the theme that they use

src/Backend/Modules/Settings/Actions/Index.php
```diff
@@ -352,20 +400,43 @@ private function validateForm(): void
...
+                );
+
+                $this->get('fork.settings')->set(
+                    'Core',
+                    'site_html_head',
+                    $this->form->getField('site_html_head')->getValue()
                 );
                 $this->get('fork.settings')->set(
                     'Core',
-                    'site_start_of_body_scripts',
```
carakas, line 362: I would write to the old one as well with a deprecation notice for BC reasons. There are people that upgrade to the latest version of fork and  this might create issues if they use that setting somewhere

src/Backend/Modules/Settings/Actions/Index.php
```diff
@@ -352,20 +400,43 @@ private function validateForm(): void
...
-                    $this->form->getField('site_start_of_body_scripts')->getValue()
+                    'site_html_start_of_body',
+                    $this->form->getField('site_html_start_of_body')->getValue()
                 );
                 $this->get('fork.settings')->set(
                     'Core',
-                    'site_html_footer',
```
carakas, line 367: I would write to the old one as well with a deprecation notice for BC reasons. There are people that upgrade to the latest version of fork and this might create issues if they use that setting somewhere

tijsverkoyen: Fixed in dabf936ae918db1eac41e60b3cfe98587fc8d0aa

src/Backend/Modules/Settings/Actions/Index.php
tijsverkoyen: Fixed in dabf936ae918db1eac41e60b3cfe98587fc8d0aa

src/Frontend/Core/Header/Header.php
tijsverkoyen: Fixed in dabf936ae918db1eac41e60b3cfe98587fc8d0aa

src/Frontend/Core/Engine/Footer.php
tijsverkoyen: Fixed in dabf936ae918db1eac41e60b3cfe98587fc8d0aa

PR #3059, opened 2020-03-26
Fix images always being put in a source directory even when disabled

carakas:
## Type

- Non critical bugfix

## Pull request description

By overwriting the constant `GENERATE_THUMBNAILS` you can prevent that a source directory is used while uploading, but it didn't check in all places correctly if it was overwritten by the parent


carakas:
Hmm, something else is wrong, I'll check and create different pr

PR #3063, opened 2020-03-31
Change to PREG_SET_ORDER

Hk-tang:
## Type

- Enhancement

## Pull request description

Using PREG_SET_ORDER (rather than the default PREG_PATTERN_ORDER) delivers an easier to process result.


carakas:
@Hk-tang I don't see any benifit of changing it, can you elaborate?

Hk-tang:
I just think the logic makes more sense to process in order by sets of matches instead of how much of the pattern is matched.

carakas:
@Hk-tang can you write a test for this proving that it doesn't break the code?

carakas:
@Hk-tang Nothing changed, you just merged master (in a way that all the commits are showing…).
You also haven't responded to my last question so why are you requesting a new review?

Hk-tang:
Hi carakas, I merged master because the fork/branch was out of date. I think it is important to keep things up to date as there is a chance the changes I make won't work when I update the branch later, as such I thought it proper to update the branch first. Merging in a way that the commits are showing is known as (rebasing)[https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase]. Since the README.md does not suggest a preference between merging and rebasing I chose to rebase as it removes an extra merge commit and keeps the history clean. I did not request a new review, that was GitHub sending you a notification because this PR was updated in some way (in this case updating to keep the branch up to date). On the right side of the PR there is an option to unsubscribe from the notifications and you will no longer get notified while the PR is being updated. That way when I do actually request a review then I can do so with the @mention like so, @carakas, or by assigning a reviewer also in the top right of the PR. I am sorry for not responding to your request so I will do so now. I will try and add a test. Thank you for responding to me on Slack as I was having trouble running the tests (and still am). If you have any other concerns or questions please feel free to ask.

carakas:
Going to close this one since it makes no noticible difference here

PR #3066, opened 2020-04-14
Different url per locale

carakas:
## Type

- Feature

## Pull request description

possible to configure different urls per locale.

You can do this by replacing `site.domain` in `app/config/parameters.yml` with `site.domain.[language]`

Example:

    parameters:
        database.driver:        mysql
        …
        site.protocol:          https
        site.domain.en:         fork-cms.com
        site.domain.nl:         fork-cms.be
        site.domain.fr:         fork-cms.fr
        site.domain.de:         fork-cms.de
        site.default_language:  en
        site.multilanguage:     true
        …
        action.rights_level:    7

If there isn't any url configured for a requested language the url for the default language will be loaded.
The domain for the current language will be available in `site.domain`    

The Locale classes extending from `\Common\Locale` also have a new method `getDomain` to get the correct domain for that locale.

Note: This doesn't work if the parameter `site.domain` already exists in `app/config/parameters.yml`


app/Kernel.php
```diff
@@ -222,7 +222,7 @@ public function getLogDir(): string
 
     public function getCacheDir(): string
     {
-        return dirname(__DIR__) . '/var/cache/' . $this->environment;
+        return dirname(__DIR__) . '/var/cache/' . $this->environment . '/' . $this->request->getHost();
```
tijsverkoyen, line 228: Won't this decrease the performance? What if I have a site that is available on multiple domains, without a redirect in .htaccess.
Or can you explain why you have a cache per host.

app/Kernel.php
```diff
@@ -225,6 +228,50 @@ public function getCacheDir(): string
...
+        $defaultLocale = $container->getParameter('site.default_language');
+        $currentLocale = $chunks[0] ?? $defaultLocale;
+
+        $this->request->setLocale(strlen($defaultLocale) === 2 ? $currentLocale : $defaultLocale);
+        $this->request->setDefaultLocale($defaultLocale);
+    }
+
+    private function getSiteDomainForCurrentLocale(ContainerBuilder $container): string
+    {
+        if ($container->hasParameter('site.domain')) {
```
tijsverkoyen, line 247: What none of the parameters is defined?

app/Kernel.php
```diff
@@ -225,6 +228,50 @@ public function getCacheDir(): string
...
+        $this->setLocaleOnRequest($container);
+
+        if ($container->hasParameter('site.domain.' . $this->request->getLocale())) {
+            return $this->verifyDomainMatchesLocale(
+                $container->getParameter('site.domain.' . $this->request->getLocale())
+            );
+        }
+
+        return $this->verifyDomainMatchesLocale(
+            $container->getParameter('site.domain.' . $this->request->getDefaultLocale())
```
tijsverkoyen, line 260: What if there is no default locale defined? Or will this always be the case?

docs/02. getting started/09. different_url_per_locale.md
```diff
@@ -0,0 +1,27 @@
...
+
+    parameters:
+        database.driver:        mysql
+        …
+        site.protocol:          https
+        site.domain.en:         fork-cms.com
+        site.domain.nl:         fork-cms.be
+        site.domain.fr:         fork-cms.fr
+        site.domain.de:         fork-cms.de
+        site.default_language:  en
```
tijsverkoyen, line 17: Should we rename this to `default_locale`?

docs/02. getting started/09. different_url_per_locale.md
```diff
@@ -0,0 +1,27 @@
...
+        site.multilanguage:     true
+        …
+        action.rights_level:    7
+
+If there isn't any url configured for a requested language the url for the default language will be loaded.
+The domain for the current language will be available in `site.domain`    
+
+The Locale classes extending from `\Common\Locale` also have a new method `getDomain` to get the correct domain for that locale.
+
+Note: This doesn't work if the parameter `site.domain` already exists in `app/config/parameters.yml`
```
tijsverkoyen, line 27: So in essence you have to chose between

    site.domain.en:         fork-cms.com
    site.domain.nl:         fork-cms.be
    site.domain.fr:         fork-cms.fr
    site.domain.de:         fork-cms.de

Or 

    site.domain:         fork-cms.com

I would explain this in the documentation that it is either one.
When I first saw the documentation, I expected that site.domain was used as the default domain

app/Kernel.php
carakas: that will always be the case, site.domain is there by default

app/Kernel.php
carakas: this one is there by default so it's always there unless you deliberetly change the config

app/Kernel.php
carakas: this was needed because the cached container parameters are different when the domain is different. Without this it would use the domain that was used to build the cache.
the first time before the cache is built might have a slight difference but it has no effect afterwards. 
Redirects in htaccess won't be needed anymore since that will happen automatically if the url is configured for a language

docs/02. getting started/09. different_url_per_locale.md
carakas: we should in fork 6, this already exists now so we can't change it within the current version

docs/02. getting started/09. different_url_per_locale.md
carakas: the default domain is based on the default language

```
site.default_language:  en
site.domain.en:         fork-cms.com
site.domain.nl:         fork-cms.be
```
if you go there in fr you'll get fork-cms.com

carakas: How would you phrase it differently?
The docs say
```
You can do this by replacing `site.domain` in `app/config/parameters.yml` with `site.domain.[language]```
and ```If there isn't any url configured for a requested language the url for the default language will be loaded.```

carakas:
Closing since symfony has a better built in solution

PR #3067, opened 2020-04-14
Use sinle quotes for background images

bjorvack:
## Type
- Non critical bugfix

## Pull request description
When using user templates the generated background image uses double quotes. Because of this the background url doesn't work. When using single quotes this problem is avoided

### DIFF
--- a/src/Backend/Modules/Pages/Js/Pages.js
+++ b/src/Backend/Modules/Pages/Js/Pages.js
@@ -1088,7 +1088,7 @@ jsBackend.pages.extras = {
       $img = $placeholder.find('#user-template-image-background-' + key + ' img')
 
       $element.attr('data-src', $img.attr('src'))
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -1088,7 +1088,7 @@ jsBackend.pages.extras = {
       $img = $placeholder.find('#user-template-image-background-' + key + ' img')
 
       $element.attr('data-src', $img.attr('src'))
-      $element.css('background-image', 'url("' + $img.attr('src') + '")')
+      $element.css('background-imag\'' + $img.attr('src') + '\')')
```
carakas, line 1091: the e is missing

bjorvack: @carakas fixed the typo 369a697

PR #3075, opened 2020-04-22
Preview editor

carakas:
## Type

- Bugfix

## Pull request description

Fixes the preview of the editor and images disappearing when something besides the image is changed


src/Backend/Modules/Pages/Js/Pages.js
```diff
@@ -125,23 +125,54 @@ jsBackend.pages.extras = {
       var $modal = $(this).closest('.modal')
       var separator = ' › '
       var $pageBlockTitle = $modal.closest('[data-role="page-block-wrapper"]').find('[data-role="page-block-title"]')
+      var $pageBlockPreview = $modal.closest('[data-role="page-block-wrapper"]').find('[data-role="page-block-preview"]')
```
Katrienvh, line 128: ideally `$modal.closest('[data-role="page-block-wrapper"]')` should be a variabele, so it's set once en the variable is used here and a line higher

PR #3079, opened 2020-04-29
Fix upscale cropping

carakas:
## Type

- Non critical bugfix

## Pull request description

When trying to crop an image to a size that is bigger than the size of the image we sometimes gor a black bar at the bottom (in this case cropping to 800x600)
src image
coronamaatregelen-in-de-vastgoedsector-wat-te-doen-nl-7
original crop
coronamaatregelen-in-de-vastgoedsector-wat-te-doen-nl-7
crop with changes from this pr
image


tijsverkoyen:
Can you give an example (current implementation and with the PR applied) for a portrait image

carakas:
I think I'm just going to refactor this entire thing and write tests for it. We've had so many issues with it that I don't really see a reason why that shouldn't be done

src/ForkCMS/Utility/Thumbnails.php
```diff
@@ -157,4 +147,87 @@ public function getFolders(string $inPath, bool $includeSourceFolder = false): a
 
         return array_values($folders);
     }
+
+    /**
+     * Calculate the desired width based on the aspect ratio if needed
```
tijsverkoyen, line 152: Isn't this useless phpdoc? The method signature explains it very well I would think

src/ForkCMS/Utility/Thumbnails.php
```diff
@@ -157,4 +147,87 @@ public function getFolders(string $inPath, bool $includeSourceFolder = false): a
...
+     *
+     * @param int|null $desiredWidth
+     * @param int|null $desiredHeight
+     * @param int $currentWidth
+     * @param int $currentHeight
+     * @return int
+     */
+    private function calculateDesiredWidth(
+        ?int $desiredWidth,
```
tijsverkoyen, line 161: I don't think it is good practice to have optional arguments at the beginning.

src/ForkCMS/Utility/Thumbnails.php
```diff
@@ -157,4 +147,87 @@ public function getFolders(string $inPath, bool $includeSourceFolder = false): a
...
+
+        // handle width too small
+        if ($width < $desiredWidth) {
+            $scaleRatio = $desiredWidth/$width;
+            $height *= $scaleRatio;
+            $width *= $scaleRatio;
+        }
+
+        // handle height too small
+        if ($height < $desiredHeight) {
```
tijsverkoyen, line 225: What happens if the height and weight are to small? I probably am missing something, but in that case you could end up with  a resized image based on the height?

src/ForkCMS/Utility/Thumbnails.php
carakas: This isn't an optional parameter, it just means that it can be null or int.
I could flip the parameters around so they match what we have in the other method

tijsverkoyen: 😱true that, seems like it was too early for me.

src/ForkCMS/Utility/Thumbnails.php
carakas: That's why I didn't use an else
When asking these images for a crop of 300x300
landscape 200x100
portrait 100x200
square 200x200

Krijg je dit
landscape
portrait
square
Hij doet dus een upscale en dan een crop

src/ForkCMS/Utility/Thumbnails.php
carakas: Changed it anyway for consistency sake
14c6ad269

src/ForkCMS/Utility/Thumbnails.php
carakas: removed it in 14c6ad269

carakas:
[thumbnails.zip](https://github.com/forkcms/forkcms/files/4579547/thumbnails.zip)

you can check the generated images from the test here

PR #3084, opened 2020-05-01
Fix broken Urls

crydotsnake:


crydotsnake:
Here: https://docs.fork-cms.com/documentation/detail/media-library/image-management 

The `view config.yml here.` Link is broken too.

PR #3086, opened 2020-05-02
Update Contribution Links.

crydotsnake:


carakas:
this will break the linking when you watch them locally in your project

PR #3102, opened 2020-05-15
Make max amount of display name changes dynamic

bjorvack:
## Type
- Enhancement

## Pull request description
Not every user needs a limitation on display name changes. This change allow the admin to set the number of changes or make it unlimited

### DIFF
--- a/src/Frontend/Modules/Profiles/Engine/Model.php
+++ b/src/Frontend/Modules/Profiles/Engine/Model.php
@@ -6,6 +6,7 @@ use Common\Uri as CommonUri;
 use Frontend\Core\Engine\Model as FrontendModel;
 use Frontend\Core\Engine\Navigation as FrontendNavigation;
 use Frontend\Modules\Profiles\Engine\Authentication as FrontendProfilesAuthentication;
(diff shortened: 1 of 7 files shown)


src/Frontend/Modules/Profiles/Engine/Model.php
```diff
@@ -22,6 +23,21 @@ class Model
      */
     private static $avatars = [];
 
+    public static function maxDisplayNameChanges(): int
+    {
+        return FrontendModel::getModuleSetting('Profiles', 'max_display_name_changes', self::MAX_DISPLAY_NAME_CHANGES);
```
carakas, line 28: @bjorvack could you maybe set this setting in the installer as well? that way it is clear when looking at the installer what settings are available?
You can keep the default here for websites that upgrade

carakas: @bjorvack can you have a look at this?

bjorvack: @carakas 6017f82

PR #3105, opened 2020-05-15
Upgrade styling ajax loader

Katrienvh:
## Type

- Enhancement

## Pull request description

Made the ajax loader more visible, because it was so tiny you would miss it

before:
image

after:
image


freshface:
The ajax loader does not matches the look and feel of the rest of the UI.

carakas:
@Katrienvh can you have a look at the comment of @freshface

bramds:
agree with @freshface

carakas:
@freshface @bramds What do you think of this
image
The same loader but in the right colour palet

sarahmey:
If I can share my opinion on this matter here too:

I think the loading indicator is a million times better then what it was. It is visible and gives the user the feedback that he/she needs: something is loading, so calm down and wait for it to finish. The loading icon that @freshface suggests does not seem right to me in this case. It is a good solution for buttons but not for this. If we will use a loading icon, it will easily go unnoticed, like how it was before. 
I think the green is also good, because it stands out more, and green goes with a positive feeling (like 'yeey it's loading' in stead of 'why is this so slow?'). In the blue version the loading indicator is more blending in and less noticable. 

I think you need to focus more on what is good for UI/UX and less on what do I think is pretty or not.

StijnVrolijk:
I agree with @sarahmey. 

Althought I'm not a fan of any of the things posted here, my preference goes to the green loading bar, purely for visibility's sake.

carakas:
@sarahmey generally speaking that loading bar won't be visible for a long time. If I quickly see something blue flashing on my screen I'll be less worried than if I would see something green.
Translations in js are loaded with ajax, so that means you can get in a situation where you don't perform any action but because of the js the locale ajax call gets triggered and you see something green flashing.
Green is something I associate with something has been changed successfully. So I would worry that I might have saved or changed something accidentally

jessedobbelaere:
+1 on the blue IMO 😌. I agree with carakas that green is for a successful task (just like red for errors, yellow for warnings). While something is loading, it's state is neutral which is the blue color in regards to Fork CMS...

carakas:
@Katrienvh can you flip the colour to blue and resolve the merge issues?

Katrienvh:
@carakas color is blue again and conflicts are fixed.

PR #3113, opened 2020-05-20
Fix layout faq feedback

Katrienvh:
## Type

- Non critical bugfix

## Pull request description

Fix layout faq feedback because it was broken

image

### DIFF
--- a/src/Backend/Modules/Faq/Layout/Templates/Edit.html.twig
+++ b/src/Backend/Modules/Faq/Layout/Templates/Edit.html.twig
@@ -103,6 +103,7 @@
                 <div class="col-lg-6 col-xl-4">
                   <div class="card card-default">
                     <div class="card-body pb-2">
+                      <small>{{ _itr_11.created_on|date('m/d/Y') }}</small>
                       <p>
                         {{ _itr_11.text }}
                       </p>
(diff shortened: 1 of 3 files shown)


src/Backend/Modules/Faq/Layout/Templates/Edit.html.twig
```diff
@@ -97,35 +97,29 @@
...
-                <div id="_itr_11-collapse-{{ _itr_11.id }}" class="card-collapse collapse" role="tabpanel" aria-labelledby="_itr_11-heading-{{ _itr_11.id }}">
-                  <div class="card-body">
-                    {{ _itr_11.text }}
-                  </div>
-                  {% if showFaqDeleteFeedback %}
-                    <div class="card-footer">
-                      <div class="btn-toolbar">
-                        <div class="btn-group pull-left" role="group">
-                          {{ macro.buttonIcon('#', 'trash-alt', 'lbl.Delete'|trans|ucfirst, 'btn-danger', 'data-id=' ~ _itr_11.id ~ ' data-role=delete-feedback') }}
+          <div id="feedback" role="tablist" aria-multiselectable="false">
```
sarahmey, line 100: I'm not sure this role attribute is used correctly here. A quick search learned me that a `role="tablist"` should have children with `role="tab"`. You should probably look into the aria-attribute too. I'm not an expert, but it looks like you just copied it without thinking about it ;)

Katrienvh: Indeed it was already there, I just changed some code in between and didn't pay attention to it, nice catch!

PR #3116, opened 2020-05-24
Add CSS class to a menu link

wouterverstuyf:
## Type
- Enhancement
- Feature

Add a css class to the menu links
The class is stored in the data column of the pages table, so no db change is reqiured.

Schermafbeelding 2020-05-21 om 15 39 41

It's usefull if you want to add icons or other stuff to the menu links, for example:

Schermafbeelding 2020-05-25 om 09 14 07

## Pull request description

Adds a text field to fill in a css class.
Only visible to god-users.


src/Backend/Modules/Pages/Installer/Data/locale.xml
```diff
@@ -2,6 +2,11 @@
 <locale>
   <Backend>
     <Pages>
+      <item type="label" name="LinkCssClass">
+        <translation language="nl"><![CDATA[link CSS class]]></translation>
+        <translation language="en"><![CDATA[link CSS class]]></translation>
+        <translation language="pl"><![CDATA[link CSS class]]></translation>
```
carakas, line 8: I don't think this is correct Polish

src/Frontend/Modules/Pages/Layout/Widgets/SubpagesDefault.html.twig
```diff
@@ -17,7 +17,7 @@
     {% block widget_body %}
       <div class="widget-body media">
         {% for subpage in widgetSubpages %}
-          <div class="media-body">
+          <div class="media-body{% if subpage.data.link_class %} {{ subpage.data.link_class }}{% endif %}">
```
carakas, line 20: you placed this class on this a-element in the other places; why did you place it on the div here?

src/Backend/Modules/Pages/Installer/Data/locale.xml
wouterverstuyf: Correct :-D
Something like "link do klasy CSS"? :)

src/Frontend/Modules/Pages/Layout/Widgets/SubpagesDefault.html.twig
wouterverstuyf: When I should use the subpages I would like to have the class on the container, so you can work with the "subpage block". For example, adding a background color/image, adding a hover effect, ... in most cases I make this block fully clickable for better UX.

src/Backend/Modules/Pages/Installer/Data/locale.xml
carakas: No clue, I don`t speak Polish :) if you`re not sure I would remove the translation

wouterverstuyf: Done :)

src/Frontend/Modules/Pages/Layout/Widgets/SubpagesDefault.html.twig
carakas: I would add it by default in the fork theme to the link to keep it consistant. You can always change it in your personal projects

wouterverstuyf: Good point. Done!

PR #3117, opened 2020-05-25
Setup overview with pages for each language

wouterverstuyf:
## Type
- Feature

## Pull request description

When you have multiple languages and add pages (after install) the hreflang link between the languages is no longer insured. As these meta info is pretty important for Google, I think it's useful to have a global overview where it's possible to quickly change linked pages. This is much faster than manually edit the page...
I came along (after a quick setup) with the following:
Schermafbeelding 2020-05-24 om 22 37 27

- at the right column you see the Working Language (still has to be ordered at the left)
- the other 2 columns are the other 2 installed languages (FR and ENG in the example)
- the dropdowns are the values you can set for each page individually in the SEO tab

To do:
- put the working language as the first column in the datagrid
- maybe we can show warnings if there are "empty values" in the dropdowns, so these links are correct?
- code optimizations

### DIFF
--- a/src/Backend/Modules/Pages/Engine/Model.php
+++ b/src/Backend/Modules/Pages/Engine/Model.php
@@ -840,11 +840,11 @@ class Model
 
                 // insert link
                 $html .= '    <a href="' . BackendModel::createUrlForAction(
-                    'Edit',
-                    null,
-                    null,
-                    ['id' => $page['page_id']]
-                ) . '"><ins>&#160;</ins>' . htmlspecialchars($page['navigation_title']) . '</a>' . "\n";
+                        'Edit',
+                        null,
+                        null,
+                        ['id' => $page['page_id']]
+                    ) . '"><ins>&#160;</ins>' . htmlspecialchars($page['navigation_title']) . '</a>' . "\n";
 
                 // get childs
                 $html .= self::getSubtree($navigation, $page['page_id']);
@@ -937,11 +937,11 @@ class Model
         // create homepage anchor from title
         $homePage = self::get(BackendModel::HOME_PAGE_ID);
         $html .= '            <a href="' . BackendModel::createUrlForAction(
-            'Edit',
-            null,
-            null,
-            ['id' => BackendModel::HOME_PAGE_ID]
-        ) . '"><ins>&#160;</ins>' . $homePage['title'] . '</a>' . "\n";
+                'Edit',
+                null,
+                null,
+                ['id' => BackendModel::HOME_PAGE_ID]
+            ) . '"><ins>&#160;</ins>' . $homePage['title'] . '</a>' . "\n";
 
         // add subpages
         $html .= self::getSubtree($navigation, BackendModel::HOME_PAGE_ID);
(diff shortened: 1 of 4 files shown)


src/Backend/Modules/Settings/Actions/Seo.php
```diff
@@ -60,6 +117,25 @@ private function validateForm(): void
...
+                        $update['data'] = $pageData['data'];
+
+                        foreach ($this->languages as $lang => $language) {
+                            if ($lang != BL::getWorkingLanguage()) {
+                                $update['data']['hreflang_' . $lang] = $this->form->getField('page_' . $lang . '_' . $pageId)->getValue();
+                            }
+                        }
+
+                        $update['data'] = serialize($update['data']);
+                        BackendPagesModel::updateByIdAndRevisionId($update);
```
carakas, line 135: Why not save it as a new revision?

wouterverstuyf: No specific reason, seems not necessary because the actual content of the page is not changed?

carakas: I get that logic, but I'm also thinking about needing to port this to the restructure core branch where the pages module is running on doctrine already, and I think adding an additional way of updating pages adds complexity that can potentially cause issues in the future

wouterverstuyf: Aha, got it. It's quickly changed to a new revision. Let me know ;)

carakas: I would just use the normal update method, that way it should be able to be ported without issues to the restructure-core branch

wouterverstuyf: Great, I will change that this week ;)

wouterverstuyf: Hmz, we need to build the whole page array to create a new revision, so also a new meta record.  What is the best way to create a new Meta record?
Building the array (title, keywords,...), add it manually to the db, and store the returned meta ID to the new revision?

carakas: Hmm, maybe then make the scope smaller like update revisionData or something ?

wouterverstuyf: Ok, updated the function and changed the parameters so it's more clear this is only to update the data column

carakas:
@wouterverstuyf can you look at this https://travis-ci.org/github/forkcms/forkcms/jobs/692153791 some code style issues

src/Backend/Modules/Pages/Engine/Model.php
```diff
@@ -1430,7 +1430,7 @@ public static function updateRevisionData(int $pageId, int $revisionId, array $d
             [$pageId, $revisionId]
         );
     }
-
+
```
carakas, line 1433: you have some whitespace remaining

wouterverstuyf: Strange, don't see it in my editor :|

Schermafbeelding 2020-06-01 om 16 20 34

src/Backend/Modules/Pages/Engine/Model.php
```diff
@@ -840,11 +840,11 @@ public static function getSubtree(array $navigation, int $parentId): string
 
                 // insert link
                 $html .= '    <a href="' . BackendModel::createUrlForAction(
-                    'Edit',
-                    null,
-                    null,
-                    ['id' => $page['page_id']]
-                ) . '"><ins>&#160;</ins>' . htmlspecialchars($page['navigation_title']) . '</a>' . "\n";
+                        'Edit',
```
carakas, line 843: hmm now that I see this, all this extra white space is also not oke

wouterverstuyf: I have changed the original model.php from the master, so I think it's ok now?

carakas: Yes

PR #3120, opened 2020-05-26
Add new empty state

Katrienvh:
## Type

- Enhancement

## Pull request description

Made the empty state a little more attractive/visible based on a design of @sarahmey. 
The empty state is in a macro now, with a default text the fits for everything. For al the existing modules in fork there is a custom text. See example of blog below.

image


src/Backend/Core/Layout/Templates/macros.html.twig
```diff
@@ -36,19 +36,14 @@
   <abbr tabindex="0" data-toggle="tooltip" aria-label="{{ title }}" title="{{ title }}">{{ macro.icon('info-circle') }}</abbr>
 {% endmacro %}
 
-{% macro dataGrid(dataGrid, noItemsMessage) %}
+{% macro dataGrid(dataGrid, addItemLink, addItemText, noItemsText) %}
```
sarahmey, line 39: Isn't it more logical to put the `noItemsText` first, this will always be available, the `addItemsLink` and `addItemsText` might not always be available

src/Backend/Core/Layout/Templates/macros.html.twig
```diff
@@ -36,19 +36,14 @@
...
     </div>
   {% else %}
-    <p>
-      {% if noItemsMessage %}
-        {{ noItemsMessage|raw }}
-      {% else %}
-        {{ 'msg.NoItems'|trans|raw }}
-      {% endif %}
-    </p>
+    {{ macro.emptyState(addItemLink, addItemText, noItemsText) }}
```
sarahmey, line 46: Same remark here

src/Backend/Core/Layout/Templates/macros.html.twig
```diff
@@ -112,3 +107,17 @@
...
 
+{% macro emptyState(addItemLink, addItemText, noItemsText, addItemLinkExtraAttributes) %}
+  {% set addItemText = addItemText|default('msg.AddItem'|trans|raw) %}
+  {% set noItemsText = noItemsText|default('msg.NoItems'|trans|raw) %}
+
+  <div class="empty-state">
+    <img class="empty-state-illustration mb-4" src="{{ BACKEND_CORE_URL }}/Layout/images/empty-state.svg" alt="">
+    <p class="text-black mb-2"><strong>{{ noItemsText|raw }}</strong></p>
+    {% if not addItemLink is null %}
+      <a class="empty-state-add" href="{{ addItemLink }}"{% for attrname, attrvalue in addItemLinkExtraAttributes %} {{ attrname }}="{{ attrvalue }}"{% endfor %}>
```
sarahmey, line 118: just a thought
i like that you can add other attributes, but you can't add extra classes to your link, wouldn't anybody want to do that?

src/Backend/Modules/MediaLibrary/Layout/Templates/MediaBrowser.html.twig
```diff
@@ -72,9 +72,9 @@
                           </form>
                           {% if not dataGrid.numberOfResults %}
                             {% if mediaFolder %}
-                              <p>{{ 'msg.NoItems'|trans|format(geturl('add', null, '&folder=#{mediaFolder.id}')) }}</p>
+                              {{ macro.emptyState(null, null, 'msg.NoItems'|trans|ucfirst) }}
```
sarahmey, line 75: I think you can add the link here
`geturl('add', null, '&folder=#{mediaFolder.id}')`

src/Backend/Modules/MediaLibrary/Layout/Templates/MediaBrowser.html.twig
```diff
@@ -72,9 +72,9 @@
...
                             {% else %}
-                              <p>{{ 'msg.NoItems'|trans|format(geturl('add') ) }}</p>
+                              {{ macro.emptyState(null, null, 'msg.NoItems'|trans|ucfirst) }}
```
sarahmey, line 77: same here, `geturl('add')`

src/Backend/Core/Layout/Templates/macros.html.twig
Katrienvh: No it is not in my eyes, because the last two items have a default value in the macro for the empty state self. So when you want to use the defaults you only have to add the Link. I switched the order multiple times based on were and how I needed in in Fork, this order seems to be the most practical way in my eyes.

src/Backend/Core/Layout/Templates/macros.html.twig
Katrienvh: Same answer

src/Backend/Core/Layout/Templates/macros.html.twig
Katrienvh: Mm I do not thinks this will be an issue. Classes are for styling(css). We strive for equality in the code. So I suggest to won't let them add css class to maintain unity. They can add attributes, id etc when you have to link from PHP or JS

src/Backend/Modules/MediaLibrary/Layout/Templates/MediaBrowser.html.twig
Katrienvh: No thats wrong, add is not an action in de Media Library module. I also replaced it in another file with `MediaItemUpload` in `MediaLibrary/Layout/Templates/MediaItemIndex.html.twig`. The thing is, `MediaItemUpload` won't work here. Because you are in de browser. For the moment I will leave it like this. Need some customisation in de module itself to add a link to upload

src/Backend/Modules/MediaLibrary/Layout/Templates/MediaBrowser.html.twig
Katrienvh: same answer

src/Backend/Core/Layout/Templates/macros.html.twig
carakas: I prefer the current order as well

PR #3121, opened 2020-05-27
Fix docker build with zlib errors

jessedobbelaere:
## Type

- Non critical bugfix

## Pull request description

During `docker-compose up -d` or `docker-compose build` or `docker build ...` the Dockerfile would crash with following error:

```
Building app
Step 1/20 : FROM php:7.1-apache
 ---> b9858ffdd4d2
Step 2/20 : LABEL maintainer="Fork CMS <info@fork-cms.com>"
 ---> Using cache
 ---> f41fa1d2c4e9
Step 3/20 : RUN a2enmod rewrite
 ---> Using cache
 ---> 7e1396dc317b
Step 4/20 : RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list
 ---> Using cache
 ---> 8be3ed855971
Step 5/20 : RUN apt-get update && apt-get install -y --no-install-recommends --allow-downgrades     libfreetype6-dev     libjpeg62-turbo-dev     libz-dev     zlib1g=1:1.2.8.dfsg-2+b1     libpng-dev &&     docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ &&     docker-php-ext-install -j$(nproc) gd &&     rm -rf /var/lib/apt/lists/*
 ---> Running in 8985795c4a80
Get:1 http://security.debian.org jessie/updates InRelease [44.9 kB]
Ign:2 http://archive.debian.org/debian jessie InRelease
Get:3 http://archive.debian.org/debian jessie Release [148 kB]
Get:4 http://archive.debian.org/debian jessie Release.gpg [2420 B]
Get:5 http://security.debian.org jessie/updates/main Sources [350 kB]
Get:6 http://security.debian.org jessie/updates/main amd64 Packages [771 kB]
Get:7 http://archive.debian.org/debian jessie/main Sources [7063 kB]
Get:8 http://archive.debian.org/debian jessie/main amd64 Packages [6818 kB]
Fetched 15.2 MB in 3s (5160 kB/s)
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 zlib1g-dev : Depends: zlib1g (= 1:1.2.8.dfsg-2+deb8u1) but 1:1.2.8.dfsg-2+b1 is to be installed
E: Unable to correct problems, you have held broken packages.
ERROR: Service 'app' failed to build: The command '/bin/sh -c apt-get update && apt-get install -y --no-install-recommends --allow-downgrades     libfreetype6-dev     libjpeg62-turbo-dev     libz-dev     zlib1g=1:1.2.8.dfsg-2+b1     libpng-dev &&     docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ &&     docker-php-ext-install -j$(nproc) gd &&     rm -rf /var/lib/apt/lists/*' returned a non-zero code: 100
```


jessedobbelaere:
Seems the line

```
# Run apt from fresh debian sources
RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list
```

is causing some conflicts with the GD dependencies all the times. Was there a specific reason you added it @StijnVrolijk ? Think https://github.com/forkcms/forkcms/pull/2891 was related too... If i leave out this line like in my PR it builds correctly... I think it causes some flakyness for reliable Docker builds

docker-compose.yml
```diff
@@ -17,6 +17,8 @@ services:
         volumes:
             - .:/var/www/html:cached
             - ./src/Frontend/Files:/var/www/html/src/Frontend/Files:cached
+            - fork-cms-cache:/var/www/html/var/cache
+            - fork-cms-logs:/var/www/html/var/logs
```
ohvitorino, line 21: Hey @jessedobbelaere ! Why do you create these volumes? Won't they make the application slower?

jessedobbelaere: Hmm, the reason why I added it now (and not earlier), is because there were errors for the logs and cache dir in the Travis job I added when Fork CMS starts up: 

```
RuntimeException: Unable to write in the "cache" directory (/var/www/html/var/cache/install)
```

At the company I work for, we also keep Symfony's cache and logs in an internal volume. I wouldn't say it slows performance down, it should actually improve performance because it will not continuously sync the big cache and logs dir from inside Docker to the host machine (your mac/windows). For that reason, [we also usually keep the `vendor/` folder in its own volume](http://blog.michaelperrin.fr/2017/04/14/docker-for-mac-on-a-symfony-app/#solution1movevendorsoutoftheshareddirectory), so Docker does not have to sync between Mac and the container all the time. But I didn't add the vendor folder here because it causes confusion for people: they run composer install on their mac and it does not reflect in the container because there's no mount of that folder... 

The `logs/` and `cache/` directories stay in their own internal volume, which means it does not mount the directory from your mac too which gave the permission problems (you would need to `chown` the directories again like the Dockerfile does). The only downside is that people need to clear the logs/cache in their container (which they normally do via the `bin/console c:c` command) instead of deleting files on their mac. Or you can do `docker-compose down -v` to drop the project's volumes and start fresh again.

ohvitorino: Then maybe we should mention this cache stuff in the README file.

carakas: @jessedobbelaere can you still add that to the docs in this PR?
Maybe also show the exact command they need to run to clear it in the docker container

jessedobbelaere: Yeah the docs about Docker can improve 👍 will take a look at this this evening

jessedobbelaere:
FYI I remember we introduced the `.env` file in Fork, and we moved the `.env` file of docker-compose to the docker folder. But I see these errors:

```
❯ docker-compose build
WARNING: The PHP_IDE_CONFIG variable is not set. Defaulting to a blank string.
WARNING: The XDEBUG_IDEKEY variable is not set. Defaulting to a blank string.
WARNING: The XDEBUG_PORT variable is not set. Defaulting to a blank string.
WARNING: The DOCKER_HOST_IP variable is not set. Defaulting to a blank string.
```

and with a `printenv` I could see that they are not set. I cleaned up the docker-compose file again which should make it work again now 👍

docker-compose.yml
jessedobbelaere: Pushed some changes to the docs 👍

jessedobbelaere: I also tried out [the new Mutagen-based caching](https://github.com/docker/for-mac/issues/2707#issuecomment-631606521) in the latest Docker for Mac (Edge!) which speeds up things a lot, so I could get rid of my Laravel Valet/Symfony server setup. 

One of the things mentioned on the [Mutagen-based caching page](https://docs.docker.com/docker-for-mac/mutagen-caching/#avoiding-synchronizing-a-subdirectory) is:

> **Avoiding synchronizing a subdirectory**
> Although two-way file sync is suitable for many types of files, sometimes containers can generate lots of data which doesn’t require copying to the host, for example, debug logs. If your project has a subdirectory that doesn’t need to be continuously copied back to the host, then use a named docker volume to bypass the sync.

Which now works well because the `var` folder (logs/cache) is already "ignored" by Mutagen caching because it's on a named volume 👍

PR #3122, opened 2020-05-27
Add new layout pages blocks

Katrienvh:
## Type

- Enhancement

## Pull request description

Made the pages block more attractive based on a design of @sarahmey.

In screenshot below you see the new layout. With an example of a block that is invisible and the hover state of the arrows left.

image


carakas:
@Katrienvh can you show a screenshot of a 4 column layout?

src/Backend/Core/Layout/Sass/_bootstrap-variables.scss
```diff
@@ -934,10 +935,10 @@ $toast-header-border-color:         rgba(0, 0, 0, .05);
 
 // Badges
 
-$badge-font-size:                   75%;
+$badge-font-size:                   $font-size-base * 0.75;
```
sarahmey, line 938: I would leave this in percentage because a badge is a span-tag, and is 75% of the font-size of the tag that it is in. P.ex. a badge in a title is larger than a badge in a p-tag, which looks better than badges that are always `$font-size-base * 0.75`

src/Backend/Core/Layout/Sass/components/_badges.scss
```diff
@@ -7,20 +7,27 @@
 
 .strength-none.badge {
```
sarahmey, line 8: I find it a bit odd that there are seperate badge-names for this. Wouldn't it be easier to use `.badge-danger`, `.badge-warning`,...?

Katrienvh: I am agree with this, but I think it wasn't that easy cause of the way it's implemented, but I will check this to be sure.

src/Backend/Core/Layout/Sass/_bootstrap-variables.scss
Katrienvh: valid point

freshface:
Wouldn't it be better to put the edit icon also in the top-right next to the hide icon?
So the non-invasive hide/edit actions are next to each other and the delete is stand alone.

src/Backend/Modules/MediaLibrary/Layout/Templates/MediaFolderTree.html.twig
```diff
@@ -6,7 +6,12 @@
     </button>
     <div class="form-group">
       <label for="pagetree-search" class="sr-only">{{ 'lbl.search'|trans|ucfirst }}</label>
-      <input id="pagetree-search" name="pagetree-search" value="" maxlength="255" type="text" class="form-control form-control-sm fork-form-text js-tree-search" placeholder="{{ 'lbl.Search'|trans|ucfirst }} {{ 'lbl.MediaFolder'|trans }}">
+      <div class="input-group input-group-sm">
```
carakas, line 9: why are there media library changes in here?

carakas: @Katrienvh ?

carakas:
> Wouldn't it be better to put the edit icon also in the top-right next to the hide icon?
> So the non-invasive hide/edit actions are next to each other and the delete is stand alone.
@freshface 
I think from a user perspective that will be less convenient, the delete button also always shows an modal first so I don't see any issues with it being next to eachother

src/Backend/Modules/MediaLibrary/Layout/Templates/MediaFolderTree.html.twig
Katrienvh: This is the tree view from the media library, I made the same changes then in the pages tree

Katrienvh:
> @Katrienvh can you show a screenshot of a 4 column layout?

image

PR #3137, opened 2020-06-26
More security fixes

carakas:
## Type

- Security

My mailbox overflowing by pen testers that all report a different variant of the same issue

## Pull request description

First of all, this is not a huge security issue as long as the people that have access to the backend can be trusted.
But since you never know when you have somebody trying out some nasty stuff or using an unsafe password and someone else gains access to an account with a limited set of rights this PR will prevent them from trying to steal admin passwords or executing code as a different user through stored xss

Also changed the default of serialised data not allowing any classes in it since it can be abused as well

- [x] TODO: update the node package after https://github.com/mathiashelin/bootstrap-tagsinput/pull/1 has been merged


src/Backend/Modules/ContentBlocks/Domain/ContentBlock/ContentBlockDataGrid.php
```diff
@@ -23,7 +23,7 @@ public function __construct(Locale $locale)
              WHERE i.status = :active AND i.language = :language',
             ['active' => Status::active(), 'language' => $locale]
         );
-
+        $this->setColumnFunction('htmlspecialchars', ['[title]'], 'title', false);
```
jessedobbelaere, line 26: Hmm seems easy to forget, I wonder if it can automatically apply this function somehow 👀

carakas: it is indeed, but we also have fields that have html in it. And you only need it on fields that have user input.
Do you have a suggestion on how to make this easier?
We could always add an extra method of colums that you can pass that need to be escaped and then run this internally?

jessedobbelaere: Yeah not sure, I'd leave it like this for now given it's a security fix. 
Maybe in future `htmlspecialchars` could be applied by default on datagrids and have a `raw` function you can apply on the few html-safe columns. Like twig has as a `raw` modifier 👀

carakas: it would probably be safer to change the datagrid implementation in the next mayor away from spoon :p

Issue #3150, opened 2020-07-18
Uploaded images are rotated incorrectly due to Fine Uploader bug


jessedobbelaere:
## Bug Report

| Q            | A     |
| ------------ | ----- |
| Critical Bug | no    |
| Version      | 5.8.3 |

### Problem description

When you upload a series of images coming from a mobile device, a lot of them are rotated incorrectly. These images contain EXIF data that defines the rotation. 

### Steps to reproduce

Given an image coming from an iOS device like this image (zipped it because exif data is removed otherwise): https://share.jessedobbelae.re/IT3dNg. 

![Screenshot 2020-07-18 at 18 38 16](https://user-images.githubusercontent.com/1352979/87857294-ec151900-c925-11ea-8d1c-92c4fdafd504.png)

![Screenshot 2020-07-18 at 18 39 11](https://user-images.githubusercontent.com/1352979/87857305-09e27e00-c926-11ea-988e-917844f145d3.png)

**This means that the image is rotated to the left, but the EXIF orientation data says that the image should be rotated 90 degrees. On any photo app on mac/windows it shows the image correctly in portrait mode.** 

⚠️ When you upload it to Fork CMS media module, it gets rotated twice though:

![bad-rotation](https://user-images.githubusercontent.com/1352979/87857326-34ccd200-c926-11ea-9e6f-d703545dd7bd.gif)

Since Fine Uploader strips the exif data, we cannot correctly parse the exif data via PHP and rotate the image. Moreover, on the Fine Uploader demo it also rotates incorrectly. So this seems to be a Fine Uploader bug:

![Screenshot 2020-07-18 at 18 41 22](https://user-images.githubusercontent.com/1352979/87857380-82493f00-c926-11ea-8a4c-0171e1d9546e.png)

### Expected behavior / Proposed solution

When I upload images from my computer, they should be uploaded on the website in the same orientation. Exif data should be correctly parsed and rotated if needed. 

Fine Uploader strips the EXIF data, and rotates the images 2x 90 degrees somehow. However, Fine Uploader is unmaintained since 2018 (https://github.com/FineUploader/fine-uploader/issues/2073) so ideally, we should not spend effort on fixing this bug with a workaround but move to another image upload library in javascript. 

E.g. [Dropzone.js](https://www.dropzonejs.com/) seems to be one of the most popular libs, and the online demo rotates the image correctly: 
[...]

PR #3151, opened 2020-07-19
Bugfix: Skip default photo cropping if unchecked

### LINKED ISSUE #2877: Uncropped MediaLibrary images aren't the size of the original
## Type

- Non critical bug

## Problem description

When uploading an image through the MediaLibrary you have the option to crop the image or upload it as is. When uploading as is you'd expect the image on the server to be exactly the same as the image locally, right? Wrong! It seems the MediaLibrary still uses the cropper for cropping the entire image, thus negating all pre-applied web optimizations, resulting in either greatly increased or decreased file size. I've attached 2 examples, one increasing in size and one decreasing.

Before: 392 KB
original-1
After: 444 KB
uploaded-1

Before: 372 KB
original-2
After: 77 KB
uploaded-2

## Steps to reproduce

Easiest way to reproduce this is to add an image to a blog post using the MediaLibrary (through CKEditor) and checking the size of the requested file on the frontend.

Don't alter the image when uploading it without a cropper.

jessedobbelaere:
## Type

- Non critical bugfix

Fixes #2877

## Pull request description

When we upload an image, it gets passed to the cropper by default, even if the cropper checkbox is unchecked before uploading. As Stijn mentions in #2877 when you upload his image of 392KB you end up with an image of 455KB. 
By adding an early return in the scaling config, we bypass the cropper logic and end up with an uploaded image of 392KB still ✅ 

However, the Fine Uploader gets initialized when the page loads. Therefore, when we toggle the checkbox we have to destroy the uploader dropzone and re-init fineuploader with no scaling config (if cropper is disabled) or with custom scaling config (if cropper is enabled). 

This fix will also allow for a fix for #3150 as the EXIF data is only removed when using the cropper 🙌 This means LiipImagineBundle can rotate images and strip the exif data properly. 

Screenshot 2020-07-19 at 14 21 17

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
+++ b/src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js
@@ -1038,10 +1038,16 @@ jsBackend.mediaLibraryHelper.upload = {
       --jsBackend.mediaLibraryHelper.upload.uploadedCount
       jsBackend.mediaLibraryHelper.group.validateMinimumMaximumCount()
     })
+
+    // bind change to "Enable cropper" checkbox
+    $('[data-role="enable-cropper-checkbox"]').on('change', function () {
+      // Reset the fineuploader upload box so we can skip or use a scaling config for the cropper
+      $('#fine-uploader-gallery').unbind().empty()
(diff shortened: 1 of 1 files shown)


carakas:
Nice :D Thanks Jesse

PR #3152, opened 2020-07-19
Bugfix: Fix rotation of images based on exif and strip metadata

### LINKED ISSUE #3150: Uploaded images are rotated incorrectly due to Fine Uploader bug

jessedobbelaere:
## Type

- Non critical bugfix

Fixes #3150

## Pull request description

As described in #3150 there's an issue with images getting incorrectly rotated if they contain EXIF rotation data. 

There is a dependency on the fix of https://github.com/forkcms/forkcms/pull/3151, where the cropper gets bypassed by default. This means that php still receives a source image with EXIF data. Before this fix, the source image would contain no EXIF data anymore. 

Using the `auto_rotate` and `strip` filter from LiipImagineBundle we can rotate the image based on EXIF data and then we still strip the exif metadata from the thumbnails as this contains privacy info (location, camera type, ...). The order of the filters is important here according to the LiipImagineBundle docs. 

https://symfony.com/doc/2.0/bundles/LiipImagineBundle/filters/orientation.html#auto-rotate
https://symfony.com/doc/2.0/bundles/LiipImagineBundle/filters/general.html#strip

End result: images are no longer rotated incorrectly
image

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -86,6 +86,10 @@ liip_imagine:
             quality: 95
             # list of transformations to apply (the "filters")
             filters:
+                # auto rotate the image using EXIF metadata and then strip the metadata (privacy)
+                auto_rotate: ~
+                strip: ~
+
                 ## Using the "thumbnail + background" combination will crop the
                 ## the centered image to create your required width/height.
                 # Use the "outbound" mode to crop the image when the size ratio of the input differs
@@ -93,15 +97,31 @@ liip_imagine:
                 background: { size : [140, 90], position : center }
         media_library_lightbox_small:
             filters:
+                # auto rotate the image using EXIF metadata and then strip the metadata (privacy)
+                auto_rotate: ~
+                strip: ~
+
                 scale: { dim : [800, 600] }
         media_library_lightbox_large:
             filters:
+                # auto rotate the image using EXIF metadata and then strip the metadata (privacy)
+                auto_rotate: ~
+                strip: ~
+
                 scale: { dim : [1600, 800] }
         media_library_slider_pano:
             filters:
+                # auto rotate the image using EXIF metadata and then strip the metadata (privacy)
+                auto_rotate: ~
+                strip: ~
+
                 scale: { dim : [1600, 600] }
         media_library_one_image_large:
             filters:
+                # auto rotate the image using EXIF metadata and then strip the metadata (privacy)
+                auto_rotate: ~
+                strip: ~
+
                 thumbnail: { size : [1600, 500], mode : inbound }
 
 services:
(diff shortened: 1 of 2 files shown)


carakas:
Is there a way to prevent needing to add these to every filter?

jessedobbelaere:
Unfortunately no, I didn't find a workable solution on "re-use" or "chain" filters, e.g. https://github.com/liip/LiipImagineBundle/issues/205 has no solution. Other sources on github search (e.g. this one https://github.com/habana-tech/kuba-reisen/blob/bf194b8455a3ea05a52c790a6d8db8eef938a149/config/packages/liip_imagine.yaml) also repeat them for every filter 😖

carakas:
Maybe best to then also add it to the media library documentation

jessedobbelaere:
Yeah, added some lines about those two filters in the docs 👍 

FYI There's still an improvement to someday change our image uploader lib (fine uploader) to a different one as it's an unmaintained project and the cropper still does not take the exif rotation data into account. But as long as you don't use the cropper it works now at least 😉 ...

image

PR #3157, opened 2020-07-29
Bugfix move blocks between positions

tijsverkoyen:
## Type

- Critical bugfix

## Pull request description

In the [use-new-sortable-js](https://github.com/forkcms/forkcms/tree/use-new-sortable-js-plugin)-branch there was an issue when you moved blocks from one position to another. In some cases the block would disappear when saving the page. 

In each position the blocks are stored with some hidden fields, but while the "position" was updated, the "index" of the field was not, so it could happen that the same "index" was multiple times in the same "position".

This PR fixes this.

### DIFF
--- a/src/Backend/Core/Js/jquery/jquery.backend.js
+++ b/src/Backend/Core/Js/jquery/jquery.backend.js
@@ -1588,8 +1588,8 @@
       var searchFor = fromBlockName
       var replaceWith = toBlockName
       if (fromBlockName !== toBlockName) {
-        searchFor += prefix + currentIndex
-        replaceWith += prefix + newIndex
+        searchFor += prefix + currentIndex + suffix
+        replaceWith += prefix + newIndex + suffix
       }
 
       return oldString.replace(searchFor, replaceWith)
--- a/src/Backend/Core/Js/jquery/jquery.backend.js
+++ b/src/Backend/Core/Js/jquery/jquery.backend.js
@@ -1585,23 +1585,32 @@
       })
     },
 
+    buildNewString: function(oldString, fromBlockName, toBlockName, currentIndex, newIndex, prefix = '', suffix = '') {
+      var searchFor = fromBlockName
+      var replaceWith = toBlockName
+      if (fromBlockName !== toBlockName) {
+        searchFor += prefix + currentIndex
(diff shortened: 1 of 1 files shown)


src/Backend/Core/Js/jquery/jquery.backend.js
```diff
@@ -1564,6 +1564,18 @@
       })
     },
 
+    getFieldIndexFromString: function(name) {
+      var chunks = name.split('_')
+
+      for(var chunk of chunks) {
+        // if the
```
carakas, line 1571: weird comment

src/Backend/Core/Js/jquery/jquery.backend.js
```diff
@@ -1573,34 +1585,60 @@
       })
     },
 
+    buildNewString: function(oldString, fromBlockName, toBlockName, currentIndex, newIndex, prefix = '', suffix = '') {
+      var searchFor = fromBlockName
```
carakas, line 1589: I would add a check that you need to give a prefix or a postfix, just replacing a number doesn't seem very safe

tijsverkoyen: The prefix and suffix are optional, in most cases the prefix will be used, the suffix not.
Can you elaborate on what is not safe? In essence this is just a refactoring from the code before to have less similar code.

src/Backend/Core/Js/jquery/jquery.backend.js
tijsverkoyen: Fixed in 849597d0b49793df4c79d9e5237e81e488430ad4

src/Backend/Core/Js/jquery/jquery.backend.js
carakas: I would add a check that checks that either a prefix or a postfix has been added.
If somebody tries to replace let's say 0 but it's also used on a different spot in the string that can give issues

tijsverkoyen: the fromBlockName is received from the Sortable event, and is something like `main`, `top`, and the currentIndex is appended, so it won't search for `0` only

PR #3158, opened 2020-07-31
Webpack 4

Katrienvh:
This pull request will introduce [ES6](https://www.w3schools.com/js/js_es6.asp) into Fork. 

| Q           | A      |
| ----------- | ------ |
| Improvement | yes    |
| RFC         | no |
| BC Break    | yes |

### Summary
While implementing … we also did some major work under the hood, mostly in the JS parts of the CMS. This also means there will be no backwards compatibility.

##### Social media plugin
In the past there was a ‘shareMenu’ function, which you could use to render share buttons for different social media platforms. We removed this functionality, as it was not well maintained anymore. More importantly it was mostly against the terms of different social media platforms. Therefore we recommend you to use the official share-widgets.

##### Removal of keyValueBox, tagsBox, multipleSelectbox 
There were some functions (keyValueBox, tagBox, multipleSelectbox) that are no longer used in the core and no longer maintained.
Instead of multipleSelectBox you can now use the [Select2](https://select2.org/) plugin, that is configured by default in Fork.
For the tagsBox there is already [Bootstrap tagsinput](https://bootstrap-tagsinput.github.io/bootstrap-tagsinput/examples/) configured.

##### urlencode and urldecode are removed
The methods Url.encode & Url.decode are removed. These functions are available in native JavaScript, so use: encodeURI() and decodeURI() if needed.

##### CKEditor js & css removed
With the new editor there is no need for the js and css related to CKEditor. If you still want to use CKEditor, which we don’t recommend, you should add the needed files yourself.

##### form utils are removed
The form util functions are removed as in most cases a global function is available. In some modules the form utils were not used, which made it pointless to keep maintaining these utility functions.
Functions that are removed:
isChecked
isEmaill
isFilled
isNumber
isURL
Example: There was a static function that checks if a checkbox is checked. This function was never used. If there is a need to check the value of a checkbox, this happens inline.

##### Theme is defined in .env
The frontend theme is set in the `.env` file as the parameter THEME. 
This is necessary for webpack to know which theme to compile. In our Symfony config we also need the theme to set the build paths of our multiple entries.

##### Double js methods
We no longer use expliciet double js methods. Some utils (examples: array, string, url) are available and used in the backend. These are also available in the frontend.
Example: The static method ‘ucfirst’ that adds a capital letter to a string. This method was below jsBackend.url.ucfirst and also below jsFrontend.url.ucfirst. Now the ucfirst is a static method in the backend, but this one is also available for the frontend.

##### No more build assets in Git
Files that are generated are no longer stored in Git. This means you won’t be able to download Fork as a zip file and upload it to your hosting provider and start.
We think the developer community has matured enough that we can expect them to know how to build the assets beforehand.
To generate the needed assets you can run:
	```npm install```
	```npm run build```

If you are developing locally you should run:
```npm install```
 Watch files and recompile whenever they change:
```npm run watch```

You can also run the watch for a certain part:
Watch frontend core: `npm run dev:frontend`
Watch frontend theme `npm run dev:frontend:theme`
Watch backend core: `npm run dev:backend`
Watch installer: `npm run dev:installer`

### Resolves the following issues
Weird location of installer js and css · Issue #3127 · forkcms/forkcms

##### Sortable JS
~~There are some conflicts with jQuery UI and Bootstrap4, more specifically the Tooltips. I have managed to remove everything that was based on jQuery UI. Only the sortable is still from jQuery.
In another branch we have convert the plugin to a new one, [SortableJS](https://sortablejs.github.io/Sortable/), which is more accessible. This branch is waiting for a [bug](https://github.com/forkcms/forkcms/pull/3157) to get tested properly, so the PR can be made. After that we can implement and convert this here.
Once jQuery UI is not loaded anymore, the issue with the tooltips will also be gone.~~
-- ISSUE FIXED --

##### Datepickers
We want to equalize datepickers. Something that works for all browsers and all devices. No exceptions. I started to add an example in this [branch](https://github.com/forkcms/forkcms/tree/use-new-datepicker). It uses a plugin that manages all our requirements and even uses native date/time pickers for mobile when available. This has to be implemented by an PHP developer in Spoon and Symfony properly.
For now the datepickers that are working with jquery ui work but do not have styling. Some others will just work because they use native date/time pickers. This needs to be fixed to fix the conflicts with jQuery UI and Bootstrap4 tooltips.

##### Select2
~~In the past we were using the default translations of Select2 for placeholder etc. I want to use the Fork translations. I have not yet managed to do this because I ran into some problems.
The only place I find an Select2 in the backend is in the pages module, by adding a block in a position, but it does not work there.
The reason it does not work there is because the select2 is positioned in a modal, that is hidden on page load. Select2 only initializes visible selects. This is an easy one to fix. Just initialize the select when the modal opens. I can’t manage to find the event that triggers the opening of the modal. The Bootstrap event `show.bs.modal` is not triggered when the modal opens, even though it’s a Bootstrap modal.~~
-- ISSUE FIXED --

##### Location add
When you enter a title, the ajax call to generateURL returns a 500

##### Editor js
Error when initializing:

### CODE REVIEW

File: src/Backend/Core/Js/Components/Forms.js [hunk 1]
```diff
@@ -0,0 +1,403 @@
...
+            if ($(event.currentTarget).prop('disabled')) {
+              return false
+            }
+            else {
+              $('form#' + formId).submit()
+            }
+          })
+
+          // dont submit the form on certain elements
+          $('form#' + formId + ' .dontSubmit').on('focus', () => {
```
jessedobbelaere, line 298: With ES6 template literals we can do this now? I personally find it more readable, not sure what you think?

```js
$(`form#${formId} .dontSubmit`).on('focus', () => {
```

src/Backend/Core/Js/Components/MultiTextBox.js
```diff
@@ -0,0 +1,296 @@
...
+        return !blockSubmit
+      })
+
+      // remove previous HTML
+      if ($('#elementList-' + id).length > 0) {
+        $('#elementList-' + id).parent('.multipleTextWrapper').remove()
+      }
+
+      // build replace html
+      const html =
```
jessedobbelaere, line 57: should be way easier to write using template literals 👀  :


```js
const html = `
    <div class="multipleTextWrapper">
    <div id="elementList-${id}" class="multipleTextList">

etc...
`;
```

src/Backend/Core/Js/Components/Resize.js
```diff
@@ -0,0 +1,29 @@
...
+
+      this.ticking = false
+    }
+
+    this.resize()
+  }
+
+  resize () {
+    $(window).on('load resize', () => {
+      console.log('RESIZE')
```
jessedobbelaere, line 18: 👀 

also, does it need a debouncer? If I resize my browser then 200+ events are easily sent

src/Backend/Core/Js/Components/StringUtil.js
```diff
@@ -0,0 +1,234 @@
...
+  static camelCase (input, splitChar) {
+    splitChar = typeof splitChar !== 'undefined' ? splitChar : '_'
+    const regex = new RegExp(splitChar + '(.)', 'g')
+    return input.toLowerCase().replace(regex, function (match, group1) {
+      return group1.toUpperCase()
+    })
+  }
+
+  /**
+   * Fix a HTML5-chunk, so IE can render it
```
jessedobbelaere, line 18: If we still support IE11? Else we might be able to drop this snippet?

src/Backend/Core/Layout/Templates/head.html.twig
```diff
@@ -8,33 +8,20 @@
...
     <script src="{{ jsFile }}"></script>{{ CRLF }}{{ TAB }}
   {% endfor %}
 
+
+  {{ encore_entry_script_tags('backend', null, 'backendConfig') }}
+  {{ encore_entry_link_tags('screen', null, 'backendConfig') }}
+
   {% if debug %}
-    <script src="http://localhost:35729/livereload.js?snipver=1"></script>
+    <script src="http://localhost:35729/livereload.js"></script>
```
jessedobbelaere, line 23: When I develop css/js for Fork themes, I use webpack dev server with hot module reloads, is that something we can use ? Not sure if it's been looked into https://symfony.com/doc/current/frontend/encore/dev-server.html

src/Frontend/Themes/undefined/Core/build/entrypoints.json
```diff
@@ -0,0 +1,3 @@
+{
+  "entrypoints": {}
```
jessedobbelaere, line 2: What is theme `undefined` ?

UPGRADE_6.0.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Introdcure ES6
+
+### Changes in Frontend
+
+#### Social media plugin
+
+In the past there was a ‘shareMenu’ function, which you could use to render share buttons for different social media platforms. We removed this functionality, as it was not maintained anymore. More importantly it was mostly against the terms of different social media platforms. Therefore we recommend you to use the official share-widgets.
+
+### Changes in backend
```
jessedobbelaere, line 12: Maybe add a section about custom modules for Fork CMS, they need to upgrade the JS files to be `index.js` instead to still get loaded automatically in the module page? Instead of having the name of the module, e.g. Instagram module would auto load `Instagram.js`

src/Backend/Core/Js/Components/Resize.js
Katrienvh: That is what the polyfill does, that I have imported at the top of the file, isn't it? https://www.paulirish.com/2011/requestanimationframe-for-smart-animating/

jessedobbelaere: Aha! Didn't look that up 😁 ✅

src/Backend/Core/Js/Components/Data.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  static get (key) {
+    // init if needed
+    if (typeof jsData === 'undefined') throw new Error('jsData is not available')
+
+    const keys = key.split('.')
+    let data = jsData
+    for (let i = 0; i < keys.length; i++) {
+      data = data[keys[i]]
```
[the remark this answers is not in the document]
Katrienvh: No plans, but feel free

src/Backend/Core/Engine/Header.php
```diff
@@ -129,7 +91,7 @@ private function buildPathForModule(string $fileName, string $module, string $su
             return '/src/Backend/Core/' . $subDirectory . '/' . $fileName;
         }
 
-        return '/src/Backend/Modules/' . $module . '/' . $subDirectory . '/' . $fileName;
+        return '/src/Backend/Modules/' . $module . '/' . $subDirectory . '/' . 'index.js';
```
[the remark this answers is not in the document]
Katrienvh: I forgot to mention that this part needs to be removed. The changes were made in the beginning when there were different plans. I would like that someone of @forkcms/backenders removes this, because the public addJS function needs to change.

UPGRADE_6.0.md
```diff
@@ -0,0 +1,87 @@
...
+#### Double js methods
+
+We no longer use expliciet double js methods. Some utils (examples: array, string, url) are available and used in the backend. These are also available in the frontend.
+
+Example: The static method ‘ucfirst’ that adds a capital letter to a string. This method was below jsBackend.url.ucfirst and also below jsFrontend.url.ucfirst. Now the ucfirst is a static method in the backend, but this one is also available for the frontend.
+
+#### No more build assets in Git
+
+Files that are generated are no longer stored in Git. This means you won’t be able to download Fork as a zip file and upload it to your hosting provider and start.
+We think the developer community has matured enough that we can expect them to know how to build the assets beforehand.
```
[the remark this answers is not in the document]
Katrienvh: Correct

UPGRADE_6.0.md
```diff
@@ -0,0 +1,87 @@
+UPGRADE FROM 5.x to 6.0
+=======================
+
+## Introdcure ES6
```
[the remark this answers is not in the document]
Katrienvh: We were not planning to drop the support for IE11 yet. It's sad the we can not use the template literals cause of that, I am also a fan of them. Since I know there is also no polyfill for this.

src/Backend/Core/Js/Backend.js
```diff
@@ -0,0 +1,99 @@
...
+    this.layout = new Layout()
+    this.tooltip = new Tooltip()
+    this.tableSequenceDragAndDrop = new TableSequenceDragAndDrop()
+    this.session = new Session()
+
+    // init block editor
+    this.blockEditor = new BlockEditor()
+
+    // init modules
+    this.analytics = new Analytics()
```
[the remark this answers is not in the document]
Katrienvh: I am not really a big fan of the if state based on an ID. Also I think is not really necessary. Once the Js is fully loaded, it is stored in the cache.

src/Backend/Core/Js/Components/Collection.js
```diff
@@ -0,0 +1,155 @@
...
+      },
+
+      initSequence: function () {
+        var $sequenceInstances = $('[data-role=collection-sequence]')
+
+        if ($sequenceInstances.length === 0) {
+          return
+        }
+
+        $.each($sequenceInstances, function (index, element) {
```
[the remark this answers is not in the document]
Katrienvh: For this one, I have not really focussed on that. I only transferred the current code and did the minimal to convert it to ES6. This does not mean I do not agree. I think it is definitely something we should do in the future.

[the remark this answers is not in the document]
jessedobbelaere: Yeah I understand, this PR is already big anyways 👍 👍

src/Backend/Core/Layout/Templates/head.html.twig
Katrienvh: I have looked at the hot module reloads but could not find a way to make it work, so I choose to do it this way. But feel free to upgrade it!

PR #3159, opened 2020-08-05
Password generator and strength meter

Katrienvh:
## Type

- Enhancement

## Pull request description

Implemented 2 external plugins. One to generate a password, another one for getting the strength of the password.

https://www.npmjs.com/package/@cityssm/simple-password-generator
https://www.npmjs.com/package/check-password-strength

These are 2 less code parts that we need to maintain.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "bootstrap-datepicker": "^1.9.0",
     "bootstrap-sass": ">=3.4.1",
     "bootstrap-tagsinput": "https://github.com/mathiashelin/bootstrap-tagsinput.git",
+    "check-password-strength": "^1.0.8",
     "ckeditor": "^4.12.1",
     "core-js": "^3.6.5",
     "cropper": "^3.0.0-rc.2",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -7,6 +7,7 @@
   "author": "Tijs Verkoyen <tijs@verkoyen.eu>",
   "license": "MIT",
   "devDependencies": {
+    "@cityssm/simple-password-generator": "^0.6.1",
```
carakas, line 10: Can you use something with a stable release?
This thing was only created 7 days ago...

StijnVrolijk: Does it really matter though? It's a random password generator, it's not like it's going to break or something.

carakas: It does, please use a different library, it is such a common issue that it can't be that hard to use one that is used by somebody else than the creator of this one

tijsverkoyen: I agree with Jelmer. 0-releases have the tendency to not being maintained. I understand that this is not a very critical part of Fork, but external packages can have bugs which are a pain in the ass to fix.

Some alternatives:
- https://www.npmjs.com/package/password-generator, seems like its being maintained, has tests.
- https://www.npmjs.com/package/generate-password

Katrienvh: I updated the plugin

StijnVrolijk (approved):

PR #3160, opened 2020-08-05
Replace SpoonFileCSV

tijsverkoyen:
## Type

- Enhancement

* Replaces https://github.com/forkcms/forkcms/pull/2687
* Resolved https://github.com/forkcms/forkcms/issues/2099

## Pull request description

SpoonFileCSV is deprecated so it should be replaced with something that is more maintained. 

With this PR we introduce the usage of [PhpSpreadsheet](https://phpspreadsheet.readthedocs.io). I choose this one over `league/csv`, as the later one does not really have a lot of features, and does not really integrate in the Symfony responses, per example: outputting a CSV file won't return a string, but just outputs the CSV and stop the code.

### Writing CSV files

So instead of working with arrays you should create a SpreadSheet instance first,
check the [official documentation](https://phpspreadsheet.readthedocs.io/en/latest/topics/creating-spreadsheet/).

The easiest way to convert existing code is:

```php
$spreadSheet = new Spreadsheet();
$sheet = $spreadSheet->getActiveSheet();
$sheet->fromArray($headers, null, 'A1');
$sheet->fromArray($rows, null, 'A2');
```

A service called `ForkCMS\Utility\Csv\Writer` is added, which allows you to force the download of a spreadsheet in CSV-format thru a Symfony StreamedResponse. If you want to force a download you can use the code below in the backend:

```php
    throw new RedirectException(
        'Return the csv data',
        $this->get(Writer::class)
            ->forBackendUser(Authentication::getUser())
            ->output($spreadSheet, date('Ymd_His') . '.csv')
    );
```

### DIFF
--- a/src/Backend/Core/Engine/Csv.php
+++ b/src/Backend/Core/Engine/Csv.php
@@ -11,7 +11,7 @@ use Symfony\Component\HttpFoundation\StreamedResponse;
 /**
  * @deprecated remove this in Fork 6, just use ForkCMS\Utility\Csv\Writer
  */
-class Csv
+class Csv extends \SpoonFileCSV
 {
     /**
      * Output a CSV-file as a download
--- a/src/Backend/Core/Engine/Csv.php
+++ b/src/Backend/Core/Engine/Csv.php
@@ -8,10 +8,15 @@ use PhpOffice\PhpSpreadsheet\Spreadsheet;
 use PhpOffice\PhpSpreadsheet\Writer\Csv as CsvWriter;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
+/**
+ * @deprecated remove this in Fork 6, just use ForkCMS\Utility\Csv\Writer
+ */
 class Csv
 {
     /**
      * Output a CSV-file as a download
+     *
+     * @deprecated remove this in Fork 6, just use ForkCMS\Utility\Csv\Writer->output()
      *
      * @param string $filename       The name of the file.
      * @param array  $array          The array to convert.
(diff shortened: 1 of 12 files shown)


src/ForkCMS/Utility/Csv/Writer.php
```diff
@@ -0,0 +1,81 @@
...
+                }
+            }
+        }
+
+        return $writer;
+    }
+
+    public function forBackendUser(User $user): self
+    {
+        $this->options['Delimiter'] = $user->getSetting('csv_split_character');
```
carakas, line 50: you need to cone the instance here and return the clone since using it after calling the forBackendUser will still run it with those changes otherwise

src/Backend/Core/Engine/Csv.php
```diff
@@ -4,20 +4,24 @@
...
-use Symfony\Component\HttpFoundation\Response;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Writer\Csv as CsvWriter;
+use Symfony\Component\HttpFoundation\StreamedResponse;
 
 /**
- * This is our extended version of SpoonFileCSV
+ * @deprecated remove this in Fork 6, just use ForkCMS\Utility\Csv\Writer
  */
-class Csv extends \SpoonFileCSV
```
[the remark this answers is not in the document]
tijsverkoyen: Fixed in a142503c07e51c3a152c5aac391aede0715f6ada

src/ForkCMS/Utility/Csv/Writer.php
tijsverkoyen: fixed in f941db102a5cefc147be9ed7c784c88e8f6a7f73

PR #3166, opened 2020-08-19
Make it possible to set an id in a usertemplate

carakas:
That way you can create anchor links

## Type

- Feature

## Pull request description

Add the option to set an id in the usertemplate making it possible to create anchor tags
`<hr id="placeholder" data-ft-label="id" data-ft-type="anchor">`

### DIFF
--- a/docs/03. user guide/03. pages.md
+++ b/docs/03. user guide/03. pages.md
@@ -79,6 +79,7 @@ currently supported:
 - editor
 - image
 - image-background
+- anchor
 
 ### Example configuration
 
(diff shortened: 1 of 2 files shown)


docs/03. user guide/03. pages.md
```diff
@@ -107,5 +108,6 @@ currently supported:
     <div data-ft-label="editor" data-ft-type="editor">Dit is de editor</div>
     <img src="https://placehold.it/128x128" alt="Placeholder" class="img-responsive img-circle" data-ft-label="Avatar" data-ft-type="image" />
     <div class="image-holder" data-src="https://placehold.it/200x200" data-ft-label="Background image" data-ft-type="image-background" data-ft-block-optional="true" style="background-image: url('https://placehold.it/200x200');"></div>
+    <hr id="placeholder" data-ft-label="id" data-ft-type="anchor">
```
Katrienvh, line 111: Is there a reason why u use and hr-element as example? I should use an div. Id's are mostly add to wrapping elements, so that way a div is more logic.

PR #3167, opened 2020-08-19
Allow html in user template textareas

carakas:
## Type

- Non critical bugfix

## Pull request description

At the moment you can't add html when using a textarea in a usertemplate, for instance to paste an embed code.
I fixed this by enabling the html.
Another solution would be to create a new type: html. Feedback and opinions welcome

### DIFF
--- a/src/Backend/Modules/Pages/Js/Pages.js
+++ b/src/Backend/Modules/Pages/Js/Pages.js
@@ -901,7 +901,7 @@ jsBackend.pages.extras = {
     }
 
     if ($element.is('[data-ft-type="textarea"]')) {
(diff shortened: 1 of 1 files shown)


Katrienvh:
You can add embed video's and other types of embeds in the editor. I think the editor gives more then enough options to add embed code:
- Embed media from external sites
- Iframe
- Source -> to add embed code yourself

So I am not 100% sure this is a bug, maybe a feature. But not sure we need it. Either way if you want to add this with a textarea I would add an extra type of textarea. So it's not always allowed.

tijsverkoyen:
I agree with @Katrienvh

carakas:
The reason why I fixed this, but a separate type is also oke, is because if you use the editor you need to click on source, and when you edit it you don't see the code and that was confusing for clients

PR #3171, opened 2020-08-26
Use fork translations for the text used in select2

Katrienvh:
## Type

- Enhancement

## Pull request description

In the past we were using the translation files from select2. We are talking about one file per language. Al these files were in git. I changed it so we can use the translations in fork. No more extra files and we can also easily changes the values.


StijnVrolijk (approved):

carakas (approved):

carakas:
I didn't see you remove the files here, did you do that in the webpack branch?

Katrienvh:
@carakas Indeed these were already removed :)

PR #3175, opened 2020-09-07
Move from Travis to GitHub Actions

jessedobbelaere:
## Type

- Enhancement

## Pull request description

Talked to @carakas about the new [Github Container Registry](https://github.blog/2020-09-01-introducing-github-container-registry/) and was wondering if we should try Github actions or stay with Travis. There were some issues (?) in the past with Travis so I could give it a try.

Example:
https://github.com/jessedobbelaere/forkcms/actions/runs/325952756
image

--- 

Some notes: ⚠️ 
* Had to upgrade `symfony/phpunit-bridge` to * as documented by Symfony, because we pinned a pretty old version (v3 instead of v5). Causing our tests to run on phpunit 5 instead of 7 on travis and locally...

image

* I found the PHPunit tests to be **very slow** on both travis and github actions. **Reason is xdebug**. Removing xdebug makes our tests run ~15min28 → 1.32 minutes (!!!). But we need xdebug for code coverage... However, we can switch to [pcov](https://dev.to/swashata/setup-php-pcov-for-5-times-faster-phpunit-code-coverage-3d9c), which does not really slow down tests. Pcov is built into phpunit 8, which we don't support yet, since we support PHP 7.1 and phpunit-bridge takes PHPUnit 7.5 by default then. If you are on PHPUnit 7, you can install https://github.com/krakjoe/pcov-clobber but this does not seem to work with the phpunit that phpunit-bridge downloads.. Only works with real phpunit installations...

image

image
image

**On that note, do we have plans to move to PHP 7.2 or 7.3 as lowest supported version for Fork CMS**? Needless to say, PHP 7.1 is EOL and PHP 7.2 very soon too: https://www.php.net/supported-versions.php

* We need to add a token `CR_PAT` to the Github actions secrets, as mentioned on [authenticating with github container registry](https://docs.github.com/en/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry). It should have read/write rights on repo packages. When we create a new release, it will push a docker container with the version tag and `:latest` tag to the container registry. This can be used for caching etc. 

* We need to add `SCRUTINIZER_ACCESS_TOKEN` to the github action secrets, which we can probably take from Travis.

* ~On travis there was a `yarn test` job in the `after_tests` **but it always failed**. Adding it back on github actions, it will fail the job so guess we need to fix the errors or decide to throw it out for now? :(  I commented it out for now.~  Fixed!

* Sadly, Github actions don't support [Yaml anchors](https://github.community/t/support-for-yaml-anchors/16128) yet, or other ways to re-use blocks of code. This causes some code duplication in every job (define MySQL service, add PHP action, install dependencies, ...)

* Added the `unzip` package to Dockerfile, since composer complained about it during docker build: 

> As there is no 'unzip' command installed zip files are being unpacked using the PHP zip extension.
> This may cause invalid reports of corrupted archives.

* Needed to fix a PHPUnit 7 issue in InstallerControllerTest.php


jessedobbelaere:
Some fun automations possible for later:

E.g. greeting a first contributor, auto labelling in github, ... https://github.com/actions/starter-workflows/tree/master/automation

image

carakas:
Nice!
I'll have a look at it later.

jessedobbelaere:
I'm downgrading to Composer v1, since we run in Fork CMS 5.x on a minimum version of PHP 7.1. But Doctrine requires `ocramius/package-versions` which depends on composerv1. There's composerv2 support in `ocramius/package-versions` 1.8+ but that requires PHP 7.4 or higher. 

image

jessedobbelaere:
I also had to pin the composer version to v1 in our Dockerfile since it was also fetching the latest composer version ✅

carakas:
Fork 6 is going to be either php 8 or 7.4

carakas (approved):

PR #3177, opened 2020-09-23
Link class / Bugfix add page

wouterverstuyf:
## Type
Critical bugfix

When adding a page (with no-god user) an error is thrown.

## Pull request description
Only save the link class field if the user is a god user.


carakas:
We merged a different version of this patch since that one included it with the other god user settings

PR #3188, opened 2020-10-12
Add canonical URL to SEO tab

jacob-v-dam:
## Type
- Enhancement

None

## Pull request description

It would be nice to have some control over the canonical URL for SEO purposes. With this commit you could change the page canonical URL to everything you want, even other websites. It also prevents duplicated content and provides the search engines with the right pages.


src/Backend/Core/Installer/Data/locale.xml
```diff
@@ -10361,6 +10361,24 @@
...
+        <translation language="ru"><![CDATA[canonical URL]]></translation>
+        <translation language="it"><![CDATA[canonical URL]]></translation>
+        <translation language="zh"><![CDATA[canonical URL]]></translation>
+        <translation language="de"><![CDATA[canonical URL]]></translation>
+        <translation language="fr"><![CDATA[canonical URL]]></translation>
+        <translation language="es"><![CDATA[canonical URL]]></translation>
+        <translation language="lt"><![CDATA[canonical URL]]></translation>
+        <translation language="uk"><![CDATA[canonical URL]]></translation>
+        <translation language="sv"><![CDATA[canonical URL]]></translation>
+        <translation language="el"><![CDATA[canonical διεύθυνση URL]]></translation>
```
carakas, line 10377: mistake in the translation?
Translations always fall back to english if the translation can't be found.
So I'd rather have the translations missing than containing an untranslated string

src/Common/Doctrine/Entity/Meta.php
```diff
@@ -79,6 +79,20 @@ class Meta
      */
     private $urlOverwrite;
 
+    /**
+     * @var string
+     *
+     * @ORM\Column(type="string", length=255, name="canonical_url", nullable=true)
+     */
+    private $canonicalUrl;
```
carakas, line 87: this is only allowed in version 6 because this is a database change. You can always implement it in the metadata and take it out of there upon serialisation.
Otherwise you need to change the base of the PR to restructure-core

carakas (changes requested):

src/Common/Doctrine/Entity/Meta.php
jacob-v-dam: This is changed. When this is merged into master I will create a new PR to restructure-core.

src/Backend/Core/Engine/Meta.php
```diff
@@ -225,6 +225,36 @@ public function getUrlOverwrite(): ?bool
...
+        }
+
+        // return value
+        return urldecode($this->data['data']['canonical_url']);
+    }
+
+    public function getCanonicalUrlOverwrite(): ?bool
+    {
+        if (!is_array($this->data['data'])) {
+            return null;
```
carakas, line 246: This should return false

src/Backend/Core/Engine/Meta.php
```diff
@@ -225,6 +225,36 @@ public function getUrlOverwrite(): ?bool
...
+        }
+
+        // not set so return null
+        if (!array_key_exists('canonical_url_overwrite', $this->data['data'])) {
+            return null;
```
carakas, line 251: This should return false

src/Backend/Core/Engine/Meta.php
```diff
@@ -225,6 +225,36 @@ public function getUrlOverwrite(): ?bool
...
+        // not set so return null
+        if (!array_key_exists('canonical_url', $this->data['data'])) {
+            return null;
+        }
+
+        // return value
+        return urldecode($this->data['data']['canonical_url']);
+    }
+
+    public function getCanonicalUrlOverwrite(): ?bool
```
carakas, line 243: Just bool, not nullable

carakas (changes requested):

carakas (approved):

PR #3192, opened 2020-10-19
Add edit button to media item within a form

jacob-v-dam:
It is nice to edit the title of a media item within a from. When
there are a large amount of files it is hard to find the right file.
With this addition it is possible to click the edit pencil and change
the title.

## Type

- Enhancement

## Pull request description

With this function it would be able to change the title of a media item within a form. This means that the user doesn't have to search the whole library to find one item.


carakas:
@jacob-v-dam can you look at the feedback of the failing yarn test?

jacob-v-dam:
Yarn tests are added. I see that codecov is failing right now, but only the storage manager is tested.

carakas (approved):

PR #3195, opened 2020-10-21
Fix calling instance method as static method

idetinkin:
## Type

- Non critical bugfix

### DIFF
--- a/src/Backend/Modules/Analytics/Tests/Actions/IndexTest.php
+++ b/src/Backend/Modules/Analytics/Tests/Actions/IndexTest.php
@@ -9,7 +9,7 @@ class IndexTest extends BackendWebTestCase
 {
     public function testAuthenticationIsNeeded(Client $client): void
     {
-        self::assertAuthenticationIsNeeded($client, '/private/en/analytics/index');
+        $this->assertAuthenticationIsNeeded($client, '/private/en/analytics/index');
     }
 
     public function testRedirectToSettingsActionWhenTheAnalyticsModuleIsNotConfigured(Client $client): void
--- a/src/Frontend/Core/Engine/Url.php
+++ b/src/Frontend/Core/Engine/Url.php
@@ -257,7 +257,7 @@ class Url extends KernelLoader
     private function setLanguageCookie(string $language): void
     {
         try {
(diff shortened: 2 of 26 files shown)


src/Frontend/Core/Engine/Url.php
```diff
@@ -257,7 +257,7 @@ private function determineLanguage(string $queryString): string
     private function setLanguageCookie(string $language): void
     {
         try {
-            self::getContainer()->get('fork.cookie')->set('frontend_language', $language);
+            $this->getContainer()->get('fork.cookie')->set('frontend_language', $language);
```
jessedobbelaere, line 260: Actually, when we switch to Symfony 4 we need to get the container again from a static property... https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing

src/Backend/Modules/Analytics/Tests/Actions/IndexTest.php
```diff
@@ -9,7 +9,7 @@ class IndexTest extends BackendWebTestCase
 {
     public function testAuthenticationIsNeeded(Client $client): void
     {
-        self::assertAuthenticationIsNeeded($client, '/private/en/analytics/index');
+        $this->assertAuthenticationIsNeeded($client, '/private/en/analytics/index');
```
jessedobbelaere, line 12: 👍  I like instance methods for phpunit, like Symfony describes in the docs https://symfony.com/doc/current/testing.html#unit-tests

src/Frontend/Modules/Tags/Tests/Engine/ModelTest.php
```diff
@@ -173,6 +173,6 @@ private function assertModuleTags($items): void
         self::assertArrayHasKey('name', $items);
         self::assertArrayHasKey('label', $items);
         self::assertArrayHasKey('items', $items);
```
jessedobbelaere, line 175: But if we change all the static methods to instance methods for phpunit, we should change them all, not leave the tests in a half state of mixed statics with instance methods...

carakas:
I'd rather move them all to static in Fork 6

carakas:
Closing since we'll have a look at it in restructure code

PR #3202, opened 2020-10-26
Allow switching in the backend to a language that is disabled in the frontend

carakas:
## Type

- Non critical bugfix

## Pull request description

It was impossible to get a backend url for a language that was disabled in the frontend. This made it hard to prepare a language before enabling it in the frontend


jessedobbelaere (approved):
👌

PR #3208, opened 2020-10-31
Fix scrutinizer & coverage upload

jessedobbelaere:
## Type

- Enhancement

### On the topic of static code analysis
* Scrutinizer seems to rely on a global config we hardcoded in the settings of the scrutinized admin panel. It seemed a bit outdated... I added the `.scrutinizer.yml` in this repo, and cleaned it up a bit. Having it in the repo itself might help to keep it more up to date? Are we still sure we benefit from scrutinizer? I actually never looked at it in the past years, but I can't speak for others 🤔  I'd be more in favour of improving our **PHPStan** integration, which seems to be the more modern approach of static code analysis for PHP? But want to hear opinions to keep or remove it?
* I removed phpcs from scrutinizer's config since we already have that in our Github Actions.
* Adding phpstan github annotations: https://twitter.com/OndrejMirtes/status/1278262546819100684

### On the topic of code coverage
Scrutinizer is being a **pain**. It never seems to receive the coverage report we sent, even though it waits a long time for it, and the `--verbose` logs say that we successfully uploaded the file. Did numerous attempts. I have previous experience with the widely known coverage tool https://codecov.io and that seemed to work out of the box. I got it working in just a few minutes, by using their nice [Github workflow action](https://github.com/codecov/codecov-action) ready for use. I think there's a lot of benefits to codecov. It allows us to set a `.codecov.yml` with rules, like any PR cannot decrease coverage with more than 1%. And it allows us to browse the code on github with the [codecov browser extension](https://docs.codecov.io/docs/browser-extension) and see the source code of fork cms colored with red or green to easily spot code with no tests.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,59 @@
+build:
+    nodes:
+        analysis:
+            environment:
+                php:
+                    version: 7.1
+            cache:
+                disabled: false
+                directories:
+                    - ~/.composer/cache
+            project_setup:
+                override: true
+            tests:
+                override:
+                    - php-scrutinizer-run
+
+filter:
+    excluded_paths:
+        - app/*
+        - bin/*
+        - docs/*
+        - src/Backend/Core/Js/ckfinder/*
+        - src/Backend/Core/Js/ckeditor/*
+        - var/*
+        - vendor/*
+
+tools:
+    external_code_coverage:
+        timeout: 3600
+        runs: 3 # Scrutinizer will wait for 3 code coverage submissions
+    php_mess_detector: true
+    sensiolabs_security_checker: true
+    php_code_coverage: true
+    php_cpd: false
+    php_sim: true
+    php_pdepend:
+        excluded_dirs:
+            - vendor
+
+application:
+    symfony2:
+        autoload_file: autoload.php
+    dependencies:
+        before:
+            - 'composer install --prefer-dist'
+
+checks:
+    php:
+        unused_variables: true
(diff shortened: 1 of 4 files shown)


carakas:
I actually look at it when I look for things to refactor or when I'm adding new code that it actually gets the A score

carakas:
The generated links for fork give a 502 error

jessedobbelaere:
> The generated links for fork give a 502 error

Yeah not the best first impression for you 😀 It seems to affect other repo's as well, e.g. https://github.com/mirumee/saleor (badge in readme) and https://codecov.io/github/mirumee/saleor?branch=master . I expect it to be fixed soon

.scrutinizer.yml
```diff
@@ -25,12 +25,8 @@ filter:
         - vendor/*
 
 tools:
-    external_code_coverage:
-        timeout: 3600
-        runs: 3 # Scrutinizer will wait for 3 code coverage submissions
```
carakas: does scrutinizer still hang after removing this one?

jessedobbelaere: No it just does not expect coverage now... Just runs static analysis and PHP mess detector etc :) 
image

jessedobbelaere:
Link (https://codecov.io/gh/forkcms/forkcms/pull/3208) is working again @carakas , they had an interruption it seems when I check their status page. However, there's no base report yet (master branch) to compare against, so not that much to be seen on the codecov dashboard yet :-)

PR #3216, opened 2020-11-06
Remove ckeditor from packages

Katrienvh:
## Type

- Non critical bugfix

## Pull request description

Ckeditor is not used anymore we forgot to remove the packages, now they are.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,11 +7,6 @@
   "author": "Tijs Verkoyen <tijs@verkoyen.eu>",
   "license": "MIT",
   "devDependencies": {
-    "@editorjs/editorjs": "^2.16.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,11 +7,6 @@
   "author": "Tijs Verkoyen <tijs@verkoyen.eu>",
   "license": "MIT",
   "devDependencies": {
-    "@editorjs/editorjs": "^2.16.1",
```
carakas, line 10: Why did you remove this as well?

Katrienvh:
Totally removed the wrong ones, started over here: https://github.com/forkcms/forkcms/pull/3217

PR #3224, opened 2020-11-12
Remove editor js import in php

Katrienvh:
## Type

- Drop feature request

## Pull request description

We do not use JS files that are stored in a vendor file. We know load all the JS plugins by importing them in de ES6 components.
So I removed the AddJS functions from the PHP.


carakas (changes requested):
Don't remove it, make it something optional. A standalone module needs to be able to load the required js if it implements a custom editor type

carakas:
Closed since the issue has been addressed in #3233

PR #3225, opened 2020-11-12
Fix tests

carakas:
## Type

- Critical bugfix

The tests are broken, you may now panic

## Pull request description

Fixing ALL the tests (hopefully)


src/Backend/Modules/MediaLibrary/Domain/MediaGroupMediaItem/Exception/MediaGroupMediaItemNotFound.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Backend\Modules\MediaLibrary\Domain\MediaItem\Exception;
+
+class MediaGroupMediaItemNotFound extends \Exception
+{
+    public static function forMediaItemId(string $id): self
+    {
+        return new self('Can\'t find a MediaGroupMediaItem with mediaItemId = "' . $id . '"".');
```
ohvitorino, line 9: String interpolation would look better here.

src/Backend/Modules/MediaLibrary/Domain/MediaItem/Type.php
```diff
@@ -61,6 +61,8 @@ public static function fromMimeType(string $mimeType): self
                 return self::fromString($mediaItemType);
             }
         }
+
+        throw new Exception('MimeType could not be resolved');
```
ohvitorino, line 65: Maybe use a less generic exception?

src/Backend/Modules/MediaLibrary/Domain/MediaItem/Type.php
```diff
@@ -87,6 +89,8 @@ public static function fromExtension(string $extension): self
                 return self::fromString($mediaItemType);
             }
         }
+
+        throw new Exception('Extension could not be resolved');
```
ohvitorino, line 93: Maybe use a less generic exception?

src/Frontend/Core/Engine/Url.php
```diff
@@ -232,27 +232,27 @@ private function determineLanguage(string $queryString): string
...
+            // default browser language
+            $language = Language::getBrowserLanguage();
+            $this->setLanguageCookie($language);
+            $this->redirectToLanguage($language);
         }
 
-        // default browser language
-        $language = Language::getBrowserLanguage();
+        // define language
+        $language = (string) $chunks[0];
```
ohvitorino, line 250: Maybe you're forgetting a return statement in the previous if statement? Otherwise `$chunks[0]` can still be empty here.

src/Backend/Core/Engine/Base/Action.php
```diff
@@ -263,8 +263,8 @@ public function getContent(): Response
      *
      * @throws RedirectException
      */
-    public function redirect(string $url, int $code = Response::HTTP_FOUND): void
+    public function redirect(string $url, int $code = Response::HTTP_FOUND)
```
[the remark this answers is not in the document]
carakas: forgot to put it back, was trying to solve an issue with phpstan

src/Frontend/Core/Engine/Url.php
carakas: Forgot to revert this. The change was part of making phpstan understand our redirects but instead I confused you :p

src/Backend/Modules/MediaLibrary/Domain/MediaItem/Type.php
carakas: Used InvalidArgument since it is about a string and not a specific class

src/Backend/Modules/MediaLibrary/Domain/MediaGroupMediaItem/Exception/MediaGroupMediaItemNotFound.php
carakas: Used InvalidArgument since it is about a string and not a specific class

src/Backend/Modules/MediaLibrary/Domain/MediaItem/Type.php
carakas: Used InvalidArgument since it is about a string and not a specific class

PR #3226, opened 2020-11-13
Bugfix consent dialog

tijsverkoyen:
## Type

- Non critical bugfix

## Pull request description

When saving the setting a level 'functional' was stored in the settings. Which is not needed, as the level is there per default.
I also fixed a bug when no levels are defined, this was stored as an empty string.

### DIFF
--- a/src/Backend/Modules/Settings/Actions/Index.php
+++ b/src/Backend/Modules/Settings/Actions/Index.php
@@ -645,7 +645,7 @@ class Index extends BackendBaseActionIndex
                 );
                 $privacyConsentLevels = [];
                 if($privacyConsentLevelsField->isFilled()) {
-                    $privacyConsentLevels = explode(',', $this->form->getField('privacy_consent_levels')->getValue());
+                    $privacyConsentLevels = explode(',', $privacyConsentLevelsField->getValue());
                 }
                 $this->get('fork.settings')->set(
                     'Core',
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Settings/Actions/Index.php
```diff
@@ -643,10 +643,14 @@ private function validateForm(): void
                     'show_consent_dialog',
                     $this->form->getField('show_consent_dialog')->getChecked()
                 );
+                $privacyConsentLevels = [];
+                if($privacyConsentLevelsField->isFilled()) {
+                    $privacyConsentLevels = explode(',', $this->form->getField('privacy_consent_levels')->getValue());
```
carakas, line 648: You can use $privacyConsentLevelsField here

carakas (changes requested):

src/Backend/Modules/Settings/Actions/Index.php
tijsverkoyen: Fixed in b31a09ce909c0486bffd97eacf51274cf23539fd

src/Backend/Modules/Settings/Actions/Index.php
```diff
@@ -643,10 +643,14 @@ private function validateForm(): void
                     'show_consent_dialog',
                     $this->form->getField('show_consent_dialog')->getChecked()
                 );
+                $privacyConsentLevels = [];
+                if($privacyConsentLevelsField->isFilled()) {
```
carakas, line 647: codestyles also dictate that there needs to be a space after if

carakas (changes requested):

src/Backend/Modules/Settings/Actions/Index.php
tijsverkoyen: fixed in 556978323d3cf8a84eeddf9db3bc4985d4696522

PR #3210, opened 2020-11-01
Upgrade phpstan to latest version

jessedobbelaere:
## Type

- Enhancement

## Pull request description

Latest PHPStan supports github annotations on pull requests for better feedback:
https://twitter.com/OndrejMirtes/status/1278262546819100684

image

However, still having issues with the autoloader, it can't find `Google_Task_Retryable` in the Analytics module. Would appreciate some help 🙌 

<details>
<summary>PHPStan errors:</summary>

```
 ------ --------------------------------------------------------------------- 
  Line   Backend/Modules/Analytics/Form/SettingsStepAccountTypeInterface.php  
 ------ --------------------------------------------------------------------- 
         Reflection error: Google_Task_Retryable not found.                   
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols   
  72     Reflection error: Google_Task_Retryable not found.                   
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols   
 ------ --------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------- 
  Line   Backend/Modules/Analytics/Widgets/RecentVisits.php                  
 ------ -------------------------------------------------------------------- 
         Reflection error: Google_Task_Retryable not found.                  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  32     Reflection error: Google_Task_Retryable not found.                  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
 ------ -------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------- 
  Line   Backend/Modules/Analytics/Widgets/TraficSources.php                 
 ------ -------------------------------------------------------------------- 
         Reflection error: Google_Task_Retryable not found.                  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  32     Reflection error: Google_Task_Retryable not found.                  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
 ------ -------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------- 
  Line   Backend/Modules/ContentBlocks/Actions/Delete.php                    
 ------ -------------------------------------------------------------------- 
  55     Method                                                              
         Backend\Modules\ContentBlocks\Actions\Delete::getContentBlock()     
         should return                                                       
         Backend\Modules\ContentBlocks\Domain\ContentBlock\ContentBlock but  
         return statement is missing.                                        
 ------ -------------------------------------------------------------------- 

[...]

PR #3227, opened 2020-11-13
Upgrade phpstan

carakas:
## Type

- Non critical bugfix
- Enhancement

Fixes #3210

## Pull request description

Upgrade phpstan to the latest version and make it pass on level 1

### DIFF
--- a/src/Backend/Modules/FormBuilder/Ajax/SaveField.php
+++ b/src/Backend/Modules/FormBuilder/Ajax/SaveField.php
@@ -264,7 +264,7 @@ class SaveField extends BackendBaseAJAXAction
         if ($label !== '') {
             $settings['label'] = \SpoonFilter::htmlspecialchars($label);
         }
-        if (isset($values)) {
+        if (count($values) > 0) {
             $settings['values'] = $values;
         }
         if ($defaultValues !== '') {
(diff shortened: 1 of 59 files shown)


src/Backend/Modules/FormBuilder/Ajax/SaveField.php
```diff
@@ -264,7 +264,7 @@ public function execute(): void
         if ($label !== '') {
             $settings['label'] = \SpoonFilter::htmlspecialchars($label);
         }
-        if (isset($values)) {
+        if (count($values) > 0) {
```
jessedobbelaere, line 267: I see an if check for  `$values !== ''` somewhere above. Doing a count on a non-array/non-countable will fail from PHP >7.2 with `count(): Parameter must be an array or an object that implements Countable`

If we are not sure its an array or countable, we could use `if (!empty($values)) {` ?

src/Backend/Modules/MediaLibrary/Domain/MediaFolder/MediaFolder.php
```diff
@@ -82,7 +82,7 @@ class MediaFolder implements JsonSerializable
     protected $items;
 
     /**
-     * @var ArrayCollection
+     * @var Collection<MediaFolder>
```
jessedobbelaere, line 85: 👌

jessedobbelaere (approved):
Awesome job!! 🥇

src/Backend/Modules/FormBuilder/Ajax/SaveField.php
carakas: Good catch, I'll check if I can make it fail :p

PR #3229, opened 2020-11-14
Fix composer issues during docker build

jessedobbelaere:
## Type

- Non critical bugfix

## Pull request description

2 issues:

Composer.lock is outdated
```
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
```

`--no-suggest` is deprecated in composer v2 and removed in composer v3

```
Step 15/19 : RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest &&     composer clear-cache
 ---> Running in 09ecbc21bef9
You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.
Installing dependencies from lock file
Verifying lock file contents can be installed on current platform.
```


carakas (approved):

carakas (changes requested):
Hmm, I wonder what happend there with all those test packages dissapearing from require-dev

jessedobbelaere:
Maybe because of a newer phpstan versions @carakas ? I only ran `composer update phpstan/phpstan phpstan/phpstan-symfony --no-scripts` on current master to fix the outdated lock

carakas (approved):

PR #3238, opened 2020-11-25
Use radio buttons for selecting templates

bjorvack:
## Type

- Enhancement

## Pull request description

Renders the template choices using radiobuttons

### DIFF
--- a/src/Backend/Modules/Pages/Resources/views/FormLayout.html.twig
+++ b/src/Backend/Modules/Pages/Resources/views/FormLayout.html.twig
@@ -293,34 +293,16 @@
         <div class="modal-body">
           {{ macro.alert('warning', 'msg.TemplateChangeWarning'|trans|ucfirst) }}
           <div class="row">
-
-            {# @todo: loop through templates  #}
-            {# loop 1 example #}
-            <div class="col-lg-6">
-              <div class="form-check mb-3">
-                <input class="form-check-input" type="radio" name="templateRadios" id="templateRadio1" value="templateName1" checked>
-                <label class="form-check-label" for="templateRadio1">
-                  Template name <br>
+            {% for template in form.templateId.children %}
+              <div class="col-lg-6">
+                <div class="form-check mb-3">
+                  {{ form_widget(template) }}
                   <div class="template-visual-small">
-                    {{ selectedTemplate.html|raw }}
+                    {{ template.vars.attr['data-html']|raw }}
                   </div>
-                </label>
+                </div>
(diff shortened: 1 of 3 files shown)


src/Backend/Modules/Pages/Resources/views/FormLayout.html.twig
```diff
@@ -293,34 +293,16 @@
...
-              </div>
-            </div>
-
-            {# loop 2 example #}
-            <div class="col-lg-6">
-              <div class="form-check mb-3">
-                <input class="form-check-input" type="radio" name="templateRadios" id="templateRadio2" value="templateName2" checked>
-                <label class="form-check-label" for="templateRadio2">
-                  Template name <br>
+            {% for template in form.templateId.children %}
```
carakas, line 296: Why not show the visualisation of the template anymore?

carakas (changes requested):

src/Backend/Modules/Pages/Resources/views/FormLayout.html.twig
bjorvack: We do, this is done on line 301

Katrienvh (approved):

src/Backend/Modules/Pages/Resources/views/FormLayout.html.twig
```diff
@@ -293,34 +293,16 @@
...
+              <div class="col-lg-6">
+                <div class="form-check mb-3">
+                  {{ form_widget(template) }}
                   <div class="template-visual-small">
-                    {{ selectedTemplate.html|raw }}
+                    {{ template.vars.attr['data-html']|raw }}
```
carakas, line 301: if you add in in the attr field, does it then not also appear on as an attribute on the checkbox?

carakas (changes requested):

carakas:
I think it is better to create a custom form type and pass it like this https://github.com/forkcms/forkcms/blob/master/src/Backend/Form/Type/MetaType.php#L307 and then use the id of the template as key or something

src/Backend/Modules/Pages/Resources/views/FormLayout.html.twig
bjorvack: @carakas changed in bd842e9ac91cadb60b24b0718a163ffe61bfbbed

bjorvack:
@carakas It wasn't lazy loaded. I've updated the code in 96de0a2

PR #3254, opened 2020-12-09
Test all php versions

carakas:
## Type

- Non critical bugfix

## Pull request description

The tests didn't run on all PHP  versions, this should fix that


jessedobbelaere (approved):
This is awesome. We went from 10-15min (travis?) to 9min (GitHub actions), to 2minutes (pcov instead of xdebug) to run the functional tests 🕺 👏

PR #3264, opened 2020-12-23
updated to fix uses of MySQL reserved words

### LINKED ISSUE #3260: user(s) and group(s) are now reserved words in mysql 8.0.1 and beyond.
## Bug Report

User edits in forkCMS error when using mysql 8.0.1 or above, due to reserved words [Source from MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/keywords.html)

| Q            | A     |
| ------------ | ----- |
| Critical Bug | no    |
| Version      | 5.9.2 |

### Problem description

User edits in forkCMS error when using mysql 8.0.1 or above, due to reserved words - when trying to edit a user, receive a server 500 error. Logs indicate a spoon DB SQL syntax error. Trying to ```select * from groups;``` in the site table fails, requiring backticks. 

Error log during server 500 error:

```
request.CRITICAL: Uncaught PHP Exception PDOException: "SQLSTATE[4200]: Syntax error or access violation: 1064 You have an error in your SQL sytnax; check the manual that corresponds to your MYSQL server version for the right syntax to use near: 'groups as I       INNER JOIN users_groups AS ug ON i.id = ug.group_id
```

### Steps to reproduce

Attempt to edit a user profile as a forkCMS admin while running MySQL 8.0.1 or greater

### Expected behavior / Proposed solution

Expected behaviour is for user edits to be successful, not throwing a server 500 error. 

**Proposed fix:** 

What seems to be working for me currently is editing ```forkcms/src/Backend/Modules/Groups/Engine/Model.php``` to wrap any standalone instance of "user", "users", "group", "groups" within SQL queries within backticks. While several replacements were needed, below is a screenshot of how one of the queries look after my quick modification:

edited SQL query

 Perhaps a more elegant solution in the future would be to use table prefixes in future installs, to avoid reserved word conflicts, such as ```forkcms_users``` and ```forkcms_groups```

initinfosec:
Per https://github.com/forkcms/forkcms/issues/3260#issuecomment-747647574 - updated with temporary fix to address issues with mysql reserved words: user(s) and group(s). For now, uses of those table names are just encapsulated in literals e.g. ```SELECT * from `group`;``` although it may be better to change the installer code to create DB table prefixes. Creator notes that v6 branch has been refactored and this is a non-issue.

## Type

Non critical bugfix

Fixes #3260 in v5

## Pull request description
Per https://github.com/forkcms/forkcms/issues/3260#issuecomment-747647574 - updated with temporary fix to address issues with mysql reserved words: user(s) and group(s). For now, uses of those table names are just encapsulated in literals e.g. ```SELECT * from `group`;``` although it may be better to change the installer code to create DB table prefixes. Creator notes that v6 branch has been refactored and this is a non-issue.

### DIFF
--- a/src/Backend/Modules/Groups/Engine/Model.php
+++ b/src/Backend/Modules/Groups/Engine/Model.php
@@ -11,14 +11,14 @@ class Model
 {
     const QUERY_BROWSE =
         'SELECT g.id, g.name, COUNT(u.id) AS num_users
-         FROM groups AS g
+         FROM `groups` AS g
          LEFT OUTER JOIN users_groups AS ug ON g.id = ug.group_id
-         LEFT OUTER JOIN users AS u ON u.id = ug.user_id
+         LEFT OUTER JOIN `users` AS u ON u.id = ug.user_id
          GROUP BY g.id';
 
     const QUERY_ACTIVE_USERS =
         'SELECT u.id, u.email
-         FROM users AS u
+         FROM `users` AS u
          INNER JOIN users_groups AS ug ON u.id = ug.user_id
          WHERE ug.group_id = ? AND u.deleted = ?';
 
@@ -44,7 +44,7 @@ class Model
     {
         return (bool) BackendModel::getContainer()->get('database')->getVar(
             'SELECT i.*
-             FROM groups AS i
+             FROM `groups` AS i
              WHERE i.name = ?',
             [$groupName]
         );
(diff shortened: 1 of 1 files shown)


jessedobbelaere (approved):
LGTM 👍 Thanks! 🙌

carakas (approved):

PR #3270, opened 2021-01-08
Clarify that html in translations is not considered a security issue

carakas:
## Type

- Enhancement

## Pull request description

We sometimes get reports of people reporting that they found a security issue with our translation system. By adding this to the readme we are clarifying what we consider a feature instead of a security issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ To install the dependencies, you can run the command below in the document-root:
 ## Security
 
 If you discover any security-related issues, please email core@fork-cms.com instead of using the issue tracker.
+HTML is allowed in translations because you sometimes need it. Any reports regarding this will not be accepted as a security issue. Owners of a website can narrow down who can add/edit translation strings using the group permissions.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -27,6 +27,7 @@ To install the dependencies, you can run the command below in the document-root:
 ## Security
 
 If you discover any security-related issues, please email core@fork-cms.com instead of using the issue tracker.
+HTML is allowed in translations because you sometimes need it. Any reports regarding this will not be accepted as a security issue. Owners of a website can narrow down who can add/edit translation strings using the group permissions.
```
jessedobbelaere, line 30: Add this to https://github.com/forkcms/forkcms/blob/master/.github/SECURITY.md as well 👀 ? This is what you see if you go to Issues > New > Security issue

carakas: Done

jessedobbelaere (approved):

Issue #3295, opened 2021-02-09
Bootstrap v5


tijsverkoyen:

tijsverkoyen:
Internal reference: https://app.activecollab.com/108877/projects/62?modal=Task-116422-62

PR #3311, opened 2021-02-15
Upgrade to bootstrap5

Katrienvh:
## Type

- Enhancement

## Pull request description

Upgrade frontend/backend to Bootstrap 5 beta.

Most importend changes I did: 
- Input group does not need an extra wrapper link pre/append.
- Removed `.form-row` because its deprecated and used new gutter utility classes instead.
- `.form-group` is deprecated, but made our own custom class so we can still use is.
- All data attributes that trigger the Bootstrap javascipt are now prefixed with `data-bs-`
- Spacing utility classes are changed. `.ml` (Margin Left) and `.mr` (Margin Right) are now `.ms` (margin start = left) en `me` (margin end = right). This counts for al the classes that use the same structure.
- Removed `.form-inline` because it's deprecated and use flexbox utility classes instead.
- All custom form element stylings are now the default stylings, so removed the deprecated custom styling classes.
- Bootstrap now uses `visually-hidden` instead of `sr-only`

This should resolve: https://github.com/forkcms/forkcms/issues/3295


src/Common/Core/Form.php
```diff
@@ -243,7 +243,7 @@ public function addDropdown(
     ): SpoonFormDropdown {
         $name = (string) $name;
         $multipleSelection = (bool) $multipleSelection;
-        $class = (string) ($class ?? 'form-control fork-form-select');
```
carakas, line 246: Was there styling on this class that needs to be removed?

carakas (approved):
LGTM

src/Common/Core/Form.php
Katrienvh: Nope, `form-control` was from bootstrap and `fork-form-select` is an old class that was used long ago in fork

Katrienvh:
To be fixed: installer -> some layout is a little broken

carakas:
@Katrienvh there is a test looking for <label for="file" class="custom-file-label"> but failing because of the new layout. Can you update the test so that it looks for the correct HTML of the label?

carakas (changes requested):
Failing test needs to be fixed first by updating the html

carakas (approved):

PR #3323, opened 2021-02-27
Add CLI command to install a module

jessedobbelaere:
## Type

- Enhancement

When I setup CI/CD for a module, I kind of lack 
* Installing Fork CMS using CLI, so i have to load a `dump.sql` with the base schema of Fork CMS instead and generate the parameters.yml myself. 
* Running the installers of the modules I want to test in my CI/CD. It's kind of cumbersome to dump the module's tables to a `dump.sql` file and keep the sql dump up to date... because we work with Doctrine entities now anyway. Ideally, I can setup a database, import `dump.sql` (basic fork cms schema) and then install the module and the theme that I need (locale, tables, images, ...)

## Pull request description

This PR adds a command that allows the developer to install a module using his CLI. E.g.

```bash
# Show an interactive question with autocomplete
bin/console forkcms:install:module

# Install a module without interaction
bin/console forkcms:install:module Faq
```

Screenshot 2021-02-27 at 22 34 58

### DIFF
--- a/src/ForkCMS/Bundle/InstallerBundle/Console/InstallModuleCommand.php
+++ b/src/ForkCMS/Bundle/InstallerBundle/Console/InstallModuleCommand.php
@@ -109,7 +109,6 @@ class InstallModuleCommand extends Command
         });
 
         return $helper->ask($input, $output, $question);
-
     }
 
     private function getAlreadyInstalledModules(): array
--- a/src/ForkCMS/Bundle/InstallerBundle/Console/InstallModuleCommand.php
+++ b/src/ForkCMS/Bundle/InstallerBundle/Console/InstallModuleCommand.php
@@ -1,0 +1,177 @@
+<?php
+
+namespace ForkCMS\Bundle\InstallerBundle\Console;
+
+use Backend\Core\Engine\Model as BackendModel;
+use Backend\Modules\Extensions\Engine\Model;
+use Backend\Modules\Extensions\Engine\Model as BackendExtensionsModel;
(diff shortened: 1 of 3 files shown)


src/ForkCMS/Bundle/InstallerBundle/Console/InstallModuleCommand.php
```diff
@@ -0,0 +1,189 @@
...
+        $module = $this->getModuleToInstall($input, $output);
+
+        if (BackendExtensionsModel::existsModule($module)) {
+            // Make sure this module can be installed
+            $output->writeln("<comment>Validating if module can be installed...</comment>");
+            $this->validateIfModuleCanBeInstalled($module);
+
+            // Reboot the kernel to trigger a kernel initialize which registers the Dependency Injection Extension of the
+            // module we would like to install. Also, make sure to replace the static cached container in our BaseModel!
+            $_SERVER['INSTALLING_MODULE'] = $module;
```
jessedobbelaere, line 65: It took me a while to understand that in the `Kernel.php` we explicitely listen to the exact route which installs a module, and then we register the DI Extension file of the module. 

I need a way to re-initialize the kernel and clearing the cache did not seem to work, so I reboot the Kernel. 

Also needed a way to pass the module name back to the `Kernel.php` file to pick up the module name and initialize the DI Extension. Feel free to suggest better options than putting it in `$_SERVER` ? The whole register-module-extension-on-install feels a bit hacky unfortunately

carakas (approved):

Issue #3104, opened 2020-05-15
Media gallery media gone after saving


Katrienvh:
## Bug Report

| Q            | A     |
| ------------ | ----- |
| Critical Bug | yes    |
| Version      | 5.8.2 |

### Problem description

Gallery does not save the media

### Steps to reproduce

Add a gallery, connect some media and add a title, save the gallery. Then edit it again and the media is not there anymore

carakas:
@Katrienvh is this fork 5.8.2 or for restructure core?

Katrienvh:
restructure core

Katrienvh:
> @Katrienvh is this fork 5.8.2 or for restructure core?

@carakas restructure core

carakas:
Can you fix the description that one is stating form 5.8.2

tijsverkoyen:
Internal reference: https://app.activecollab.com/108877/projects/2553?modal=Task-95842-2553

PR #3350, opened 2021-03-18
Fix mediagallery js

bjorvack:
## Type
- Critical bugfix

#3104

## Pull request description
Fixes bug with media items aren't saved

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Js/Helper/Upload.js
+++ b/src/Backend/Modules/MediaLibrary/Js/Helper/Upload.js
@@ -165,14 +165,17 @@ export class Upload {
    * Add uploaded media to group
    */
   addUploadedMediaToGroup () {
+    let config = this.config
     // loop remaining items in uploaded media and push them to current group
     $('#uploadedMedia').find('li').each(function () {
       // get id
       const id = $(this).attr('id').replace('media-', '')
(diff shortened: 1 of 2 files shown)


src/Backend/Modules/MediaLibrary/Js/Helper/Upload.js
```diff
@@ -165,15 +165,18 @@ export class Upload {
...
   addUploadedMediaToGroup () {
+    let config = this.config
     // loop remaining items in uploaded media and push them to current group
     $('#uploadedMedia').find('li').each(function () {
       // get id
       const id = $(this).attr('id').replace('media-', '')
 
       // add each id to array
-      this.config.currentMediaItemIds.push(id)
+      config.currentMediaItemIds.push(id)
```
Katrienvh, line 175: There is a nicer way to do this. On line 170 you need to use an arrow function and by doing that 'this' wil be available  on line 175, by that you do not need to use the variable `config` set on line 168

bjorvack: fixed b227c51b2643f5005f3ce6ac6d035ee2d0c87dbf

Katrienvh (approved):

PR #3351, opened 2021-03-18
Security fixes

carakas:
## Type

- Security

## Pull request description

Fixes a couple of security issues with uploads (images, themes), and block a whole lot of extra extensions that we don't want in case they get through

Also makes sure the session id changes when logging in or out

Add checks for csrf tokens in backend widgets


.htaccess
```diff
@@ -49,7 +49,27 @@ deny from env=stayout
 
     # forbidden folders
     RewriteRule .*\.php - [F]
-    RewriteRule .*\.twig - [F]
+    RewriteRule \.php5 - [F]
```
jessedobbelaere, line 52: Q: Any reason we use `.*\.php` and the other rules don't have the wildcard?

src/Backend/Modules/Extensions/Actions/UploadTheme.php
```diff
@@ -242,6 +242,7 @@ private function getValidatedFilesList(ZipArchive $zip): array
             // We skip all the files that are outside of the theme folder or on the ignore list.
             if ($this->checkIfPathContainsIgnoredWord($fileName) ||
                 (!empty($this->parentFolderName) && mb_stripos($fileName, $this->parentFolderName) !== 0)
+                || mb_strpos(basename($fileName), '.') === 0 // ignore hidden files
```
jessedobbelaere, line 245:
```suggestion
            if ($this->checkIfPathContainsIgnoredWord($fileName)
                || (!empty($this->parentFolderName) && mb_stripos($fileName, $this->parentFolderName) !== 0)
                || mb_strpos(basename($fileName), '.') === 0 // ignore hidden files
```

For consistency?

src/Backend/Modules/Pages/Ajax/UploadFile.php
```diff
@@ -23,11 +23,17 @@ public function execute(): void
...
-        $fileName = $this->writeFile(
-            $this->getFileContentFromRequest($request),
-            $this->getFileNameFromRequest($request),
-            $request->get('type')
-        );
+        try {
+            $fileName = $this->writeFile(
+                $this->getFileContentFromRequest($request),
+                $this->getFileNameFromRequest($request),
+                $request->get('type')
```
jessedobbelaere, line 30: These check warnings by codecov are quite annoying while code reviewing files without unit tests. I wonder if we can disable that somehow

src/Backend/Modules/Pages/Ajax/UploadFile.php
```diff
@@ -100,6 +106,9 @@ private function writeFile(string $content, string $fileName, string $destinatio
         // convert the filename to url friendly version
         $baseName = Uri::getUrl(pathinfo($fileName, PATHINFO_FILENAME));
         $extension = pathinfo($fileName, PATHINFO_EXTENSION);
+        if (!in_array(strtolower($extension), ['jpg', 'jpeg', 'gif', 'png'])) {
```
jessedobbelaere, line 109: That list is a bit restrictive.... Earlier this week I was using a user template block in Fork CMS and uploaded `.webp` and `.svg` files for example. `.avif` is a good one to have too in a year or so.

.htaccess
carakas: I think it had to do with the index.php but I'll have a look if we can tighten it more

src/Backend/Modules/Pages/Ajax/UploadFile.php
carakas: I'll add those

src/Backend/Modules/Pages/Ajax/UploadFile.php
carakas: the patch ones are disabled so I don't know why they are showing

carakas: 4bf38b485a718091613a7889ce78d3b650e19de9

src/Backend/Modules/Pages/Ajax/UploadFile.php
carakas: 0117c1e4e6c180ed134653066d1196745b31a43a

jessedobbelaere: Awesome!

jessedobbelaere (approved):

PR #3376, opened 2021-04-22
Offline page

bjorvack:
## Type

- Feature

## Pull request description

Adds service worker for offline page
Internal reference https://app.activecollab.com/108877/reports/team-timeline?modal=Task-84793-62


Katrienvh (changes requested):

carakas:
@bjorvack can you include a screenshot or something so we can see what it looks like?

app/Kernel.php
```diff
@@ -100,6 +100,8 @@ public function defineForkConstants(): void
         defined('ACTION_GROUP_TAG') || define('ACTION_GROUP_TAG', $container->getParameter('action.group_tag'));
         defined('ACTION_RIGHTS_LEVEL') || define('ACTION_RIGHTS_LEVEL', $container->getParameter('action.rights_level'));
 
+        defined('OFFLINE_PATH') || define('OFFLINE_PATH', PATH_WWW . '/offline');
```
carakas, line 103: We are working on getting those constants out, not a fan of adding new ones

offline/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
carakas, line 1: why is this here? You have a file in that directory called offline.html but you are also telling git to ignore it?

offline/offline.html
```diff
@@ -0,0 +1,33 @@
+<!doctype html>
```
carakas, line 1: why a html page like this? Would it not be possible to include it in the cms? so the text won't need to be hardcoded?

offline/offline.html
```diff
@@ -0,0 +1,33 @@
...
+
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+
+  <title>Offline page</title>
+  <meta name="description" content="Offline page of this awesome website">
+  <meta name="author" content="Sarah Meysman">
```
carakas, line 9: I would remove this line

offline/offline.html
```diff
@@ -0,0 +1,33 @@
...
+  <link href="/src/Frontend/Themes/Fork/Core/Layout/Css/screen.css?m=1593680880" rel="stylesheet" />
+
+</head>
+
+<body>
+<div class="container">
+  <div class="row justify-content-center mt-5">
+    <div class="col-md-8 col-xl-6">
+
+      <h1>You're offline</h1>
```
carakas, line 21: wouldnt it be possible to use the translation system so these text are in the language you are visiting the site?

package-lock.json
```diff
@@ -1,8 +1,16451 @@
 {
   "name": "forkcms",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
```
carakas, line 6: why is there suddenly so much npm stuff here? seems to contain a bunch of things that arent related

src/Backend/Modules/Pages/Actions/Settings.php
```diff
@@ -22,32 +33,95 @@ public function execute(): void
...
+    private function renderOfflinePage(): void
+    {
+        /** @var Environment $twig */
+        $twig = $this->get('twig');
+
+        /** @var ParseBlocksExtension $extension */
+        $extension = $twig->getExtension(ParseBlocksExtension::class);
+
+        $html = $twig->render(
+            BACKEND_MODULES_PATH . '/' . $this->getModule() .'/Layout/Templates/Offline.html.twig',
```
carakas, line 101: wouldn't it make more sense to put this in the frontend? That way you can also override it in your theme

src/Backend/Modules/Pages/Actions/Settings.php
```diff
@@ -22,32 +33,95 @@ public function execute(): void
...
+                'image' => $this->get('media_library.repository.group')->findOneById(
+                    $this->get('fork.settings')->get(
+                        $this->getModule(),
+                        'offline_image_' . Language::getWorkingLanguage()
+                    )
+                )
+            ]
+        );
+
+        $offlinePage = fopen(OFFLINE_PATH . '/offline_' . Language::getWorkingLanguage() . '.html', 'w');
```
carakas, line 123: Please use Filesystem  dump of symfony for this.
Looks like the page is generated but you committed the generated code. Can you have a look at removing it?
Is this also enabled by default? If so if it generated when installing fork?

src/Backend/Modules/Pages/Domain/Page/SettingsDataTransferObject.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Backend\Core\Language\Locale;
+use Backend\Modules\MediaLibrary\Domain\MediaGroup\MediaGroup;
+use Backend\Modules\MediaLibrary\Domain\MediaGroup\MediaGroupRepository;
+use Backend\Modules\MediaLibrary\Domain\MediaGroup\Type as MediaGroupType;
+use Common\ModulesSettings;
+
+class SettingsDataTransferObject
+{
+    const MODULE = 'Pages';
```
carakas, line 13: not a big fan of doing this

src/Backend/Modules/Pages/Domain/Page/SettingsDataTransferObject.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public $offlineText;
+
+    /** @var MediaGroup */
+    public $offlineImage;
+
+    public function __construct(ModulesSettings $settings, MediaGroupRepository $mediaGroupRepository)
+    {
+        $this->metaNavigation = $settings->get(
+            $this::MODULE,
```
carakas, line 27: why $this and not self if you use this?

src/Backend/Modules/Pages/Domain/Page/SettingsDataTransferObject.php
carakas: Please also add public protected or private if you do this

src/Backend/Modules/Pages/Layout/Templates/Offline.html.twig
```diff
@@ -0,0 +1,32 @@
+<!doctype html>
+
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+
+  <title>Offline page</title>
```
carakas, line 7: Use {{title}} here as well

src/Backend/Modules/Pages/Layout/Templates/Offline.html.twig
```diff
@@ -0,0 +1,32 @@
...
+  <meta name="description" content="Offline page of this awesome website">
+  <meta name="author" content="Sarah Meysman">
```
carakas, line 9: Remove those two lines please

src/Backend/Modules/Pages/Layout/Templates/Offline.html.twig
```diff
@@ -0,0 +1,32 @@
...
+
+  <!--<link href="css/bootstrap.min.css" rel="stylesheet" />-->
+  <link href="/src/Frontend/Themes/Fork/Core/Layout/Css/screen.css?m=1593680880" rel="stylesheet" />
```
carakas, line 12: You should load the css in a proper way, you hardcoded the theme and the timestamp

src/Frontend/Core/Header/Header.php
```diff
@@ -125,6 +126,23 @@ public function __construct(KernelInterface $kernel)
         $this->jsData = new JsData($jsData);
 
         $this->meta = new MetaCollection();
+
+        $this->addJS('/src/Frontend/Modules/Pages/Js/Pages.js', true, true, Priority::module());
```
carakas, line 130: Not a fan of this, include it in the ES6 instead of this way.

carakas (changes requested):

carakas:
@bjorvack could you have a look at the feedback please?

carakas:
Closing stale PR, let me know if you want to finish the feedback

PR #3377, opened 2021-04-24
Throw event when changing session id

wouterverstuyf:
## Type
- Enhancement

When a public profile login/logout the session id changes.
Could be handy for other modules to hook in through an event, to get the new Session ID.

## Pull request description
You can create a listener in your own module that detects the login/logout action. In the listener you can get the new session ID and do your stuff

/Frontend/Modules/MyModule/EventListener/ProfilesSessionIdChangedListener.php

<pre>
namespace Frontend\Modules\MyModule\EventListener;

use Common\Events\ForkSessionIdChangedEvent;

class ForkSessionIdChangedListener
{
    public function onForkSessionIdChanged(ForkSessionIdChangedEvent $event): void
    {
        // do your stuff
        // $event->getSessionId()
        // $event->getOldSessionId()
    }
}
</pre>


wouterverstuyf:
Question : when you login/logout in the backend section, the Session ID also changes. Should this be event be triggered from the backend login/logout function also? Is it ok to trigger a frontend event from within the backend section?

carakas:
@wouterverstuyf I would add it in the Common namespace, not in the profiles one. It is tied to the session, not a profile

src/Frontend/Modules/Profiles/Engine/Authentication.php
```diff
@@ -243,13 +247,21 @@ public static function login(int $profileId, bool $remember = false): void
...
+        );
+
         // load the profile object
         self::$profile = new FrontendProfilesProfile($profileId);
     }
 
     public static function logout(): void
     {
         $session = FrontendModel::getSession();
+        $oldSession = FrontendModel::getSession()->getId();
```
carakas, line 263: you can use $session->getId() here

wouterverstuyf:
> @wouterverstuyf I would add it in the Common namespace, not in the profiles one. It is tied to the session, not a profile, and then you can throw the event in the backend as well

Yes, got the same feeling this morning, I check it ;)

src/Frontend/Modules/Profiles/Event/ProfilesSessionIdChangedEvent.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * @var string
+     */
+    protected $sessionId;
+
+    public function __construct(string $oldSessionId)
+    {
+        $this->oldSessionId = $oldSessionId;
+        $this->sessionId = FrontendModel::getSession()->getId();
```
carakas, line 23: Pass this one in the constructor as well.

carakas (changes requested):

carakas (approved):

PR #3381, opened 2021-04-27
Vue social buttons

Katrienvh:
## Type

- Feature

## Pull request description

vueJs component for social media share buttons.

Platforms:
- Facebook
- Twitter
- LinkedIn
-> easily expandable

Component props:

- facebook: optional (booleaan)
- twitter: optional (booleaan)
- linkedin: optional (booleaan)
- page_url: optional (when != then current url)

image


src/Frontend/Core/Js/Vue-components/VShareButton.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div>
```
tijsverkoyen: Would it be an idea to add a `data-role="share-button"` so we can access it later on through JS? Or is this overengineering?
Or add a class on the wrapper so it can be styled more easily?

src/Frontend/Themes/Fork/Core/Layout/Sass/_bootstrap-variables.scss
```diff
@@ -51,6 +51,18 @@ $green:         #2C9F42;
...
 $cyan:          #5bc0de;
 
+// brands
+$brands: ();
+// stylelint-disable-next-line scss/dollar-variable-default
+$brands: map-merge(
+    (
+      "facebook": #3b5998,
+      "twitter":  #00acee,
+      "linked-in": #0072b1
```
tijsverkoyen, line 61: Shouldn't this be `#0077b5`

Katrienvh: That color I found on google, when I look back there are 9999 different hex values on google for LinkedIn. I updated the color to the one I found on there website that they are using to.

src/Frontend/Core/Js/Vue-components/VShareButton.vue
Katrienvh: I think that is overkill. Besides that I think there are other ways to access the component. The styling is on de buttons its self. I should not add an unnecessary class if not needed. Just to keep it simple.

tijsverkoyen: Agreed

tijsverkoyen (approved):

carakas (approved):

Issue #3100, opened 2020-05-15
Profiles data grid add to group


Katrienvh:
## Bug Report

| Q            | A     |
| ------------ | ----- |
| Critical Bug | no    |
| Version      | 5.8.2 |

### Steps to reproduce

When you are on Profiles module overview in backend. Select a user and execute 'add to group' on that user (when there aren't created any groups yet). You now get the modal to confirm. In that modal the select element isn't there, because there are no groups yet but you can confirm. After the page refresh, you got the following error: 'You must select a group to perfom this action.' 

### Expected behavior / Proposed solution

I think it would be less of a hassle if the option is just hidden at the bottom of the datagrid when there are no groups. My vision if it can't/doesn't work, do not let the user even try or give the possibility.

tijsverkoyen:
Internal url: https://app.activecollab.com/108877/projects/62?modal=Task-116457-62

PR #3391, opened 2021-05-04
Fix profiles datagrid

dreadnip:
## Type

- Non critical bugfix

fixes https://github.com/forkcms/forkcms/issues/3100

## Pull request description

Fixes a couple of inconsistencies when handling profiles/groups from the mass action select.

### DIFF
--- a/src/Backend/Modules/Profiles/Actions/Index.php
+++ b/src/Backend/Modules/Profiles/Actions/Index.php
@@ -130,14 +130,26 @@ class Index extends BackendBaseActionIndex
         );
         $this->dgProfiles->setColumnFunction('htmlspecialchars', ['[display_name]'], 'display_name', false);
 
-        // add the mass action controls
-        $this->dgProfiles->setMassActionCheckboxes('check', '[id]');
-        $ddmMassAction = new \SpoonFormDropdown(
-            'action',
-            [
+        // check if
+        $groups = BackendProfilesModel::getGroups();
+
+        if (count($groups) > 0) {
+            $actions = [
                 'addToGroup' => BL::getLabel('AddToGroup'),
                 'delete' => BL::getLabel('Delete'),
-            ],
+            ];
+        } else {
+            $actions = [
+                'delete' => BL::getLabel('Delete'),
+            ];
+        }
+
+        // add the mass action controls
+        $this->dgProfiles->setMassActionCheckboxes('check', '[id]');
+
+        $ddmMassAction = new \SpoonFormDropdown(
+            'action',
+            $actions,
             'addToGroup',
             false,
             'form-control form-control-sm',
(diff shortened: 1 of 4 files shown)


src/Backend/Modules/Profiles/Actions/Index.php
```diff
@@ -130,14 +130,26 @@ private function loadDataGrid(): void
         );
         $this->dgProfiles->setColumnFunction('htmlspecialchars', ['[display_name]'], 'display_name', false);
 
+        // check if
+        $groups = BackendProfilesModel::getGroups();
```
tijsverkoyen, line 134: Would't I be more logical to have the $actions populated at the start. So:

```
$actions = [
  'delete' => BL::getLabel('Delete'),
];

if (count($groups) > 0) {
  $actions['addToGroup'] = BL::getLabel('AddToGroup');
}
```

dreadnip: Updated in 22b604e

tijsverkoyen (approved):

src/Backend/Modules/Profiles/Engine/Model.php
```diff
@@ -691,9 +691,15 @@ public static function insertProfileGroup(array $membership): int
         $profile = BackendModel::get('profile.repository.profile')->find($membership['profile_id']);
         $group = BackendModel::get('profile.repository.profile_group')->find($membership['group_id']);
 
+        $startsOn = new DateTime();
+        $startsOn->setTimestamp($membership['starts_on']);
+        $startsOn->format('Y-m-d H:i:s');
```
carakas, line 696: you are not using the output here.

src/Backend/Modules/Profiles/Engine/Model.php
```diff
@@ -691,9 +691,15 @@ public static function insertProfileGroup(array $membership): int
...
+
         $expiresOn = null;
         if (array_key_exists('expires_on', $membership)) {
-            $expiresOn = DateTime::createFromFormat('Y-m-d H:i:s', $membership['expires_on']);
+            $expiresOn = new DateTime();
+            $expiresOn->setTimestamp($membership['expires_on']);
+            $expiresOn->format('Y-m-d H:i:s');
```
carakas, line 702: you are not using the output here.

carakas (changes requested):

src/Backend/Modules/Profiles/Engine/Model.php
dreadnip: Fixed in 3a36fa8

src/Backend/Modules/Profiles/Engine/Model.php
dreadnip: Fixed in 3a36fa8

dreadnip:
3a36fa8 should fix the datagrid mass actions add to group behaviour. While testing I noticed that the profile/group add/update flow is broken. I didn't dig deep but it seemed like the form fields still used the old notation `expiration_date`, while the back-end code expected `expiry_on`. Should be looked into.

carakas (approved):

carakas (changes requested):
@dreadnip there is a failing test for 7.4, can you fix this one please?

dreadnip:
Removed a couple of old strtotime() instances in the test. Should be okay now.

carakas (approved):

PR #3419, opened 2021-06-11
Replace translation type action with slug

carakas:
## Type

- Enhancement

## Pull request description

The type of action is actually a slug. Using the category name slug makes more sense imho


tijsverkoyen:
I don't have anything against it, but should we bother, as I would think we will move to Symfony translations at some point?

tijsverkoyen (approved):

carakas:
@tijsverkoyen I'm busy changing the implementation of the translations that it works by loading in our XML since there are some things in symfony 5 that make this easier and then in the database we'll only have the once that have changed, it was while making that that I noticed this thing.

PR #3424, opened 2021-06-28
make it possible to upload svg files to medialibrary

davysumo:
## Type

- Enhancement

## Pull request description

Enables upload of svg files in MediaLibrary

### DIFF
--- a/src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
+++ b/src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
@@ -225,32 +225,34 @@ class MediaItem implements JsonSerializable
 
     private function setResolutionForSvg(string $path)
     {
-        if ($this->getType()->isImage()) {
-            try {
-                $xmlget = simplexml_load_file($path);
-                $viewBoxString = $xmlget->attributes()->viewBox;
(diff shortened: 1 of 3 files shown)


src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
```diff
@@ -218,6 +223,36 @@ private function setResolutionFromPath(string $path)
         }
     }
 
+    private function setResolutionForSvg(string $path)
+    {
+        if ($this->getType()->isImage()) {
```
carakas, line 228: Can you rewrite it to 
`if (!$this->getType()->isImage()) {return;}`
That will create less indentation and makes it easier to see what happens, or maybe even throw a LogicException since you shouldn't be calling that method for non images

carakas (changes requested):

src/Backend/Modules/MediaLibrary/Domain/MediaItem/MediaItem.php
davysumo: See https://github.com/forkcms/forkcms/pull/3424/commits/92d1accc3a53edab9602b64cde92d4e13513a505

carakas (approved):

PR #3431, opened 2021-07-14
WIP: Restructure core with master

jonasdekeukelaere:
PR with master changes for restructure-core branch


jonasdekeukelaere:
@Katrienvh has add 1 more change from master, with changes to src/Backend/Modules/MediaLibrary/Js/MediaLibraryHelper.js

carakas:
We should also wait for @davysumo to fix the svg stuff in the media library

src/Common/Doctrine/Entity/Meta.php
```diff
@@ -124,6 +124,8 @@ public function __construct(
         bool $titleOverwrite,
         string $url,
         bool $urlOverwrite,
+        ?string $canonicalUrl,
```
carakas, line 127: @jonasdekeukelaere could you maybe make a small PR for this here?
This is a BC break that I missed since it doesn't have default values

jonasdekeukelaere: A small PR for what?

carakas:
@jonasdekeukelaere the 7.4 tests are failing, once they work we can merge this

jonasdekeukelaere:
I also have to add some more master MRs, will try to have a look this week

PR #3432, opened 2021-07-15
overwrite filterAction of ImagineController to skip filter for svg im…

davysumo:
…ages

## Type

- Non critical bugfix

Don't apply a filter on svg images so they can be displayed

### DIFF
--- a/src/ForkCMS/ImagineController.php
+++ b/src/ForkCMS/ImagineController.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace ForkCMS;
+
+use Imagine\Exception\RuntimeException;
+use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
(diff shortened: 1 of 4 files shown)


src/ForkCMS/ImagineController.php
```diff
@@ -0,0 +1,75 @@
...
+     * @param Request $request
+     * @param string  $path
+     * @param string  $filter
+     *
+     * @throws \RuntimeException
+     * @throws BadRequestHttpException
+     *
+     * @return RedirectResponse
+     */
+    public function filterAction(Request $request, $path, $filter)
```
carakas, line 27: Would it not be possible to check if it is an svg, then handle it here yourself but if it isn't just to kick it up to the parent?

carakas (changes requested):

carakas (approved):

PR #3438, opened 2021-07-28
Duplicate page image when copying pages from one locale to another

jonasdekeukelaere:
## Type
- Non critical bugfix

When copying pages from one locale to another (with the hidden action /en/pages/copy?from=en&to=nl), the page image will be linked to the same file. Which when you change the image in one of the user template, the other will be broken as the original image gets deleted.

## Pull request description

Duplicate the page image when copying pages from one locale to another.

### DIFF
--- a/src/Backend/Modules/Pages/Engine/Model.php
+++ b/src/Backend/Modules/Pages/Engine/Model.php
@@ -4,6 +4,8 @@ namespace Backend\Modules\Pages\Engine;
 
 use Backend\Modules\ContentBlocks\Domain\ContentBlock\Command\CopyContentBlocksToOtherLocale;
 use Backend\Modules\Location\Command\CopyLocationWidgetsToOtherLocale;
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Pages/Engine/Model.php
```diff
@@ -1657,4 +1671,24 @@ private static function updateUrlAfterMove(int $pageId, array $page, int $newPar
 
         $database->update('meta', ['url' => $newUrl], 'id = ?', [$page['meta_id']]);
     }
+
+    private static function getImage(?string $image, string $metaUrl): ?string
```
tijsverkoyen, line 1675: Niet logischer om dit `copyImage` te noemen? of `getCopiedImage`

jonasdekeukelaere: Fixed 686917856

tijsverkoyen (approved):

PR #3468, opened 2021-10-03
Only run tests once on PR

jessedobbelaere:
## Type

- Enhancement

## Pull request description

When a PR is submitted, or changes are pushed to a branch, you see that tests are running twice: one for `push` and one for `pull_request`, for a whopping total of 37 tests on a single PR. See a random screenshot from current PR's:

image

I checked how API Platform is doing their CI, https://github.com/api-platform/api-platform/blob/main/.github/workflows/ci.yml#L3-L8 and that seems to be a better way. It's running only one time (for `pull request`) and also still allows push to master (if we ever do that)

You'll see on this PR that there's only one series of tests (pull request event)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -3,7 +3,7 @@ name: run-tests
 on:
     push:
         branches:
-            - main
+            - master
     pull_request: ~
     workflow_dispatch: ~
 
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,6 +1,11 @@
 name: run-tests
 
-on: [push, pull_request]
+on:
+    push:
+        branches:
+            - main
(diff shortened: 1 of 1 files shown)


carakas:
Nice :D

.github/workflows/run-tests.yml
```diff
@@ -1,6 +1,11 @@
 name: run-tests
 
-on: [push, pull_request]
+on:
+    push:
+        branches:
+            - main
```
carakas, line 6: Should still be changed to master instead of main

carakas (changes requested):

.github/workflows/run-tests.yml
jessedobbelaere: haha oops, fixed!

carakas (approved):

PR #3484, opened 2021-12-14
Translation issue in template

bart-webleads:
The message "HelpWordpressFilter" was missing the "trans" modifier.

## Type

- Non critical bugfix

## Pull request description

Added "trans" modifier for the message "HelpWordpressFilter".


carakas (approved):

carakas:
Thank you for fixing this, is this a feature you use a lot?

bart-webleads:
No, I don't use this feature at all.
Stumbled on it by accident.

carakas:
Oke, cause thinking of removing it from the 6

Issue #3474, opened 2021-10-18
Breadcrumbs: code "if loop last" removes delimiter in pure bootstrap


Sellato:
## Bug Report

| Q            | A     |
| ------------ | ----- |
| Critical Bug | no    |
| Version      | 5.x.x |

### Problem description

When we try to use pure bootstrap 4 or 5 classes in `Breadcrumb.html.twig` we lost delimiter as shown in attach

![breacrumbs](https://user-images.githubusercontent.com/5456419/137673206-e8fa628c-2be6-47cd-b1ab-3607c6143683.png)

This code `<li {% if loop.last %} class="breadcrumb-item active" {% endif %}` removes delimiter when we use original bootstrap.css . In fork-cms `screen.css` works fine, class name is `active`

When we change `<li {% if loop.last %} class="breadcrumb-item active" {% endif %}` to `<li class="breadcrumb-item active"` as in original bootstrap examples all is OK. ( in attach )

**Original `Breadcrumb.html.twig`**

```
<ul class="breadcrumb" itemscope itemtype="https://schema.org/BreadcrumbList" role="navigation">
  {% for breadcrumb in breadcrumb %}
    <li{% if loop.last %} class="active"{% endif %} itemprop="itemListElement" itemscope
      itemtype="https://schema.org/ListItem">
      {% if breadcrumb.url and not loop.last %}
        {# set the full url or the item link will not validate #}
        <a href="{{ SITE_URL }}{{ breadcrumb.url }}" title="{{ breadcrumb.title }}" itemtype="http://schema.org/Thing"
          itemprop="item">
          <span itemprop="name">{{ breadcrumb.title }}</span>
        </a>
      {% else %}
        <span itemprop="name">{{ breadcrumb.title }}</span>
      {% endif %}
      <meta itemprop="position" content="{{ loop.index }}" />
    </li>
  {% endfor %}
</ul>
```

### From bootstrap example {% if loop.last %} removed

```
  <nav aria-label="breadcrumb">
    <ol class="breadcrumb">
      {% for breadcrumb in breadcrumb %}
        <li class="breadcrumb-item active" aria-current="page">
          {% if breadcrumb.url and not loop.last %}
              <a class="" href="{{ SITE_URL }}{{ breadcrumb.url }}" itemprop="item"><span itemprop="name">{{ breadcrumb.title }}</span></a>
              {% else %}
              <span itemprop="name">{{ breadcrumb.title }}</span>
          {% endif %}
        </li>
[...]

PR #3487, opened 2021-12-24
BUGFIX: Fix Breadcrumb Style

crydotsnake:
This PR fixes the bug described in the issue #3474

### DIFF
--- a/src/Frontend/Core/Layout/Templates/Breadcrumb.html.twig
+++ b/src/Frontend/Core/Layout/Templates/Breadcrumb.html.twig
@@ -1,7 +1,7 @@
 <span class="sr-only">{{ 'lbl.YouAreHere'|trans }}</span>
 <ul class="breadcrumb" itemscope itemtype="https://schema.org/BreadcrumbList" role="navigation">
   {% for breadcrumb in breadcrumb %}
-    <li{% if loop.last %} class="breadcrumb-item{% endif %}" itemprop="itemListElement" itemscope
+    <li class="breadcrumb-item{% if loop.last %} active{% endif %}" itemprop="itemListElement" itemscope
       itemtype="https://schema.org/ListItem">
       {% if breadcrumb.url and not loop.last %}
         {# set the full url or the item link will not validate #}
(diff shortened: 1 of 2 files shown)


src/Frontend/Core/Layout/Templates/Breadcrumb.html.twig
```diff
@@ -1,7 +1,7 @@
 <span class="sr-only">{{ 'lbl.YouAreHere'|trans }}</span>
 <ul class="breadcrumb" itemscope itemtype="https://schema.org/BreadcrumbList" role="navigation">
   {% for breadcrumb in breadcrumb %}
-    <li{% if loop.last %} class="active"{% endif %} itemprop="itemListElement" itemscope
+    <li{% if loop.last %} class="breadcrumb-item active"{% endif %} itemprop="itemListElement" itemscope
```
carakas, line 4: I think it should be
```suggestion
    <li class="breadcrumb-item{% if loop.last %} active{% endif %}" itemprop="itemListElement" itemscope
```

carakas (changes requested):

src/Frontend/Core/Layout/Templates/Breadcrumb.html.twig
crydotsnake: I fixed it!

src/Frontend/Core/Layout/Templates/Breadcrumb.html.twig
```diff
@@ -1,7 +1,7 @@
 <span class="sr-only">{{ 'lbl.YouAreHere'|trans }}</span>
 <ul class="breadcrumb" itemscope itemtype="https://schema.org/BreadcrumbList" role="navigation">
   {% for breadcrumb in breadcrumb %}
-    <li{% if loop.last %} class="active"{% endif %} itemprop="itemListElement" itemscope
+    <li{% if loop.last %} class="breadcrumb-item{% endif %}" itemprop="itemListElement" itemscope
```
carakas, line 4: it always needs the class breadcrumb-item, and active when it is the last one like I showed in the example

carakas (changes requested):

src/Frontend/Core/Layout/Templates/Breadcrumb.html.twig
crydotsnake: Okay. Now I got it!

carakas (approved):

PR #3491, opened 2022-01-20
[Feature] Reset password CLI command

SavageTiger:
## Type

- Feature / Enhancement

## Pull request description

I maintain a Website that runs on ForkCMS, it commonly happens that I get a call that somebody forgot their password since they do not use it on a regular basis.

I know there is a reset password feature, but that requires you to have access to an email-box, and as the server admin I would just like to help out the person right away.

So I developed this CLI command to reset the password using the `Symfony Console`, and since it's so useful to me, I though maybe this is also useful to other site maintainers.


src/Backend/Modules/Users/Console/ResetPasswordCommand.php
```diff
@@ -0,0 +1,62 @@
...
+
+use Backend\Modules\Users\Engine\Model as BackendUsersModel;
+use Backend\Core\Engine\User as BackendUser;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+
+class ResetPasswordCommand extends ContainerAwareCommand
```
carakas, line 13: Do you need ContainerAwareCommand or does it also work with it's parent class Command since I don't see you using the container in this command

src/Backend/Modules/Users/Resources/config/services.yml
```diff
@@ -0,0 +1,4 @@
+services:
+    users.cli.reset_password:
```
carakas, line 2: you can write the class name as service name and then it will automatically use that as the class so you don't need that argument

carakas (changes requested):
Some small things you should check but overall cool feature
Thank you for your contribution

SavageTiger:
@carakas Done and Done

carakas (approved):

PR #3499, opened 2022-03-23
Fix xss though the description in the info.xml file of a theme or module

carakas:
## Type

- Security

## Pull request description

Sanitise the HTML that is allowed in the description of a module or theme to prevent XSS

### DIFF
--- a/src/Backend/Modules/Extensions/Engine/Model.php
+++ b/src/Backend/Modules/Extensions/Engine/Model.php
@@ -839,7 +839,7 @@ class Model
         $information['name'] = (string) $module->name;
         $information['version'] = (string) $module->version;
         $information['requirements'] = (array) $module->requirements;
-        $information['description'] = (string) $module->description;
+        $information['description'] = strip_tags((string) $module->description, '<h1><h2><h3><h4><h5><h6><p><li><a>');
         $information['cronjobs'] = [];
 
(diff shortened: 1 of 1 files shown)


src/Backend/Modules/Extensions/Engine/Model.php
```diff
@@ -839,7 +839,7 @@ public static function processModuleXml(\SimpleXMLElement $xml): array
         $information['name'] = (string) $module->name;
         $information['version'] = (string) $module->version;
         $information['requirements'] = (array) $module->requirements;
-        $information['description'] = (string) $module->description;
+        $information['description'] = strip_tags((string) $module->description, '<h1><h2><h3><h4><h5><h6><p><li><a>');
```
tijsverkoyen, line 842: Fine by me, but shouldn't this be some generic function where we can centralise the safe tags?

tijsverkoyen (approved):

src/Backend/Modules/Extensions/Engine/Model.php
carakas: symfony is busy making a package (no 1.0.0 release yet) that will be included in version 6.1 that I'm already using in fork 6 so didn't think of making something custom for this.
Fixed this because of a security issue that was reported to me

PR #3507, opened 2022-04-27
Make privacy consent dialog scrollable on mobile

daphneslootmans:
## Type

- Enhancement

## Pull request description

Make the privacy consent dialog scrollable on mobile


src/Frontend/Core/Layout/Templates/Base.html.twig
```diff
@@ -1,6 +1,6 @@
 {% include "Core/Layout/Templates/Head.html.twig" %}
 
-<body class="{{ LANGUAGE }}" itemscope itemtype="http://schema.org/WebPage">
+<body class="{{ LANGUAGE }} {% if not privacyConsentDialogHide %}modal-open{% endif %}" itemscope itemtype="http://schema.org/WebPage">
```
carakas, line 3: can you explain why you have the class modal-open here? if it has to do with javascript you should use a data attribute. On top of that I'm not really a fan of having to add something there in the templates since that should be handled by the js itself

src/Frontend/Themes/Fork/Core/Layout/Templates/Base.html.twig
```diff
@@ -1,6 +1,6 @@
 {% include "Core/Layout/Templates/Head.html.twig" %}
 
-<body class="{{ LANGUAGE }}" itemscope itemtype="http://schema.org/WebPage">
+<body class="{{ LANGUAGE }} {% if not privacyConsentDialogHide %}modal-open{% endif %}" itemscope itemtype="http://schema.org/WebPage">
```
carakas, line 3: idem

carakas (changes requested):

src/Frontend/Core/Layout/Templates/Base.html.twig
daphneslootmans: To block the body from scrolling when the modal is open and scrollable. This way the class is only added when the consent dialog is visible. The consent dialog html is always rendered and just set to `display: none` when it doesn't need to be shown, so the element is always there.
I can change it so the modal html is only added when `privacyConsentDialogHide = false `and add the class to the body with js instead if you prefer.

daphneslootmans:
@carakas changes made as requested, can you check again please

daphneslootmans:
@carakas changes made as requested, can you check again please

carakas:
@daphneslootmans sorry for the delay, I was out for almost the whole month of September because of Covid

carakas (approved):

PR #3512, opened 2022-05-09
Move the apple touch icon to the theme

bjorvack:
## Type

- Enhancement

## Pull request description

Use the apple-touch-icon of the theme instead of a hardcoded path


carakas:
Is not backwards compatible

carakas (approved):

PR #3513, opened 2022-05-10
Fix redirecting in parseWidget

alessandrocraeye:
## Type

- Non critical bugfix

/

## Pull request description

This fixes the following error that occurs when submitting a form that's added via `parseWidget`:
**Type error: Argument 2 passed to Twig\Error\Error::__construct() must be of the type int, null given**

Set `lineno` to the default value `-1` (to enable its automatic guessing)


jonasdekeukelaere (approved):

carakas:
Thanks, sorry for the late merge, must have missed the notification

PR #3520, opened 2022-07-11
Make function recursiveHtmlspecialchars work with PHP8

henk-webleads:
## Type

- Non critical bugfix

## Pull request description
The code suggests that the key of an array will be corrected if there are htmlspecialchars, but it isn't.
This also works with PHP8.1

### DIFF
--- a/src/Common/Core/Model.php
+++ b/src/Common/Core/Model.php
@@ -351,10 +351,7 @@ class Model extends BaseModel
     {
         array_walk_recursive(
             $data,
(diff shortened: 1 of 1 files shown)


carakas:
Sorry for only getting back to you now for this, @henk-webleads but if you are no longer using the key why are you still referencing it. Is the key used anywhere? Can you add a bit more explanation of why you made the changes?

carakas:
Included in https://github.com/forkcms/forkcms/pull/3538

PR #3538, opened 2022-12-05
Fixed php 8.0 support

jacob-v-dam:
## Type

- Enhancement

Adds PHP 8.0 support. PHP 7.1 is not supported anymore.

## Pull request description

Upgrade packages and code to support PHP 7.4 and PHP 8.0. PHP 7.1 is not supported anymore.


.github/workflows/run-tests.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         strategy:
             fail-fast: true
             matrix:
-                php: [7.1, 7.2, 7.3, 7.4]
+                php: [7.4]
```
carakas, line 16: Can you add the php 8 versions here as well

carakas (changes requested):

carakas:
Thanks for updating @jacob-v-dam, I'll review it this weekend

jacob-v-dam:
> Thanks for updating @jacob-v-dam, I'll review it this weekend

That would be perfect! The next step is to jump to Symfony 5.4 which is LTS for now.

carakas (approved):

PR #3560, opened 2023-09-20
Add ToggleSecret to backend scripts

bjorvack:
## Type

- Feature

## Pull request description

Adds a script to toggle password/text input


src/Modules/Backend/assets/Backend/webpack/js/Components/ToggleSecret.js
```diff
@@ -0,0 +1,23 @@
+export default class ToggleSecret {
```
carakas, line 1: TogglePasswordInputType seems like a better class name that is more descriptive

carakas (changes requested):

src/Modules/Backend/assets/Backend/webpack/js/Components/ToggleSecret.js
carakas: And maybe also add this to the core components since I can see a use for this as well in the frontend if needed

tijsverkoyen:
Just an idea. So discuss: Wouldn't it be more logical to have this as a Symfony Form Type? So the developer does not need to remember to add the needed JS, but it is done automatically?

jonasdekeukelaere:
Form type with this logic built is is a good idea

src/Modules/Backend/assets/Backend/webpack/js/Components/ToggleSecret.js
jonasdekeukelaere: Fixed in [8f4137e](https://github.com/forkcms/forkcms/pull/3560/commits/8f4137eac1b25aec5dea5c703c88afa53c0b71b5) and [e8bab89](https://github.com/forkcms/forkcms/pull/3560/commits/e8bab89fee74ab2f9f37641b21d12be2068b6a87). The only thing now is that the "trigger" JS is duplicated in Frontend.js and Backend.js. Plus the block template for the form type is duplicated in (Frontend)FormLayout and (Backend)formTheme.

jonasdekeukelaere:
@tijsverkoyen Form type added

src/Modules/Backend/assets/Backend/webpack/js/Backend.js
```diff
@@ -70,6 +72,12 @@ export class Backend {
       element.passwordStrengthMeter = new PasswordStrenghtMeter($(element))
     })
   }
+
+  static initTogglePasswordInputType () {
+    $('[data-role="toggle-visibility"]').each((index, element) => {
```
tijsverkoyen, line 77: Isn't `toggle-visibility` a bit too broad? I would use `toggle-password` to be inline with all other naming

src/Modules/Backend/assets/Backend/webpack/js/Backend.js
```diff
@@ -70,6 +72,12 @@ export class Backend {
...
+      element.toggleSecret = new TogglePasswordInputType(element)
```
tijsverkoyen, line 78: `toggleSecret` → `togglePassword`?

src/Modules/Backend/templates/base/formTheme.html.twig
```diff
@@ -642,3 +642,12 @@
     {{ form_widget(form.children|last) }}
   </div>
 {% endblock %}
+
+{% block toggle_password_widget %}
+  <div class="input-group mb-3">
+    {{ form_widget(form) }}
+    <button class="btn btn-default" type="button" data-role="toggle-visibility" data-target="{{ form.vars.id }}">
```
tijsverkoyen, line 649: Wouldn't it make more sense to set the data-role on the input group? So the JS can be more scoped on the full element instead of only the button.

src/Modules/Frontend/assets/Frontend/webpack/js/_Components.js
```diff
@@ -40,4 +43,10 @@ export class Components {
       })
     }
   }
+
+  static initTogglePasswordInputType () {
+    $('[data-role="toggle-visibility"]').each((index, element) => {
```
tijsverkoyen, line 48: I find this to generic.
The inner working is very specific, as in: change the html in the element AND also change the type of an input field. While `toggle-visibility` suggests show or hide something. I think the name is not very descriptive. I suggest: `toggle-password-visibility`

src/Modules/Backend/assets/Backend/webpack/js/Backend.js
jonasdekeukelaere: Fixed a0c40abde

src/Modules/Backend/assets/Backend/webpack/js/Backend.js
jonasdekeukelaere: Fixed a0c40abde

src/Modules/Frontend/assets/Frontend/webpack/js/_Components.js
jonasdekeukelaere: Fixed a0c40abde

src/Modules/Backend/templates/base/formTheme.html.twig
jonasdekeukelaere: Fixed 2e56f912c

tijsverkoyen (approved):

carakas (approved):

PR #3562, opened 2023-09-20
Add OAuth Module

bjorvack:
## Type

- Feature

## Pull request description

Allows the user to login with microsoft azure


tijsverkoyen:
@carakas I think the changes in the Core code prevent this from being able to be a composer package. 
I don't know if this is something that should be maintained in forkcms/forkcms for the moment. But we will use this at SumoCoders for clients working with Microsoft Azure.

Although: feedback is appreciated!

carakas:
Let's leave it out for now and maybe at a later point we can see if we can implement this in a more generic way

PR #3565, opened 2023-09-22
Update toast

daphneslootmans:
## Type

- Non critical bugfix
- Enhancement

## Pull request description

- Fix toast autohiding
- initialise toast in DOM on page load
- add progress bar to toast with autohide active
- set toast autohide to true by default

### DIFF
--- a/src/Modules/Backend/assets/Backend/webpack/js/Components/InitToasts.js
+++ b/src/Modules/Backend/assets/Backend/webpack/js/Components/InitToasts.js
@@ -1,0 +1,15 @@
+export class InitBsToasts {
+  constructor () {
+    this.initToasts()
+  }
+
+  // init toasts present on page load
+  initToasts () {
+    const messageNodes = document.querySelector('[data-messaging-wrapper]').querySelectorAll('.toast')
+    const messageNodesList = Array.from(messageNodes)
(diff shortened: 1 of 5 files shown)


src/Modules/Backend/assets/Backend/webpack/js/Components/InitToasts.js
```diff
@@ -0,0 +1,15 @@
+export class InitBsToasts {
+  constructor () {
+    this.initToasts()
+  }
+
+  // init toasts present on page load
+  initToasts () {
+    const messageNodes = document.querySelector('[data-messaging-wrapper]').querySelectorAll('.toast')
```
tijsverkoyen, line 8: Wouldn't be more future proof to not use the class `.toast` to select the elements? But a data-attribute instead?

tijsverkoyen (approved):

src/Modules/Backend/assets/Backend/webpack/js/Components/InitToasts.js
daphneslootmans: True, updated

tijsverkoyen (approved):

carakas (approved):

PR #3568, opened 2023-09-29
Fork6 functional testing part 1

carakas:
## Type

- Feature

It's hard to write tests for fork actions

## Pull request description

I've implemented doctrine fixtures, they can be used in tests by loading them with the function `ForkCMS\Core\tests\WebTestCase::loadFixture()` or if you want to load them for all of your tests in that class you can simply return them as an array by overwriting the function `ForkCMS\Core\tests\WebTestCase::getClassFixtures()`
The benefit of working with the fixtures is that we can in a later stage also use them to bring back the demo data option or people can just load them using the command `bin/console doctrine:fixtures:load --append`

When creating a fixture you should extend from the class `ForkCMS\Modules\Extensions\tests\ForkFixture`. That way your fixture will automatically get the module name as group making it simple to import only that module for instance `bin/console doctrine:fixtures:load --append --group=Pages`

If the test database does not exist yet it should also be automatically be created and populated. If you want to repopulate the database run the tests with the environment variable `TEST_DATABASE=fresh`

Every test starts with a freshly installed fork, we roll back changes made during the previous test so this goes quickly.

The class `ForkCMS\Core\tests\WebTestCase` contains a bunch of extra asserts that make it easier to write functional tests for ForkCMS. If you write tests for the backend you can use `ForkCMS\Modules\Backend\tests\BackendWebTestCase`, this class adds extra test methods specifically needed for the backend like authentication and datagrid asserts


src/Modules/Backend/assets/installer/translations.xml
```diff
@@ -1589,6 +1674,10 @@
         <translation locale="nl"><![CDATA[Het ingegeven wachtwoord is te kort.]]></translation>
         <translation locale="en"><![CDATA[The password is too short.]]></translation>
       </item>
+      <item type="error" name="EmailIsTooLong">
```
tijsverkoyen, line 1677: Where is this used? And why is it needed?

src/Modules/Backend/tests/Backend/Actions/UserAddTest.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public function testValidData(): void
+    {
+        self::loadPage();
+
+        self::submitForm(
+            'Add',
+            [
+                'user[user][ec05aaca240e74a0604d93f9e5a7caef][displayName]' => 'Jelmer Prins',
```
tijsverkoyen, line 41: Where is the string `ec05aaca240e74a0604d93f9e5a7caef` generated? Will it always be the same? How can a tester get this?

src/Modules/Backend/tests/Backend/Actions/UserAddTest.php
```diff
@@ -0,0 +1,87 @@
...
+        self::assertHasLink('Cancel', '/private/en/backend/user-index');
+    }
+
+    public function testEmptyFormShowsValidationErrors(): void
+    {
+        self::loadPage();
+        self::assertEmptyFormSubmission('user', 3);
+    }
+
+    public function testValidData(): void
```
tijsverkoyen, line 34: `testWithInvalidData` as you are testing with invalid data?

src/Modules/Backend/tests/BackendWebTestCase.php
```diff
@@ -0,0 +1,141 @@
+<?php
+
+namespace ForkCMS\Modules\Backend\tests;
```
tijsverkoyen, line 3: Why not `ForkCMS\Modules\Backend\Tests`, tests with uppercase?

.github/workflows/run-tests.yml
```diff
@@ -69,22 +69,49 @@ jobs:
...
+                  APP_ENV: test
+                  FORK_DATABASE_HOST: 127.0.0.1
+                  FORK_DATABASE_PORT: 3306
+                  FORK_DATABASE_NAME: forkcms_test
+                  FORK_DATABASE_USER: root
+                  FORK_DATABASE_PASSWORD: kingtriton
 
             - name: Execute tests without coverage
               if: ${{ matrix.php != '8.1' }}
-              run: bin/simple-phpunit --testsuite=${{ matrix.testsuite}}
```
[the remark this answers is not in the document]
carakas: Rebased the commit to remove the -v
The original test suite was already configured like this, I think it's mostly to win speed since once you have coverage of one you don't really need it of the other ones, unless you know of a reason why we should run the coverage in all versions of the matrix?

.github/workflows/run-tests.yml
```diff
@@ -69,22 +69,49 @@ jobs:
...
+            - name: Install Fork CMS
+              run: cp src/Modules/Installer/tests/fork-cms-installation-configuration.yaml fork-cms-installation-configuration.yaml && bin/console forkcms:installer:install && sed -i 's/forkcms_test/forkcms/g' .env.local
+              env:
+                  FORK_DATABASE_USER: root
+                  FORK_DATABASE_PASSWORD: kingtriton
+
             - name: Execute tests with coverage
               if: ${{ matrix.php == '8.1' }}
-              run: bin/simple-phpunit --testsuite=${{ matrix.testsuite}} --coverage-clover=${{ matrix.testsuite}}.clover
+              run: bin/simple-phpunit --testsuite=${{ matrix.testsuite}} --coverage-clover=${{ matrix.testsuite}}.clover -v
```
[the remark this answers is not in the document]
carakas: Left over from debugging indeed, rebased the commit and removed it

src/Core/tests/WebTestCase.php
```diff
@@ -0,0 +1,288 @@
...
+
+    final protected static function assertHasLink(string $text, string $url): void
+    {
+        static::assertMinCount(
+            1,
+            static::getCrawler()->filter('a:contains("' . $text . '")[href="' . $url . '"]')
+        );
+    }
+
+    public static function assertMinCount(int $expectedCount, Countable|iterable $haystack, string $message = ''): void
```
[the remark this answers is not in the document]
carakas: dded this one because I'm expecting at least 3 errors for instance, but then on the sumo fork there is like a fourth required field it would fail. I normally also check that the error messages I'm expected are visible.
But it might be better that if a fourth error message is shown the test actually should fail, in that case we don't need this one

src/Modules/Backend/assets/installer/translations.xml
carakas: In the user DTO, has to do with max length of the field in the database because of indexes. Otherwise it would get cut off without letting the user know but on login the email address wouldn't match. Email max length is set at 180

src/Modules/Backend/tests/BackendWebTestCase.php
carakas: I based the naming of the directory on https://symfony.com/doc/current/bundles/best_practices.html#tests.
It does look a bit weird, but it is in lowercase because the directory name is lowercase.
We could deviate here and rename the tests directory to Tests and then we could have it in upper case?

src/Modules/Backend/tests/Backend/Actions/UserAddTest.php
carakas: It comes from here: https://github.com/forkcms/forkcms/blob/b1b0c9226fe8fbcf14792848835a9eb553d022b8/src/Core/Domain/Form/TabsType.php#L17
It's always the same indeed, it's the md5 has of the label of the tab like lbl.Authentication.
I used md5 because if I just use the label I get the following error
`The name "lbl.Authentication" contains illegal characters. Names should start with a letter, digit or underscore and only contain letters, digits, numbers, underscores ("_"), hyphens ("-") and colons (":").`

We could use something different as naming strategy though for the tab part of the form, I'm open for suggestions

src/Modules/Backend/tests/Backend/Actions/UserAddTest.php
carakas: Fixed in f9a393b8c7ae39260fdf9933aa0282408c2cac35

src/Modules/Backend/DataFixtures/UserFixture.php
```diff
@@ -0,0 +1,56 @@
...
+    public const USER_REFERENCE = 'user-super-admin';
+
+    public function __construct()
+    {
+    }
+
+    public function load(ObjectManager $manager): void
+    {
+        $createSuperAdmin = new CreateUser();
+        $createSuperAdmin->email = 'super-admin@fork-cms.com';
```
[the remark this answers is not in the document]
carakas: Fixed in f9a393b8c7ae39260fdf9933aa0282408c2cac35

src/Core/tests/WebTestCase.php
```diff
@@ -0,0 +1,288 @@
...
+            static::getClient()->getCrawler()->selectLink($linkText)->link()->getUri(),
+            null,
+            $expectedContent,
+            $httpStatusCode,
+            $requestMethod,
+            $requestParameters
+        );
+    }
+
+    final protected static function assertResponseHasContent(string ...$content): void
```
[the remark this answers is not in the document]
carakas: Rebased and fixed

.github/workflows/run-tests.yml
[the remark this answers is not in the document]
tijsverkoyen: nm, I did not spot the that it will only be done on PHP 8.1

src/Core/tests/WebTestCase.php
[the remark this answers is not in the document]
tijsverkoyen: hmm, in PHPUnit itself the methods are without the `Has` part, `assertContains`, `assertStringContainsString`, `assertContainsOnly`, ... That was why I proposed `assertResponseContains` instead of `assertResponseHasContent`.

But it is naming, so it will always be a personal preference. So if you don't agree: fine by me.

src/Modules/Backend/assets/installer/translations.xml
tijsverkoyen: Ok, good to know.
I don't know if this is something that should be documented in the `User`-entity class. Because according to the RFC about email addresses:

> That limit is a maximum of 64 characters (octets) in the "local part" (before the "@") and a maximum of 255 characters (octets) in the domain part (after the "@") for a total length of 320 characters. However, there is a restriction in RFC 2821 on the length of an address in MAIL and RCPT commands of 254 characters.

Maybe just add a comment to explain that it is because of the indexes in the database.

src/Modules/Backend/tests/Backend/Actions/UserAddTest.php
tijsverkoyen: I would prefer something readable. Maybe use: https://symfony.com/doc/current/components/string.html#slugger?

src/Modules/Backend/tests/BackendWebTestCase.php
tijsverkoyen: No, follow the best practices.

src/Core/tests/WebTestCase.php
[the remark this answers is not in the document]
tijsverkoyen: I just find it somewhat confusing as it is just a oneliner. So it seems to be just an alias. 
Or something that is not really needed and only more "magic" that is not really documented, while all methods from PHPUnit are well documented and can be found more easily.

But I do understand it from a developer point of view. So fine by me.

src/Core/tests/WebTestCase.php
```diff
@@ -0,0 +1,288 @@
...
+    {
+        $response = static::getResponse();
+
+        foreach ($content as $expectedContent) {
+            static::assertStringContainsString($expectedContent, $response->getContent(), 'Page does not contain "' . $expectedContent . '".');
+        }
+    }
+
+    final protected static function assertResponseDoesNotHaveContent(string ...$content): void
```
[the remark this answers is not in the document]
tijsverkoyen: See above.

PR #3588, opened 2023-11-30
Symfony 6.4

carakas:
## Type

- Enhancement

## Pull request description

I've updated the packages to 6.4, if we want to do 7 the minimum PHP version would be 8.2, should we bump to 8.2?

I'll do Symfony 7 later, still waiting for some dependencies to update and not sure how long that will take yet, symfony 6.4 is the same as 7, just with deprecation warnings


webpack.config.js
```diff
@@ -17,6 +17,10 @@ const THEME_PATH = { output: 'public/assets/themes', public: '/assets/themes' }
 const MODULE_PATH = { output: 'public/assets/modules', public: '/assets/modules' }
 const EXPORTS = []
 
+if (!Encore.isRuntimeEnvironmentConfigured()) {
+  Encore.configureRuntimeEnvironment(process.env.NODE_ENV || 'dev');
```
tijsverkoyen: Wouldn't `prod` be a more sane default?

carakas: fixed in f8027ef

tijsverkoyen (dismissed):

